PR #68, opened 2022-12-12
how to use annotation with create_widget

### LINKED ISSUE #5340: ComboBox automatically filled with viewer.layers for `QWidget` class plugins
## 📚 Documentation

It would be great if this page https://napari.org/stable/guides/magicgui.html#using-magicgui-in-napari-plugin-widgets would also explain, how to create a Layers `ComboBox` in a `QWidget` class plugin. It is straight-forward with the `@magicgui` decorator. But if you want something similar in a `QWidget` class plugin, information on how to do this is not so plentyful. The following code snippet worked for me, but I do not know whether this is the best way to go:
```python
from magicgui.widgets import ComboBox
from qtpy.QtWidgets import QWidget

from napari.types import LabelsData
from napari.utils._magicgui import get_layers_data


class SelectLabelsLayerWidget(QWidget):
    def __init__(self, napari_viewer):
        super().__init__()
        self.viewer = napari_viewer

        self.labels_data = ComboBox(
            label="Labels layer: ",
            annotation=LabelsData,  # essential: either layer or LayerData
            choices=get_layers_data,  # internal function that finds the appropriate layers, essential
        )
        # update the choices when new layer is inserted
        self.viewer.layers.events.inserted.connect(
            self.labels_data.reset_choices
        )
        # update the choices when layer is removed
        self.viewer.layers.events.removed.connect(
            self.labels_data.reset_choices
        )
        # update the choices when layer is converted
        self.viewer.layers.events.changed.connect(
            self.labels_data.reset_choices
        )
```
It should be mentioned that `choices=napari.utils._magicgui.get_layers_data` is needed for `annotation=napari.types.*Data` and `choices=napari.utils._magicgui.get_layers` is needed for `annotation=napari.layers.*`. 

If you confirm that the above approach is the way to go, and that this should be part of the above mentioned guide in the docs, I will create a PR for this.

gatoniel:
# Description

Adds to the documentation how to use annotations with `magicgui.widgets.create_widget` in a plugin subclass of `QtWidgets.QWidget`.

## Type of change

# References

closes https://github.com/napari/napari/issues/5340

uses information from https://github.com/PartSeg/PartSeg_bioimageio/blob/2afcbd558e52d6af6a13f0b844e80d280cd6ba9d/src/PartSeg_bioimageio/_widget.py#L421

## Final checklist:
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added [alt text](https://webaim.org/techniques/alttext/) to new images included in this PR


docs/guides/magicgui.md
```diff
@@ -552,3 +552,31 @@ class MyGui(FunctionGui):
 def napari_experimental_provide_dock_widget():
     return MyGui
 ```
+
+### `magicgui.widgets.create_widget`
+
+If you want to create a layer selection as [shown above](magicgui-and-type-annotations) but you cannot use the {func}`@magicgui <magicgui.magicgui>` decorator, you can create the selection widget with the {func}`create_widget <magicgui.widgets.create_widget>` function. In the following example we subclass `QtWidgets.QWidget` and add a widget that is annotated as {attr}`napari.types.ImageData`
```
melissawm:
```suggestion
If you want to create a layer selection as [shown above](#magicgui-and-type-annotations) but you cannot use the {func}`@magicgui <magicgui.magicgui>` decorator, you can create the selection widget with the {func}`create_widget <magicgui.widgets.create_widget>` function. In the following example we subclass `QtWidgets.QWidget` and add a widget that is annotated as {attr}`napari.types.ImageData`:

```

melissawm (commented):
Hi @gatoniel - welcome and thanks for your PR! I left one minor comment (to sctually create a link to the `magicgui-and-type-annotations` section) but will leave the actual example to be reviewed by one of the core maintainers. Cheers!

Czaki:
This is a much more generous mechanism. `magicgui.widgets.create_widget` allows you to get magicgui widget for any type registered in magicgui. 

So getting layer selector should be mentioned as an example of usage, not primary usage. 

Maybe it could also be nice to add an example using `magicgui.widgets.Container` for complex examples that do not require direct Qt usage.

melissawm:
Hi folks - I'd love to see this one go in. I will link to it in zulip to maybe get more eyes on it 😄

docs/guides/magicgui.md
Czaki: Based on my experience now I prefer to modify the example and suggest using `magic_factory` instead of magicgui.

jni: @Czaki can you elaborate here? The example doesn't use the magicgui decorator at all, which is what magic_factory replaces. So what would you suggest here?

Anyway TIL about create_widget so I like this! 😅

Czaki: I mean update this docstring to point magic factory.

melissawm:
Hi @gatoniel - do you need help with updating your branch, or moving this forward? Thanks!

psobolewskiPhD:
@lucyleeow @DragaDoncila 
This fell through the cracks -- how does this fit in with the current state of extending napari docs?

lucyleeow:
This looks useful but I think it is missing some context. In what situations would you not be able to use the `@magicgui` decorator? 

I think this would be a better example than what is used here: https://napari.org/dev/howtos/extending/magicgui.html#qtwidgets-qwidget and we can over-write it. Could we add some more detail about how `create_widget` gives you more control over the selection widget created (i.e. we can specify the widget type (e.g., slider, RangeEdit etc)). I would also make it more general and not specifically about layer selection, though we can use that as an example. 

@gatoniel are you still interested in continuing this PR?

melissawm:
Gentle ping, @gatoniel ! Let us know if you need help 😄

lucyleeow:
> I think this would be a better example than what is used here: https://napari.org/dev/howtos/extending/magicgui.html#qtwidgets-qwidget and we can over-write it.

Reading again, I think it would be a good addition to the existing example. We could explain why/how `create_widget` makes things easier?

gatoniel:
Hi, sorry for the late reply. I was / am busy with other stuff. Maybe I will have time this weekend.

gatoniel:
> This looks useful but I think it is missing some context. In what situations would you not be able to use the @magicgui decorator?

I do not know whether this question requires an answer as long as the information is available at https://napari.org/dev/howtos/extending/magicgui.html#qtwidgets-qwidget. For now, I have added the example as a special case there as people using `QtWidgets.QWidget` might be interested in still using the type annotated widgets with automatic synchronization.

> Could we add some more detail about how create_widget gives you more control over the selection widget created (i.e. we can specify the widget type (e.g., slider, RangeEdit etc)). I would also make it more general and not specifically about layer selection, though we can use that as an example.

Unfortunately, I do not know enough about the other possibilities of `create_widget`.

> Maybe it could also be nice to add an example using magicgui.widgets.Container for complex examples that do not require direct Qt usage.

However, in the example under https://napari.org/dev/howtos/extending/magicgui.html#magicgui-widgets-container `create_widget` is already used. Hence, I would actualy propose to close this PR without merging as now it becomes clear how to use `create_widget` to get for example layer selections without use of `@magicgui`.

lucyleeow:
I think there is very useful and important information here (e.g., regarding refreshing the widget) that is not documented anywhere else. Fitting it in with existing documentation is more tricky and probably more work than originally intended. How about I try and add this example, along with some more info, and you can review if you are interested @gatoniel ?

gatoniel:
Sounds good. Otherwise, I could also add some more description about the refreshing into the surrounding text.
Additionally, I guess it would be good to have a link from https://napari.org/dev/howtos/extending/magicgui.html#parameter-annotations to the availability of `create_widget`.

gatoniel:
I have now added some more description about the refreshing and a note box in the #parameter-annotations section to direct the reader directly to this example if she is not concerned about `@magicgui`.

lucyleeow:
Ah that is interesting. @DragaDoncila and I were talking about whether the `reset_choices` method in your example somehow overrides the parent `CategoricalWidgets` [`reset_choices`](https://github.com/napari/docs/pull/68#issuecomment-2033969633) method or if it is just called manually. If it is the latter, then technically it doesn't matter what this method is called right? Its just something we manually call in `ShowEvent` ?

lucyleeow:
Now that I understand better, I think we could just have one `QWidget` example. The first example widget doesn't really do anything useful, but it does show how to connect a function to an event. I would maybe amend your example to add a function to execute when a layer is selected.

I think some more details about what is happening in the 2 methods would be useful too. And highlight that we need to update ourselves as magicgui is not doing this for us.

There are also several improvements I would like to make to that page, but this can be separate to this addition.

Overall, maybe this is too much extra work for you?

docs/howtos/extending/magicgui.md
```diff
@@ -783,3 +787,36 @@ viewer.window.add_dock_widget(my_widg)
 
 As above to turn this into a [plugin widget contribution](widgets-contribution-guide),
 simply provide the class definition and add to the plugin manifest.
+
+#### `magicgui.widgets.create_widget`
+
+You might want to add a layer selection as [shown above](#parameter-annotations) into your highly customizable `QtWidgets.QWidget` that is always synchronized with the available {attr}`napari.types.ImageData` layers in your viewer. For this you can use the {func}`create_widget <magicgui.widgets.create_widget>` function as described in the following example.
+
+To synchronize the information between the napari viewer (i.e. the available layers) the function `reset_choices` of the dropdown widget needs to be manually called whenever the main widget is shown. In the following example, this is done by defining two class methods `showEvent` and `reset_choices`.
```
lucyleeow: I would have thought it needs to be called when the input selector widget is shown? If the main widget is open. then layers are changed, you'd expect the input selector widget to update?

gatoniel:
> Ah that is interesting. @DragaDoncila and I were talking about whether the reset_choices method in your example somehow overrides the parent CategoricalWidgets https://github.com/napari/docs/pull/68#issuecomment-2033969633 method or if it is just called manually. If it is the latter, then technically it doesn't matter what this method is called right? Its just something we manually call in ShowEvent?

Yes, technically it doesn't matter what this method is called. One could also call the `self.layer_select.reset_choices(event)` in the `showEvent` method.

> Overall, maybe this is too much extra work for you?

Depends. If I get rather precise orders, I can try :D

docs/howtos/extending/magicgui.md
gatoniel: I tested it. Napari somehow internally calls `reset_choices` of the main widget. So `showEvent` and `reset_choices` need to be named exactly like that. I updated it in the PR

gatoniel:
> Yes, technically it doesn't matter what this method is called. One could also call the self.layer_select.reset_choices(event) in the showEvent method.

This was wrong. I tested the available options. It works only smoothly, when the function is exactly called `reset_choices`. It seems that napari calls this function of the docked widgets whenever something changed within napari. But I do not know where in the source code this call comes from. I've updated the PR accordingly. `showEvent` should just simply also call `reset_choices` as the choices should be reset whenever the widget is being shown.

DragaDoncila:
> It works only smoothly, when the function is exactly called reset_choices

Yeah I found this also but I couldn't for the life of me figure out where the connection is made. 

I can see that for magicgui widgets we have a [signal](https://github.com/pyapp-kit/magicgui/blob/main/src/magicgui/widgets/bases/_categorical_widget.py#L64) `native_parent_changed`, that calls the `CategoricalWidget`'s `reset_choices` method - so I guess this would call `layer_select.reset_choices` when the containing widget, `ExampleLayerListWidget` changes. 

What I don't understand is how having a method named `reset_choices` on a parent widget of a magicgui widget (in this case `ExampleLayerListWidget`) ensures that this method gets connected to `layerlist` events. There's no inheritance going on since `ExampleLayerListWidget` just inherits from `QWidget`, so `magicgui` must be reaching up from the `layer_select` widget to the parent widget somehow.

Pinging @tlambert03 who can hopefully unravel some of the magic in magicgui.

tlambert03:
> What I don't understand is how having a method named reset_choices on a parent widget of a magicgui widget (in this case ExampleLayerListWidget) ensures that this method gets connected to layerlist events.

a grep of napari for reset_choices suggests its probably this:

https://github.com/napari/napari/blob/3cb47b5fc393427a09f3b98d641b3086ff93405e/napari/_qt/qt_main_window.py#L1093-L1100

(i.e. it's not magicgui magic, it's napari magic... i'm not sure if you want to "codify" that as a first class API here, but that's why it's working 😄 )

DragaDoncila:
> it's not magicgui magic, it's napari magic...

:grimacing: Thanks for that Talley, I just figured it must be magicgui but this is worse...

> i'm not sure if you want to "codify" that as a first class API here

I kinda don't... it's just an accident of magicgui `Categorical`s having `reset_choices` methods, and it's much more brittle than I was imagining. I'll have a bit of a think about a blessed way to keep `magicgui` widgets in sync when they're part of a bigger non-magicgui widget

DragaDoncila:
@gatoniel @lucyleeow I would say to move this PR forward while we think about how/if we want to maintain this `reset_choices` behavior, we should update the example to explicitly connect the `layer_select.reset_choices` method to `viewer.layers` events. It's a little more boilerplate, but a lot less magic, and I would prefer to have that documented rather than this brittle kinda weird thing we're doing in napari specifically for `CategoricalWidget`s.

lucyleeow:
@gatoniel thanks for uncovering all of this, sorry this PR has gotten a bit derailed.

gatoniel:
I changed the example. Now the `reset_choices` is directly connected to the three events in `viewer.layers.events`. I agree, this makes the example more straightforward and less magic :D Thanks @tlambert03 and @DragaDoncila for finding the mechanism...

lucyleeow:
Thanks @gatoniel.

I still think that we only need one `QWidget` example. I would connect your widget to  a function that does something vaguely useful with an input layer. Also mention that we need to connect to the layer events because magicgui is not doing this automatically for us, as with the above examples.

Could you also split your lines so max line length is 79 characters (this is what we've set for napari (see: [here](https://github.com/napari/napari/blob/3cb47b5fc393427a09f3b98d641b3086ff93405e/pyproject.toml#L203)) but the docs repo has no linting). Thank you

docs/howtos/extending/magicgui.md
```diff
@@ -178,6 +178,10 @@ The following napari types may be used as *parameter* type annotations in
   {attr}`napari.types.ImageData` or  {attr}`napari.types.LabelsData`
 - {class}`napari.Viewer`
 
+```{note}
+If you're interested in getting synchronized information from the napari viewer into widgets without `@magicgui`, see the [`create_widget` example](#magicgui-widgets-create_widget) below.
+```
```
lucyleeow: I don't think this is necessary. I think we automatically update for all magicgui registered napari types; layer, layer subclasses and layer type data (e.g., `napari.types.ImageData`). 
I am not sure if there would be any cases where you would want an input that is not in the above list.

If there is, we could mention that for non registered types you need to manually synchronize, see `QWidget` example for more...

cc @DragaDoncila

gatoniel: I have made the note more precise. I added it, because I was aware that one can easily create these synchronized dropdown lists of the available layers with `@magicgui`. But then, they were part of a bigger widget, and I was not able to figure out to create only these widgets which is possible with `create_widget`.

gatoniel:
> Could you also split your lines so max line length is 79 characters (this is what we've set for napari (see: [here](https://github.com/napari/napari/blob/3cb47b5fc393427a09f3b98d641b3086ff93405e/pyproject.toml#L203)) but the docs repo has no linting). Thank you

Done. Is it ok now?

> I still think that we only need one QWidget example. I would connect your widget to a function that does something vaguely useful with an input layer. Also mention that we need to connect to the layer events because magicgui is not doing this automatically for us, as with the above examples.

It is only one widget now. Clicking on the button prints the shape of the currently selected Image Layer. I mentioned that it needs manual connection to the layers events.

docs/howtos/extending/magicgui.md
```diff
@@ -754,29 +762,63 @@ simply provide the class definition and add to the plugin manifest.
...
 For the most control over your widget, subclass
-[`QtWidgets.QWidget`](https://doc.qt.io/qt-5/qwidget.html):
+[`QtWidgets.QWidget`](https://doc.qt.io/qt-5/qwidget.html).
+In the following example, we create a button and a dropdown list.
+The available choices for the dropdown list are the
+current layers in the viewer. For this, we use
+{func}`create_widget <magicgui.widgets.create_widget>` with the annotation
+{attr}`napari.types.ImageData`.
+In opposition to
+using `@magicgui` as [shown above](#parameter-annotations), we now need to
```
lucyleeow: I think we are still using parameter annotations with `create_widget` (`annotation=ImageData`), it's just that the 'parent' widget is not `magicgui` widget but a `QWidget`, thus we have to manually connect the `reset_choices`. See: #399

docs/howtos/extending/magicgui.md
```diff
@@ -754,29 +762,63 @@ simply provide the class definition and add to the plugin manifest.
...
         self.viewer = viewer
 
+        # Create a button
         btn = QPushButton("Click me!")
+        # Connect the click event to a function
         btn.clicked.connect(self._on_click)
 
+        # create new widget with create_widget and type annotation
+        self.layer_select = create_widget(annotation=ImageData)
+        # The widget has to be connected to viewer.layers.events
```
lucyleeow: we're connecting the widgets `reset_choices` method.
```suggestion
        # The `layer_select` widgets `reset_choices` method has to be connected to viewer.layers.events
```

docs/howtos/extending/magicgui.md
```diff
@@ -178,6 +178,14 @@ The following napari types may be used as *parameter* type annotations in
   {attr}`napari.types.ImageData` or  {attr}`napari.types.LabelsData`
 - {class}`napari.Viewer`
 
+```{note}
+If you want to get the dropdown list widgets resulting from the layer
+annotations (e.g., {class}`~napari.layers.Layer`,
+{class}`~napari.layers.Image`, {attr}`napari.types.ImageData`)
+without using the `@magicgui` decorator, 
+see the [`QWidget` example](#qtwidgets-qwidget) below.
```
lucyleeow: I see your point. What about:

```suggestion
When creating a widget that is not a 
{class}`~magicgui.widgets.bases.ContainerWidget` subclass,
adding a layer input widget requires more than just
parameter annotation.
See the [`QWidget` example](#qtwidgets-qwidget) below.
```

formating may be off but the content is there.

lucyleeow (commented):
Looks good, just some nits, thanks!

docs/howtos/extending/magicgui.md
gatoniel: I do not understand that comment. Can you elaborate more on this?

lucyleeow: Ah sorry, I wasn't clear and it's difficult to word this part as its complicated.

Whenever you use a `magicgui` `ContainerWidget`, type annotating or `create_widget` widget works without any more work. So this means e.g.,
* the `@magicgui` decorator,
* subclassing `ContainerWidget`,
* subclassing `FunctionGui` (see https://pyapp-kit.github.io/magicgui/widgets/#the-widget-hierarchy).

(not just `@magicgui` as we've said in the text).

I guess the #parameter-annotations link is also awkward, because it only talks about using type hints in `@magicgui` but not in `create_widgets`. I think this page itself just needs work. I'm happy to leave the link as is and we can think about improving this page later.

DragaDoncila: I think it would be clearest here to say the following:

```suggestion
Because the layer selection widget will be housed by a native `QWidget`
and not by a `magicgui` subclass as [shown above](#parameter-annotations), we now need to
```

DragaDoncila (approved):
Thanks for your work and patience on this @gatoniel! I suggested a more explicit comment about why we have to manually connect to `reset_choices` now, but am approving otherwise. Let's get this in!

docs/howtos/extending/magicgui.md
```diff
@@ -754,29 +762,63 @@ simply provide the class definition and add to the plugin manifest.
...
-[`QtWidgets.QWidget`](https://doc.qt.io/qt-5/qwidget.html):
+[`QtWidgets.QWidget`](https://doc.qt.io/qt-5/qwidget.html).
+In the following example, we create a button and a dropdown list.
+The available choices for the dropdown list are the
+current layers in the viewer. For this, we use
+{func}`create_widget <magicgui.widgets.create_widget>` with the annotation
+{attr}`napari.types.ImageData`.
+In opposition to
+using `@magicgui` as [shown above](#parameter-annotations), we now need to
+manually connect the `reset_choices` of the list with the
```
lucyleeow: Last nit. Technically `create_widget` makes a `CategoricalWidget` which has a `reset_choices`. Maybe this will do?
```suggestion
manually connect the `reset_choices` of  created widget with the
```

docs/howtos/extending/magicgui.md
gatoniel: Now I understand. Yes, that is an important point that I was not aware of / thinking of.

docs/howtos/extending/magicgui.md
gatoniel: I added that change manually, there was a problem adding it here automatically....

docs/howtos/extending/magicgui.md
```diff
@@ -754,29 +762,63 @@ simply provide the class definition and add to the plugin manifest.
...
+In the following example, we create a button and a dropdown list.
+The available choices for the dropdown list are the
+current layers in the viewer. For this, we use
+{func}`create_widget <magicgui.widgets.create_widget>` with the annotation
+{attr}`napari.types.ImageData`.
+Because the layer selection widget will be housed by a native `QWidget`
+and not by a `magicgui` subclass as [shown above](#parameter-annotations), we now need to
+manually connect the `reset_choices` of the created widget with the
+`viewer.layers.events` so that the available choices are synchronized
+with the current layers of the viewer:
```
Czaki, line 775:
```suggestion
define `reset_choices` method that will need to call `reset_choices`
of magicgui widgets.
```

docs/howtos/extending/magicgui.md
```diff
@@ -754,29 +762,63 @@ simply provide the class definition and add to the plugin manifest.
...
+        self.layout().addWidget(self.layer_select.native)
         self.layout().addWidget(btn)
 
     def _on_click(self):
-        print("napari has", len(self.viewer.layers), "layers")
+        print(
+            "Selected layer has shape: ",
+            self.layer_select.value.shape,
+        )
+
```
Czaki, line 814:
```suggestion

    def reset_choices(self):
        self.layer_select.reset_choices()
```

docs/howtos/extending/magicgui.md
```diff
@@ -754,29 +762,63 @@ simply provide the class definition and add to the plugin manifest.
...
+        # Connect the click event to a function
         btn.clicked.connect(self._on_click)
 
+        # create new widget with create_widget and type annotation
+        self.layer_select = create_widget(annotation=ImageData)
+        # The `layer_select` widgets `reset_choices` method has to be connected to viewer.layers.events
+        layers_events = self.viewer.layers.events
+        layers_events.inserted.connect(self.layer_select.reset_choices)
+        layers_events.removed.connect(self.layer_select.reset_choices)
+        layers_events.reordered.connect(self.layer_select.reset_choices)
```
Czaki, line 802:
```suggestion
```

Czaki (commented):
If a widget has `reset_choices` method, then it will be called on all required events.

Fell free to reformat my text.

docs/howtos/extending/magicgui.md
lucyleeow: https://github.com/napari/docs/pull/68#issuecomment-2036178071

@Czaki

gatoniel: Maybe, we can explain that part of napari magic, too, and show both possibilities with a marker that one of them is potentially deprecated?

Czaki: We are working on nested Layers (Layer Groups). It may change a list of events. So then people will need to update their plugins if they do not depend on napari calling `reset_choices`.

DragaDoncila: > So then people will need to update their plugins if they do not depend on napari calling reset_choices.

@Czaki I'm not happy to enshrine this behaviour in our docs right now (see original comment [here](https://github.com/napari/docs/pull/68#issuecomment-2036033694)). It's very brittle and in my opinion, quite an overreach that could lead to surprising effects on widgets totally unrelated to the layerlist. I would strongly prefer to keep the example as it currently is, because it involves no magic and no reliance on `napari` internals. If we one day change the `layerlist` events then yes, people may have to update their plugins, but I have no issue with that.

gatoniel: Could we explain the shortcut via `reset_choices` in the docs and also mention that there are plans that both `viewer.layers.events` and `reset_choices` might change in future, so that developers should check the corresponding source code once there plugins stop working as expected.

`reset_choices` https://github.com/napari/napari/blob/50b314f7ad2e7cd4ef21b5705b3c6b363c0857ed/napari/_qt/qt_main_window.py#L1093 shows all the relevant `layers.events` that would need to be connected individually.

`viewer.layers.events` currently is a instance subclassing https://github.com/napari/napari/blob/50b314f7ad2e7cd4ef21b5705b3c6b363c0857ed/napari/utils/events/containers/_selectable_list.py#L12

jni:
Hi folks,

I'm with Draga and Lucy here — I find the napari magic um, concerning, 😂, insomuch as it introduces a way-too-tight coupling between napari and magicgui, so I would rather we use the explicit connections in the docs. Yes @Czaki, the events might change, but they are part of the public API so in that case we would use a deprecation warningemitter cycle and so on and we would be able to update. (Of course, keeping the docs in sync is still a challenge, but that is the case either way — we might also remove the totally-private reset_choices magic and then the example would stop working — with no warning.)

@gatoniel thanks for offering to document both options — that is appreciated, but even if we decide to go that route, in my opinion the example as it is now is good and should go in, and we can expand on the magic in a subsequent PR.

I'll mark this as ready-to-merge and if @Czaki agrees with my rationale above, he can merge. 🙏 🚀 

Thanks all for the discussion! I had actually not seen this page @lucyleeow, it is friggin GOLD. 🥇

Czaki:
I still do not understand why you call this magic. 
This is only the interface.

jni:
it is magical to connect events behind the scenes based on a method *name*. (Not the signature, nor what the method does.) This is especially true if the interface is not annotated anywhere, specifically in the function that is adding the widget.

docs/howtos/extending/magicgui.md
jni: Thanks for the offer @gatoniel —**if** we decide to take you up on it or parts of it, I think expanding on the reset_choices magic should come in a separate PR. For now I think we are still internally deciding whether we want it in the docs, as you can see 😅, so I think this should go in as-is.

Regarding the future of the event connections: in my opinion, if we start worrying in the docs about all the things we are planning to deprecate, we will never document anything. 😂 I think in general the docs should be written with regard to the current public API, unless we *really* want to discourage use of a specific API. Which is not the case here.

jni (approved):

jni:
At any rate @Czaki I don't think you can argue that layer events are public and therefore can be considered at least as stable as this internal connection, whether or not you call it magic? And that the docs are improved by this PR. I'd be happy to move this discussion to a new PR that discusses *both* options and their tradeoffs, as proposed by @gatoniel.

PR #71, opened 2022-12-13
Add napari survey text and link for announcement banner

melissawm:
# Description
Adds an announcement banner to the docs pointing to the napari survey.

## Type of change

# References
This works because of the recent update on the napari-sphinx-theme: https://github.com/napari/napari-sphinx-theme/pull/127

## Final checklist:

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -100,7 +100,7 @@ html_theme_options = {
         "json_url": "https://napari.org/version_switcher.json",
         "version_match": version_match,
     },
-    "announcement": "Take the 8 minute <a href='https://bit.ly/napari-survey-n'>annual napari survey</a>! Your feedback is vital to ensure napari serves community needs.",
+    "announcement": "Take the 8 minute <a href='https://bit.ly/napari-survey-n' target='_blank' rel='noopener'>annual napari survey</a>! Your feedback is vital to ensure napari serves community needs.",
 }
 
 # Add any paths that contain custom static files (such as style sheets) here,
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -100,6 +100,7 @@ html_theme_options = {
         "json_url": "https://napari.org/version_switcher.json",
         "version_match": version_match,
     },
+    "announcement": "Take the 8 minute <a href='https://bit.ly/napari-survey-n'>annual napari survey</a>! Your feedback is vital to ensure napari serves community needs.",
 }
 
 # Add any paths that contain custom static files (such as style sheets) here,


kandarpksk:
LGTM!

docs/conf.py
```diff
@@ -100,6 +100,7 @@
         "json_url": "https://napari.org/version_switcher.json",
         "version_match": version_match,
     },
+    "announcement": "Take the 8 minute <a href='https://bit.ly/napari-survey-n'>annual napari survey</a>! Your feedback is vital to ensure napari serves community needs.",
```
codemonkey800: since it's an external link, could we add attributes to make it open in another tab?

```html
<a href="..." target="_blank" rel="noopener">...</a>
```

kandarpksk (approved):

docs/conf.py
melissawm: Sure, let me fix it

psobolewskiPhD:
Where should this announcement appear?
Here's what I see in Safari using the docs artifact from CI:
image

Edit: I think the `napari-sphinx-theme` repo needs to make a release for this to work?

melissawm:
For this to work, we need a new release of the napari sphinx theme (@codemonkey800 is working on this right now :)

Czaki:
As I understand, this will only appear in dev docs, not the stable one?

jni (changes requested):
@melissawm @kandarpksk can we have more details here? How long do you intend for this to be up? Does this put the banner on the home page or every single page on napari.org?

I also think at a minimum that the wording should clearly label the survey as a CZI survey.

I'd like to leave a bit of time for more core devs to chime in here, too.

jni:
@Czaki I *think* the fact that it's a sphinx config option is in order to show it on all pages.

Czaki:
> @Czaki I _think_ the fact that it's a sphinx config option is in order to show it on all pages.

but as I understand we only build dev docs from this repository.

jni:
> but as I understand we only build dev docs from this repository.

🤷 well anyway, that's why I was asking for more details LOL

alisterburt:
+1 - this isn't a napari community survey, it's a CZI survey about napari. The survey should not be called the napari community survey.

brisvag (changes requested):
> Adds an announcement banner to the docs pointing to the napari survey.

I don't love this way of proposing to add an advertisement to a _CZI_ initiative to the website. It's quite irritating to find out that a whole suite of PRs aimed at this went by without any non-CZI core-dev being in the loop.

At a minimum:
1. clearly state it's from CZI
2. clarify that data is not collected by, shared with, or analysed by the community and/or core devs.
3. I can't tell from just the PRs how the banner looks like, but definitely move it to a place where it's not obnoxious (not the top/bottom of the page, but maybe an expandable notification button or something similarly out-of-the-way)

Then, we can discuss whether we want this banner to be on the napari website. Either way, considering point 2., I don't think this banner should be on the community website at all.

tlambert03:
yeah, this is a bit bs

kandarpksk:
Hi core devs,

Apologies for the confusion; let me start by sharing some context: Lucy Obus, the original UX researcher on the team, has been running a survey in the last two years to benchmark and measure napari user sentiment and community makeup, among other things. Last year, the survey was announced on the napari home page: [napari/napari.github.io#300](https://github.com/napari/napari.github.io/pull/300).

With that as the starting point, I refreshed the survey for 2022:
• This year’s survey has eliminated some questions and reworded others for clarity, but otherwise remains similar to the last two years, focusing on core napari user experience interests (we do have a short section on acquiring data).
• There are no questions evaluating CZI efforts/impact/brand.
• The introduction page of the survey only mentions CZI as a precursor to legalese on data privacy.

I sent a preview of the survey with a request for feedback to the napari steering council on November 30. No feedback was offered, however, I did not realize I should have better engaged, and with the entire group of core developers. **Here’s a [preview](https://chanzuck.co1.qualtrics.com/jfe/preview/previewId/531ba870-5455-44de-8f32-075c55965822/SV_6hTsV7i6dhSrE7Y?Q_CHL=preview&Q_SurveyVersionID=current) of the survey (responses are _not_ recorded via this link). Please email any feedback to [me](mailto:kkhandwala@chanzuckerberg.com) or add to this [sheet](https://docs.google.com/spreadsheets/d/1mjdN7l2G_xZoeIp03ixHP_GC_79Gm14td0oBuzdY6mM/edit#gid=513654450).**

The survey is planned to be live until January 16 (so about a month from now), but may close earlier/later depending on participation rates. The wording of the proposed banner is a shortened version of last year’s announcement (to better fit the banner element), and it would appear below the header on all pages of napari.org. Here is a preview of the how the banner would look:
banner-example

I plan to share results as Lucy has done before ([2021 share-out](https://docs.google.com/presentation/d/16lR7-b9PKUZM_tIbZDDRY41UxpJ_ar93cDHN8fp-hLQ/edit#slide=id.g116e22da033_0_27) for example), and engage closely during the analysis phase (just as I will with colleagues at CZI) – I will be sure to address questions and requests for further analysis.

I hope this starts to address your concerns, and will gladly take **guidance on what would be appropriate ways to share out the survey**. Again, I personally thought it would be fine because of last year, but I understand now that the circumstances and implementation are different.

Kandarp

PS: As you explore the preview, note that users with 6 months of experience or less will see two additional questions related to onboarding.

jni:
Hi all,

I've had an offline discussion with Kandarp about this. It would have been good to have time to hash this out properly, but unfortunately it fell into the Christmas holidays. Because the survey is/was time sensitive, there's no point going through with the review here. Additionally, because of the way we build the site, this PR would only affect the `/dev/` site, which is not the default site that people navigate to, so I expect that this PR would not have driven that much traffic to the survey anyway.

Over the coming months, we need to discuss:
- when we have such announcements, do we want (1) a banner on the home page, (2) a banner on every page, or (3) an announcement elsewhere on the home page or a dedicated announcements page? (I favour 1+3, and this would work well with the idea of a home page that is updated independently of the versioned docs.)
- what kind of announcements do we want on the home page? We should figure out some guidelines for this.
- what level of involvement do the core devs want in developing the survey, if any, and is CZI comfortable with that involvement? Is there a case in which we want to develop our own independent survey that we control?

There's probably other questions I'm not considering here. But, we should make separate issues for these, and close this PR in the meantime. Thanks everyone for your consideration.

PR #83, opened 2022-12-28
Proposing hybrid npe1-npe2 plugins

haesleinhuepf:
Hi all,

I just read NAP-6. Fantastic work @DragaDoncila et Al! Thanks working tthis.

Just a minor proposal from my side: consider enabling plugins using the napari-tools-menu (based on npe1) and menus using npe2 at the same time - for a temporary period. It would harm users otherwise if they are forced to do the update of their workflows instantly: imagine suddenly all menus are in different places and you can't find things anymore. It's an aspect of backwards compatibility and thus, I'm adding this to that section.

Let me know what you think! Thanks again.

Best,
Robert

### DIFF
--- a/docs/naps/6-contributable-menus.md
+++ b/docs/naps/6-contributable-menus.md
@@ -409,6 +409,11 @@ backward compatibility. -->
 This work does not have any backward compatibility considerations for
 existing features.
 
+To make the transition manageable for plugin-developers who used the 
+napari-tools-menu to structure their menus, it would be good to enable hybrid
+npe1-npe2 plugins for an intermediate period. 
+[See also #4708](https://github.com/napari/napari/issues/4708)
+
 In future, backward compatibility concerns could arise when a menu name/ID is
 changed in `napari`, or when `napari` removes a menu that was previously 
 contributable.


Czaki:
@haesleinhuepf Why can your plugin not dynamically generate `napari.yaml` file based on another plugin that uses your API? 

It may need to perform an additional restart after installing the plugin, but it will benefit from not importing code until it is required, which allows  to not slowdown napari start.

haesleinhuepf:
> @haesleinhuepf Why can your plugin not dynamically generate `napari.yaml` file based on another plugin that uses your API?

I could create that yaml (manually) but it wouldn't work with menus. the last version of napari that supported npe1 and npe2 was 0.4.15. my goal is to enable a smooth transition between the tools menu (which is based on npe1, and still works in 0.4.17) and the new menu that is described in nap6.

psobolewskiPhD:
I'm not sure I fully understand: is the idea to have plugins provide duplicated menu items, i.e. in both places simultaneously? If so, as a user, I'm not sure this is such a good idea because the duplications would confuse me—which ones should I use? should I be learning new locations? why learn something new if the old still works? 
Also, what about a new user just picking up the plugin in this "hybrid state" that would be quite the UI disaster for them.
I think it's best to make a clean break with a new more-major version release and document the new menu locations. Yes, this requires a bit of relearning, but it makes everything clearer.

Czaki:
> I could create that yaml (manually) but it wouldn't work with menus. the last version of napari that supported npe1 and npe2 was 0.4.15. my goal is to enable a smooth transition between the tools menu (which is based on npe1, and still works in 0.4.17) and the new menu that is described in nap6.

Maybe I do not know everything, but I was pretty sure that current tool menu is produced by your plugin, not by multiple plugins itself? 

Or there are more than one plugins that generates menus?

haesleinhuepf:
> duplicated menu items,

For a temporary period yes. 

More importantly: this would allow plugins that are backwards compatible. They run in napari < 0.4.15 and > 0.4.17

> should I be learning new locations? why learn something new if the old still works?

Some of us also have to work with napari. Let them decide when to learn new menu locations and do not force them.

> Also, what about a new user just picking up the plugin in this "hybrid state" that would be quite the UI disaster for them.

Using deprecation warnings, a well established mechanism in the field.

> I think it's best to make a clean break

Are you ready to make many users unhappy? This is what breaking backwards compatibility without prewarning typically causes.

haesleinhuepf:
> I was pretty sure that current tool menu is produced by your plugin, not by multiple plugins itself?
> 
> Or there are more than one plugins that generates menus?

I certainly have no full overview but I was able to [print out some stats](https://github.com/haesleinhuepf/napari-tools-menu/tree/print_stats) on my computer. Here are 22 plugins with in total about 230 menu entries installed.

* 'napari_simpleitk_image_processing', 
* 'napari_cupy_image_processing', 
* 'napari_segment_blobs_and_things_with_membranes', 
* 'napari_pyclesperanto_assistant',
* 'napari_accelerated_pixel_and_object_classification', 
* 'the_segmentation_game', 
* 'napari_clusters_plotter', 
* 'napari_skimage_regionprops', 
* 'napari_plot_profile', 
* 'napari_process_points_and_surfaces', 
* 'devbio_napari', 
* 'napari_assistant', 
* 'napari_time_slicer', 
* 'napari_crop', 
* 'napari_folder_browser', 
* 'napari_layer_details_display', 
* 'napari_mouse_controls', 
* 'napari_tabu', 
* 'napari_plugin_search', 
* 'napari_workflow_inspector', 
* 'napari_workflow_optimizer', 
* 'napari_brightness_contrast'

Edit: more plugins I'm aware of:
* 'napari-owncloud'
* 'napari-dexp'
* 'napari-aydin'
* 'napari-n2v'
* 'natari'

Of course, most of these plugins are maintained by myself. And I'm happy to transition to npe2-based menus. But _please_ give me some time for doing this and allow the users to have a smooth transition. All we need to fix is this one bug:
* https://github.com/napari/napari/issues/4708

alisterburt:
Given the user base of the existing tool menu I am okay with a slower transition here with temporary enablement of tools as long as it is understood that this menu will be deprecated... proper break could be 0.5 for example

haesleinhuepf:
Awesome, thanks for the support @alisterburt ! ❤️💚💙

> as long as it is understood that this menu will be deprecated...

On the [napari-tools-menu readme](https://github.com/haesleinhuepf/napari-tools-menu) it says:

"Note: This implementation is based on a [monkey patch](https://en.wikipedia.org/wiki/Monkey_patch) of napari, tested with napari 0.4.11. Thus, it might stop working with a future version of napari, e.g. when the https://github.com/napari/napari/issues/3115 is finished. For now it is a nice workaround to ease the life of end-users."

So yes, this was always the plan. Unfortunately, I cannot transition towards npe2 yet, because menus don't work there. That's why this discussion here.

Thanks again!

Czaki:
> Of course, most of these plugins are maintained by myself. And I'm happy to transition to npe2-based menus. But _please_ give me some time for doing this and allow the users to have a smooth transition. All we need to fix is this one bug:

I think that You do not understand my proposition. 
The `napari.yaml` is just a plain file that is a reader from the disc. So there could be a "monkeypatch" again. 

The `napari-tool-menu` could rewrite (after installation) `napari.yaml` based on environment content and provide all tool menu entries as their entries.

haesleinhuepf:
> So there could be a "monkeypatch" again

Please read https://github.com/napari/napari/issues/4708 . npe1 and npe2 cannot coexist since napari 0.4.16. The existence of a napari.yaml file kills the npe1-compatibility, the monkey patch and the Tools menu.

Edit: also feel free to propose technical solutions in the GitHub issue. Let's focus our discussion here on the goal of achieving backwards compatibility (or not).

Czaki:
> Please read [napari/napari#4708](https://github.com/napari/napari/issues/4708) . npe1 and npe2 cannot coexist since napari 0.4.16. The existence of a napari.yaml file kills the npe1-compatibility, the monkey patch and the Tools menu.

Yes, but the new tools menu build as described in this NAP could generate `napari.yaml`.

 

> Let's focus our discussion here on the goal of achieving backwards compatibility (or not).

Currently, I do not see a core developer who could spend time implementing and maintaining such a feature. The main problem with npe1 is a big impact on the startup of the application which generates a bad user experience. So we cannot just simply reenable npe1.

haesleinhuepf:
> Currently, I do not see a core developer who could spend time implementing and maintaining such a feature.

I'm happy to do it. I just need guidance and an agreement that it will be merged once it's implemented. It's anyway just reverting changes made between napari 0.4.15 and 0.4.16

haesleinhuepf:
Again: let's focus the discussion here one this question: Do napari-core developers want to have the new npe2-based menus temporarily backwards-compatible with npe1-based Tools menu? Yes or no?

jni (changes requested):
My personal preference is to give you time @haesleinhuepf *using the main branch*. ie we merge the napari.yaml-based implementation at some point and then ~1-2 months later we release. At any rate the tools menu already pins napari to 0.4.15, doesn't it? So one release more or less is imho not a huge deal? Apologies if I misunderstand something here.

The truth is that in the core team we mostly feel that it's been a burden *already* to maintain npe1 compatibility by itself, without even accounting for (a) dual plugins, and (b) supporting features that we never even supported! 😂 If it was exactly *free* to support hybrid plugins, I'd be fine with it, but it's not, it requires significant design and engineering effort, that I think would be better served by improving npe2 capabilities.

haesleinhuepf:
> At any rate the tools menu already pins napari to 0.4.15, doesn't it?

No, we downgraded all formerly hybrid npe1-npe2 plugins to npe1 plugins, because npe2 doesn't support menus yet 



> My personal preference is to give you time @haesleinhuepf _using the main branch_. ie we merge the napari.yaml-based implementation at some point and then ~1-2 months later we release.

That means we cannot have plugins that work in napari 0.4.x with the Tools menu AND in napari 0.5.x with the npe2-based menu. Thus, you break backwards compatibility for the users without prewarning. Without any chance of a smooth transition. Bad idea from my perspective 



> The truth is that in the core team we mostly feel that it's been a burden _already_ to maintain npe1 compatibility

I'm not asking for full npe1 compatibility, I'm asking for compatibility with the npe1-based Tools menu through hybrid npe1-npe2 based plugins.

I could imagine the following works: If napari 0.4.18 runs in an environment where `napari-plugin-engine` aka npe1 is NOT installed, it will not support the Tools menu. Thus people with only npe2 plugins installed can happily use napari with all the great new features.
In case npe1 is installed, npe1-based plugins get the chance to register their Tools menus. I think I would know how to implement that. Again, before spending time on this I need an agreemet with the core-developers that this would be a reasonable compromise with a certain chance to get merged.

Czaki:
Any solution that could be accepted cannot be executed on napari startup. Could not impact napari startup time. 
The napari startup's slowdown caused the transition from npe1 to npe2. As you mentioned, the support of a wide range napari versions is important, so multiple plugins still are npe1 but benefit from the npe1-to-npe2 adaptor. With Your proposition, all of them will be loaded on the napari start-up again. And this will be treated as napari, not plugins fault. 

Because of this, we trying to suggest that any such change needs to be able to cache information between sessions. Maybe we should add some options to handle information about installing/uninstalling plugins through the plugin dialog or something different. But nothing that will be executed every startup.

jni:
> > At any rate the tools menu already pins napari to 0.4.15, doesn't it?
> 
> No,

Oh. Weird, I must have it confused with another plugin. Sorry for the misunderstanding.

> That means we cannot have plugins that work in napari 0.4.x with the Tools menu AND in napari 0.5.x with the npe2-based menu. Thus, you break backwards compatibility for the users without prewarning. Without any chance of a smooth transition. Bad idea from my perspective

Leaving aside the fact that napari is still technically in alpha and 0.x versions, napari itself has never supported a Tools menu. Thus I do not see it as our responsibility to maintain backwards compatibility for this.

> Again, before spending time on this I need an agreement with the core-developers that this would be a reasonable compromise with a certain chance to get merged.

Unfortunately it's very hard to commit to this without seeing the solution, but I am not optimistic about it. Honestly the core team is overwhelmed by the existing maintenance burden and we are reluctant to take any additional burden, to support something that was not supported in the first place. This is not to point any fingers at a nebulous "core team" — I fully own this position, I am merely saying that my feeling is that it is the majority position.

From our perspective, we have had a significant period in which we supported both npe1 and npe2 plugins with feature parity between the two. As you point out, users of the Tools menu haven't had such a luxury, but frankly, I consider that a danger that comes with the territory of monkey patching instead of contributing upstream.

With regards to supporting plugins that depend on Tools menu, I honestly think that, rather than implementing hybrid plugins, your time is better spent participating in the spec design and actually migrating the plugins (once NAP-6 is fleshed out), and our time is better spent improving napari in the thousand ways it needs improving. Then when both napari 0.5 (or whatever) is released, hopefully all your plugins have been migrated and users simply need to upgrade everything. And if you haven't had a chance, then your users can stay on older versions of things temporarily — it's not as if they have zero working options.

Anyway, to summarise: yes, in the world of infinite developer time, ideally we always maintain backwards compatibility with everything in the ecosystem. But we have severely **finite** developer time, and are *already* at the point where maintenance cost exceeds available developer time, so we are aiming to prioritise ongoing development rather than extensive backwards compatibility. (At least, that is my reading of the core dev "temperature" from recent meetings.)

I'm inclined to close this, but I'll let a few more people vote on your ["poll"](https://github.com/napari/docs/pull/83#issuecomment-1369199335) before doing that.

haesleinhuepf:
> Thus I do not see it as our responsibility to maintain backwards compatibility for this.

Again, I'm happy to do this. I presume/hope it's a three-liner.

brisvag:
> Again, before spending time on this I need an agreemet with the core-developers that this would be a reasonable compromise with a certain chance to get merged.

> Again, I'm happy to do this. I presume/hope it's a three-liner.

I'm not sure even you believe it's just three lines @haesleinhuepf :P But if it is, and does not grind napari startup to a halt, I'm sure we can consider hitting the merge button :) However, we definitely don't want to be stuck in the limbo of "supporting" the monekypatch after `0.5`.

But in general, I agree with @jni that our (and your) time is better spent actually migrating napari and those plugins to a functional state.

DragaDoncila:
I would strongly hesitate to merge a PR that introduces new support for `napari-plugin-engine` plugins and other tools that use its entry point. Removing all `npe1` supporting code is a very high priority for me - building around it has been a significant burden over the last 6 to 8 months. I therefore would see this change as a step backwards. 

I agree that migrating plugins can take time, and I'm sympathetic to that. I want to make these menus available for contribution as soon as possible, so that plugin developers can start migrating, with our support. Auto-generating a manifest for an `npe1` plugin is trivial in most cases, and many of the plugins I looked at using the tools menu already had a manifest file and are registering their commands via the supported pathway. Ideally, with core dev support, we would see most plugins converted before the 0.5 release. It would also give us a great opportunity to further curate the menu options based on the kinds of contributions we see, before a major release where users will have to learn new menus.

haesleinhuepf:
> I therefore would see this change as a step backwards.

That's why the term "backwards compatibility" is used here. 

So you are ok breaking the Tools menu without prewarning users, without the chance to have it together with the new menu for a limited amount of time @DragaDoncila ? Users will be angry about that breaking change.

haesleinhuepf:
> Auto-generating a manifest for an `npe1` plugin is trivial in most cases, and many of the plugins I looked at using the tools menu already had a manifest file

Show me one.

Czaki:
@haesleinhuepf When the specification will be ready I will prepare a patch for `napari-tool-menu` that will handle this until the user will install `napari-tool-menu` in a writable directory.

DragaDoncila:
> That's why the term "backwards compatibility" is used here.

I don't see this as an issue of backwards compatibility. Support for `napari-tools-menu` has been, as I understand it, an accidental side effect of loading original plugin engine entry points.  Introducing code to explicitly load `npe1` plugins does not therefore seem like a backwards compatibility change to me.

Automated conversion for `npe1` plugins works for all plugins I've come across, so in my perspective, the `npe2` shimming mechanism provides backwards compatibility for original `npe1` plugins.

> So you are ok breaking the Tools menu without prewarning users, without the chance to have it together with the new menu for a limited amount of time @DragaDoncila ? Users will be angry about that breaking change.

I believe users will make use of the `Tools` menu until it disappears, at which point they will be prompted to learn the new menus. I do not therefore think that displaying both the `Tools` menu and the newly proposed menus is a useful migration strategy. It may be useful to update the `napari-tools-menu` to advise users of its incompatibility with upcoming `napari` versions though, if you believe migration prior to release is infeasible.

Whether this NAP gets released as part of 0.5 or a future release, we will give ourselves and users ample time to get the NAP approved, release menus for contribution, and reach out to plugin developers (yourself included) to assist with the migration process. We will of course also communicate this change across whatever channels we can, and I hope that plugin developers do the same, if they have known users who may be affected. In the worst case if a plugin is not 100% ready for migration come release time, they could advise users to hold off on updating napari until the migration process is complete.

haesleinhuepf:
> In the worst case if a plugin is not 100% ready for migration come release time, they could advise users to hold off on updating napari until the migration process is complete.

Do you suggest to pin napari < 0.5 in the affected plugins @DragaDoncila ?

Czaki:
> Do you suggest to pin napari < 0.5 in the affected plugins @DragaDoncila ?

At this moment tools menu works in napari dev (with unchecked npe2 adaptor). 

obraz

I do not know a plan for when to disable it. But I expect that it should work with 0.5.0, maybe pin below 0.6.0

haesleinhuepf:
Wow, I was not aware of this checkbox @Czaki. Does unticking it also fix https://github.com/napari/napari/issues/4708 ? That would be amazing!

Czaki:
You need to check it. I do not know.

DragaDoncila:
> Do you suggest to pin napari < 0.5 in the affected plugins @DragaDoncila ?

No, I would suggest they work on migrating to npe2 and advise users to continue using current version of napari if migration is not ready for release.

> I do not know a plan for when to disable it

The npe2 adaptor will be default as soon as we can make that happen, and that checkbox will be removed. This will be part of the refactor removing original plugin engine code. I would not plan to rely on this feature in the long term.

haesleinhuepf:
> No, I would suggest they work on migrating to npe2

We cannot migrate to npe2 because npe2 doesn't support menus yet. That's the point in this discussion @DragaDoncila .

DragaDoncila:
> We cannot migrate to npe2 because npe2 doesn't support menus yet. That's the point in this discussion @DragaDoncila .

Perhaps I was unclear, so my apologies. Here (very roughly) is what I envision a plugin developer relying on the `napari-tools-menu` would ideally do:

- Currently: Read the NAP if they're interested, suggest new menu items or different configurations. Consider where their widget contributions might go in these menus. Add a manifest file (even if not using the entry point) so that they're familiar with it and it's ready for menu contributions. `npe2 convert . --dry-runs` will print the manifest of an original plugin engine plugin, without changing anything.
- Once contributable menus are in main: Add menu contributions for widgets currently being shown in the `Plugins` menu and/or `Tools` menu. Give us any feedback they have with respect to the menus. Let users know (if they work closely with them) that the `Tools` menu will eventually not be there any more, and there will be new menus available.
- Once menus are released: If conversion to `npe2` was smooth and the new menu items have been added, nothing. If it's taking longer than expected or just haven't had the time to do so yet, reach out to us for support - if we haven't reached out to you already. 
- Once `npe1` code is fully deprecated/the npe2 adaptor is the only choice: If you were still using the `Tools` menu, this is when it would no longer work with new napari versions. In this case, you should advise users to stick to the current napari version and not update. If that means pinning, then do so, though I would advise migrating as soon as possible.

Again, this is my ideal scenario. I understand real life may not play out so smoothly and I'm ready to put in the effort with plugin devs to work out the wrinkles and give everyone the best possible chance of having their plugin ready to go with the new menu system before the `Tools` menu is no longer an option. 

I also get that the timeline right now is a bit vague - when will menus be released? When will npe1 code be fully deprecated? This is a matter of current discussion and I will be sure to update the NAP with a more concrete expected timeline when we have one.

haesleinhuepf:
Great plan @DragaDoncila! I'm just afraid this will lead to plugins that are not compatible with napari <= 0.4.17 AND > 0.4.17. In one of both the menu won't work. That's why I'm proposing to reenable hybrid npe1-npe2 support for a limited amount of transition period. It will make users unhappy otherwise. And user happiness is a high goal on my agenda.

jni:
> Thus, you break backwards compatibility for the users without prewarning. Without any chance of a smooth transition. Bad idea from my perspective

> Users will be angry about that breaking change.

> It will make users unhappy otherwise. And user happiness is a high goal on my agenda.

@haesleinhuepf if this is true then I strongly suggest that next time you work with the broader community to develop a feature that is officially supported, rather than monkey patching and relying on private attributes because you are, in your own words, impatient. In my opinion those approaches are fine to experiment with and to make proofs of concept, but not to build user and developer communities.

Frankly I am disappointed and even angry at the implication that we don't care about our users if we break a completely unsupported feature that was outside of our control from day 1. (And I don't anger easily! This is honestly a disconcerting experience for me!) *You* built a ton of plugins and a community around an unsupported monkey patch, and are now trying to pass on responsibility of that hack to us.

I genuinely think you will have time to migrate your plugins in order to do a "simul-release" or similar, but if you don't, and users indeed get angry ("why can't I install napari-devbio on napari 0.5?"), feel free to tell them we chose to make a clean break, but please also include the fact that you built it all based on an unsupported monkey patch and this was always the expected outcome. (As you wrote in your readme.)

In the meantime, I feel that this discussion has gone around in circles long enough, and based on the overall core dev sentiment (such as in your poll), I am going to close it. As mentioned multiple times, we will endeavour to give you an easy migration path, and plenty of lead time before a release, but we will do so in the main branch.

And by the way, I am genuinely sorry about the way this has turned out and hope we can have a friendly debrief over a beer someday.

tlambert03:
I whole heartedly agree with @jni's sentiments, and would gently remind you (and request that you uphold) your own comments [here](https://github.com/chanzuckerberg/napari-hub/discussions/299#discussioncomment-1528734):

> Anyway, my (teams) mission focuses on what non-coders need, as we have many of them on campus, and thus, we are going in that direction even though napari is still experimental. This train is gaining speed and we can't stop it. We are fully aware that things may break and are happy to fix them. (Did you know that Fiji breaks all the time?😉)

here's your chance to be happy about fixing it ;)

haesleinhuepf:
I would have implemented this backwards compatibility issue discussed here, if napari core-developers would have agreed that we should do it. Unfortunately, there is no agreement. I will find another way. Don't worry. Thanks everyone for the constructive discussion!

PR #51, opened 2022-11-18
Added info for conda installation problems

Chris-N-K:
# Description
As in napari/napari#5345 at the `napari` main repository discussed I added a small info box for the case that the `conda` installation does not work propperly.
In this case `mamba` is a good alternative.

## Type of change

- [ ] Bug-fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# References
@psobolewskiPhD recommended `mamba`

# How has this been tested?

- [ ] example: the test suite for my feature covers cases x, y, and z
- [ ] example: all tests pass with my change

## Final checklist:
- [x] My PR is the minimum possible work for the desired functionality
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] If I included new strings, I have used `trans.` to make them localizable.
      For more information see our [translations guide](https://napari.org/docs/dev/guides/translations.html).

@jaimergp

PR #74, opened 2022-12-15
Update Bundled App Installation Instructions

jaimergp:
# Description

Add detailed installation instructions for the new conda-based bundled app, updated with screenshots from 0.5.0+.

# References

Closes #67

This PR brings napari/napari#4383 over to `napari/docs`. It's not a perfect port, and I had to edit some patches, but it keeps most of the original history.

Inspired by https://github.com/napari/napari/pull/5240#issuecomment-1282076004, I followed these steps:

1. Obtain the commit-by-commit patch from the original PR:

```bash
$ cd ~/devel/napari/napari
$ gh pr checkout 4383
$ gh pr diff --patch > ~/tmp/napari-4383.patch
```

2. Apply patch interactively on `napari/docs`:

```bash
$ cd ~/devel/napari/docs
# Create new branch from up-to-date main
$ git checkout main
$ git pull upstream main
$ git checkout -b bring-napari-4383
# Apply patch
$ git am ~/tmp/napari-4383.patch
```

Sadly, the naive `git am` command with no flags didn't work, so I had to do it "interactively" with `-i --rejected`, which iteratively creates `*.rej` files with the rejected chunks of each patch:

```bash
$ git am -i --rejected ~/tmp/napari-4383.patch
# analyze rejection and modify target file; then:
$ git am -i --rejected --continue
# rinse and repeat til you are done
```

---

Issue #67, opened 2022-12-11
Updating napari's Bundled App Installation Instructions


melissawm:
This is a follow up to napari/napari#4383. Please see the discussion there.

psobolewskiPhD:
As noted in the PR discussion, I think this is really important.
I would argue that we need a top top level tab (like Usage, Community, Plugins) for Installation. Then the Getting started, etc. guides and tutorials could focus on Usage.

jaimergp:
I can work on this!

jaimergp:
@psobolewskiPhD can I get this assigned? I cannot self-assign on this repo 😬

psobolewskiPhD:
@jaimergp you got it! Let me know if I can help!

jaimergp:
Opened https://github.com/napari/docs/pull/74 for now. Let's see if my git fu is good enough :D

PR #84, opened 2023-01-05
Feature: implement python version substitution in conf.py

psobolewskiPhD:
# Description
This PR is a leftover from https://github.com/napari/docs/pull/51
The idea was from @Czaki https://github.com/Chris-N-K/napari-docs/pull/1#issuecomment-1336260259 
Basically, the python version shown in various places should be globally set in `conf.py` once and then substitutions used.

At the moment, I've set the python version to 3.9, but I at some point we should consider switching to 3.10.

## Type of change

# References

## Final checklist:

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -171,7 +171,7 @@ myst_substitutions = {
    "python_version_range": python_version_range,
    "python_minimum_version": python_minimum_version,
    "python_version_code": f"`python={python_version}`",
-   "conda_env_python_version": f"```sh\nconda create -y -n napari-env -c conda-forge python={python_version}\nconda activate napari-env\n```",
+   "conda_create_env": f"```sh\nconda create -y -n napari-env -c conda-forge python={python_version}\nconda activate napari-env\n```",
 }
 
 nb_output_stderr = 'show'
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -160,11 +160,15 @@ myst_enable_extensions = [
 myst_heading_anchors = 3
 
 version_string = '.'.join(str(x) for x in __version_tuple__[:3])
+python_version = '3.9'
 
 myst_substitutions = {
    "napari_conda_version": f"`napari={version_string}`",
    "napari_version": version_string,
- }
+   "python_version": python_version,
+   "python_version_code": f"`python={python_version}`",
+   "conda_env_python_version": f"```sh\nconda create -y -n napari-env -c conda-forge python={python_version}\nconda activate napari-env\n```",
+}
 
 nb_output_stderr = 'show'
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -161,13 +161,15 @@ myst_heading_anchors = 3
 
 version_string = '.'.join(str(x) for x in __version_tuple__[:3])
 python_version = '3.9'
-python_version_range = '3.9–3.10'
+python_version_range = '3.8–3.10'
+python_minimum_version = '3.8'
 
 myst_substitutions = {
    "napari_conda_version": f"`napari={version_string}`",
    "napari_version": version_string,
    "python_version": python_version,
    "python_version_range": python_version_range,
+   "python_minimum_version": python_minimum_version,
    "python_version_code": f"`python={python_version}`",
    "conda_env_python_version": f"```sh\nconda create -y -n napari-env -c conda-forge python={python_version}\nconda activate napari-env\n```",
 }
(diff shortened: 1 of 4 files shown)


docs/tutorials/fundamentals/installation.md
```diff
@@ -79,7 +76,7 @@ Choose one of the options below to install napari as a Python package.
 :class: dropdown
 
 napari can be installed on most macOS, Linux, and Windows systems with Python
-3.7, 3.8, and 3.9 using pip:
+3.8, 3.9, and 3.10 using pip:
```
Czaki: Could we also move the list of supported python to conf.py?

psobolewskiPhD:
@Czaki I added `python_minimum_version` and `python_version_range` as substitutions and I think I've now got the substitutions used in all the places the python version was mentioned in docs, outside of release notes and such.

melissawm (approved):
This looks good to me. I know this may be out of scope for this PR, but while we're at it, could we also bump the version mentioned in the quickstart guide? Currently it says

```
This tutorial uses napari 0.4.14. 
```

Czaki:
I have found these places that for me should be addressed in this PR:

https://github.com/napari/docs/blob/d1c287825d065a3d7511979f97f9194d63d6da7d/docs/tutorials/fundamentals/installation.md?plain=1#L215-L216

https://github.com/napari/docs/blob/d1c287825d065a3d7511979f97f9194d63d6da7d/docs/tutorials/fundamentals/installation.md?plain=1#L256-L258

https://github.com/napari/docs/blob/d1c287825d065a3d7511979f97f9194d63d6da7d/docs/tutorials/fundamentals/quick_start.md?plain=1#L27

https://github.com/napari/docs/blob/d1c287825d065a3d7511979f97f9194d63d6da7d/docs/tutorials/fundamentals/quick_start.md?plain=1#L53-L55

psobolewskiPhD:
@Czaki I saw those—are any of them still actual? 
I think the whole bundle is in flux to the conda bundle?
🤷‍♂️

Czaki:
The content of installation.md is connected with artifacts from https://github.com/napari/napari/blob/1eb5e3142063630a49dd30705c9aa2ff3cc1866b/.github/workflows/make_bundle.yml - it is pure napari repository. 

We still have workflow for conda build https://github.com/napari/napari/blob/1eb5e3142063630a49dd30705c9aa2ff3cc1866b/.github/workflows/make_bundle_conda.yml

So when looking at the current state of the repository, both are actual. And I do not see an Issue or PR for removing this now.

psobolewskiPhD:
> This looks good to me. I know this may be out of scope for this PR, but while we're at it, could we also bump the version mentioned in the quickstart guide? Currently it says
> 
> ```
> This tutorial uses napari 0.4.14. 
> ```

@melissawm I didn't change that because I'm not sure we want versions in guides/tutorials changing automatically when the variable is bumped in `conf.py`. It's probably a place where it's best to manually bump the number after ensuring the content matches the new napari version. Of course, in this case I doubt much changed between 0.4.14 and 0.4.17...

psobolewskiPhD:
@czaki I updated the quick_start links to use the substituted version to link directly to latest stable. However, looking at it more, I think the whole guide for the bundle app
https://napari.org/stable/tutorials/fundamentals/installation.html#install-as-a-bundled-app
should probably be updated—see https://github.com/napari/docs/issues/67
Looks like @jaimergp has a WIP PR for this https://github.com/napari/docs/pull/74

So I think this is out-of-scope of this PR, which is intended to provide the framework of substitutions. 

I did go ahead and add a link in the first paragraph to go directly to the current stable release using `napari_version`.

docs/conf.py
```diff
@@ -160,11 +160,19 @@
...
 
 myst_substitutions = {
    "napari_conda_version": f"`napari={version_string}`",
    "napari_version": version_string,
- }
+   "python_version": python_version,
+   "python_version_range": python_version_range,
+   "python_minimum_version": python_minimum_version,
+   "python_version_code": f"`python={python_version}`",
+   "conda_env_python_version": f"```sh\nconda create -y -n napari-env -c conda-forge python={python_version}\nconda activate napari-env\n```",
```
Czaki: Maybe change variable name to `conda_create_instruction`?

Czaki (approved):

docs/conf.py
psobolewskiPhD: I like that.
The other option is to move that substitution inline in the .md, instead of in conf.py.
I think something like this would work:
```
{{ "```sh\nconda create -y -n napari-env -c conda-forge python=PYTHON_VERION}\nconda activate napari-env\n```".replace(PYTHON_VERION, python_version }}
```
But those conda instructions are reused  3 times...

Czaki: But substitution declaration in the markdown file is mainly for things specific for a given file.

psobolewskiPhD: Agreed....
...and done.

PR #90, opened 2023-01-17
Move napari workshop template link to top of page

### LINKED ISSUE #88: Move link to the napari-workshop-template to the top of the page
## 📚 New content request

In https://napari.org/dev/further-resources/napari-workshops.html (discussed on the Docs working group meeting)

melissawm:
# Description
Moves information about the napari workshop template (and how to include workshops to the list) to the top of the Workshops page.

## Type of change

# References
Closes #88 

## Final checklist:

### DIFF
--- a/docs/further-resources/napari-workshops.md
+++ b/docs/further-resources/napari-workshops.md
@@ -7,7 +7,7 @@ This page contains a list, to make them easier to find online.
 You can find more videos of talks, tutorials and demos on the
 [napari youtube channel](https://www.youtube.com/channel/UCbTgw84ew4pxTJ9qu3W2hqg/playlists).
 
-```{note}
+```{admonition} Got a workshop?
 If you are looking to create your own workshop, you can use the
 [napari workshop template](https://github.com/napari/napari-workshop-template)
 as a starting point.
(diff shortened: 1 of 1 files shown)


psobolewskiPhD:
I'm of two minds about this.
An end-user would rather get to the materials, up-front. After all, this is the Usage menu.
A contributor/organizer doesn't care and wants to get to the instructions—this would even fit in Community>Contributing. In fact, maybe we should put this on a page there and link from Usage>Workshops?

BTW, what about someone wanting to announce an up-coming napari workshop/event?

melissawm:
I am not sure as we didn't discuss this case. Maybe @jni @DragaDoncila @chili-chiu have opinions?

DragaDoncila:
Hmm I guess I feel like it's not really a big enough section for a user to be "confused" about where to go? Yes a usage focused visitor might not care about the template/adding a workshop to the list, but it's also pretty easy for them to skim past this. And I also feel like it's too short to warrant its own page.

@psobolewskiPhD what if we keep the template & add your own text at the top, but we encase it in a little note or info admonition, to make sure it's even easier to skim past, and takes up less room so the list of workshops is more likely to be above the fold?

melissawm:
Here's a first try:

Screenshot_20230119_115642

psobolewskiPhD:
I like the note idea 🤷‍♂️

melissawm:
Ok, I'm pushing this up but if you prefer another kind of admonition we can also do that. Thanks for the feedback!

DragaDoncila:
I like the admonition! Is it possible to make it one with a custom title so instead of Note it says "Got a workshop?" or something similar to grab people's attention.

melissawm:
Here you go:

Screenshot_20230120_144235

DragaDoncila (approved):
Awesome, thanks @melissawm!

PR #100, opened 2023-01-30
Replace pip with python -m pip

dstansby:
# Description
This is best practice, to make sure the `pip` being picked up matches the current `python` executable.

## Type of change

# References

## Final checklist:

### DIFF
--- a/docs/developers/benchmarks.md
+++ b/docs/developers/benchmarks.md
@@ -16,7 +16,7 @@ development environment, then if using ``venv`` you may install the requirement 
 
 ```bash
 source napari-dev/bin/activate
-pip install asv
+python -m pip install asv
 ```
 
 If you are using conda, then the command:
@@ -162,14 +162,14 @@ Note that we in both these cases we have sent the output of the profiling to [sn
 which you can pip install with
 
 ```bash
-pip install snakeviz
+python -m pip install snakeviz
 ```
 
 and we use `--python=same` to profile against our current python environment.
 
 ## Running benchmarks on CI
 
-Benchmarking on CI has two main parts - the Benchmark Action and the Benchmark Reporting Action. 
+Benchmarking on CI has two main parts - the Benchmark Action and the Benchmark Reporting Action.
 
 ### The Benchmark Action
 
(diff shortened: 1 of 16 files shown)


Czaki (approved):

melissawm (approved):
+1!

psobolewskiPhD:
For my personal edification, can y'all give me a ELI5—or a link to read up?
🙏

Czaki:
> For my personal edification, can y'all give me a ELI5—or a link to read up?

In general, all operating systems are using environment variable named `PATH` to store information where binaries are stored. You could preview it by calling `echo $PATH` in the terminal. 

This path stores a list of directories that are checked in order from this variable to find where the variable is. After finding the first directory that contains the proper named executable with the proper name, the OS stops checking. 

But for performance reasons system also cache search result. 

So it may happen that, because of a mess in the `PATH` variable or some wrong working command, it may happen that calling `pip` may lead to running `pip` from a different environment than `python -m pip` (so different than python)

You may see a sample bug report here https://github.com/microsoft/vscode-python/issues/8852

brisvag:
To add to the above:

- calling `python -m modulename` is actually running `module.__main__.py`, which for pip is the command line interface
- you might even not have `pip` in the active environment, which would basically guarantee that  using `pip install` would run the system pip

brisvag (approved):

psobolewskiPhD:
Thanks guys!
So if I follow, the reason I've not (to my knowledge) run into this is because of using conda envs and conda installs python and pip into the env?

melissawm:
@psobolewskiPhD probably, but it is still recommended (and I have personally found this to be a problem using conda as well on one or two occasions)

Czaki:
> So if I follow, the reason I've not (to my knowledge) run into this is because of using conda envs and conda installs python and pip into the env?

maybe you directly execute instructions without using "shortcuts".

dstansby:
For what it's worth this is also what `pip` use in their docs: https://pip.pypa.io/en/stable/getting-started/

brisvag:
> So if I follow, the reason I've not (to my knowledge) run into this is because of using conda envs and conda installs python and pip into the env?

I generally use `pip` for brevity, so it's not a common issue. But when it happens it's a pain to debug unless you're familiar with it, so it's good to have this in the docs :)

PR #101, opened 2023-01-31
change blob example to be self contained

orena1:
# Description
Currently  for running the blob example, one need to import 2 other modules, they were indeed imported in the previous code, but I think it is better to have this example self-contained as this is a tutorial. Also, because the screenshot does not correspond to the output of running both examples one after the other. Thanks

## Type of change

## Final checklist:


Czaki (approved):
Thanks

psobolewskiPhD (approved):
Thanks for catching that! Definitely want the code blocks to be copy-paste-able.

Issue #98, opened 2023-01-27
Feature request: PR comments with build preview website link


psobolewskiPhD:
## 🧰 Task

PR comment with docs build preview ala numpy:
https://github.com/numpy/numpy.org/pull/596#issuecomment-1236103657
they use netlify: https://docs.netlify.com/site-deploys/notifications/#github-pull-request-comments

Looks like vercel also has something similar:
https://vercel.com/docs/concepts/deployments/comments/integrations

I'm not clear on the strengths and weaknesses of these, but it would be a **super** nice feature.
Disclaimer: my personal site is hosted by netlify.

melissawm:
Oh that is different though. For the numpy.org website we do use netlify but that's not for the documentation. The documentation site uses [Circle CI](https://github.com/numpy/numpy/blob/main/.circleci/config.yml). I proposed this for napari, and unfortunately I can't find the discussion anymore, but the preference was for using github actions. Unfortunately there are several issues with that workflow that I haven't come around to fixing yet. 

Related:
- https://github.com/napari/napari/issues/3572
- https://github.com/napari/napari/issues/5055
- https://github.com/napari/napari/pull/5014
- https://github.com/napari/napari/pull/5148

psobolewskiPhD:
The thing I like about the numpy thing is that it gives a link to just open the preview website in the browser, rather than downloading (another) docs.zip, unpacking, etc.
My internet is pretty slow, so downloading the docs.zip is a few minutes...

melissawm:
Circle CI also does that - include a link to a live preview of the docs. Here's an example:

https://output.circle-artifacts.com/output/job/1630c903-f0d1-48be-980f-59962ec0d254/artifacts/0/doc/build/html/index.html

This comes from https://github.com/numpy/numpy/pull/23092

See the entry point here
![Screenshot_2023-01-28-08-29-36-345-edit_com android chrome](https://user-images.githubusercontent.com/3949932/215264222-400b7adc-01db-4384-bd8a-fe59166c3f87.jpg)

psobolewskiPhD:
That PR is closed, but I went to another numpy/numpy one and I can't figure out where to see the build docs from the PR. 
The CicleCI `details` page, which seems the same as the one you put in the red box, doesn't show anything I can understand:
https://app.circleci.com/pipelines/github/numpy/numpy/17776/workflows/80b03de5-2ee0-4e74-b964-0815f6f63814/jobs/29828/steps

[...]

PR #104, opened 2023-02-10
CircleCI setup for PR documentation previews

melissawm:
# Description
Initial CircleCI set up to allow for PR documentation previews.

## Type of change

# References
Addresses #98 

## Final checklist:

### DIFF
--- a/.github/workflows/circleci.yml
+++ b/.github/workflows/circleci.yml
@@ -1,0 +1,23 @@
+# To enable this workflow on a fork, comment out:
+#
+# if: github.repository == 'napari/docs'
+
+name: CircleCI artifact redirector
+
+on: [status]
+jobs:
+  circleci_artifacts_redirector_job:
+    runs-on: ubuntu-latest
+    if: "github.event.context == 'ci/circleci: build-docs'"
+    permissions:
+      statuses: write
+    name: Run CircleCI artifacts redirector
+    # if: github.repository == 'napari/docs'
+    steps:
+      - name: GitHub Action step
+        uses: larsoner/circleci-artifacts-redirector-action@master
+        with:
+          repo-token: ${{ secrets.CIRCLECI_TOKEN }}
+          artifact-path: 0/docs/_build/index.html
+          circleci-jobs: build-docs
+          job-title: Check the rendered docs here!
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,70 @@
+# Use the latest 2.1 version of CircleCI pipeline process engine.
+# See: https://circleci.com/docs/2.0/configuration-reference
+version: 2.1
+
+# Orbs are reusable packages of CircleCI configuration that you may share across projects, enabling you to create encapsulated, parameterized commands, jobs, and executors that can be used across multiple projects.
+# See: https://circleci.com/docs/2.0/orb-intro/
+orbs:
+  # The python orb contains a set of prepackaged CircleCI configuration you can use repeatedly in your configuration files
+  # Orb commands and jobs help you with common scripting around a language/tool
+  # so you dont have to copy and paste it everywhere.
+  # See the orb documentation here: https://circleci.com/developer/orbs/orb/circleci/python
+  python: circleci/python@1.5.0
+
+# Define a job to be invoked later in a workflow.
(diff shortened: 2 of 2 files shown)


psobolewskiPhD:
OMG so excited for this! Let me know if I can help with anything!!
❤️

melissawm:
Ok! I got the first part - CircleCI is set up and you can get a status update with the link to the artifact!!! 🎉 

Please check it out here: https://github.com/melissawm/napari-docs/pull/3/

Note that this doesn't work on this PR, as the setup only works if the CircleCI job is already committed to the main branch. I did that to my own fork, so you can navigate to the PR above and click on the Details of the "Check the rendered docs here" check:

Screenshot_20230210_175119

This is a proof of concept, and there are no hard feelings if this doesn't get merged in the end 😄 I personally like this, but it needs a little more work because we would need to rewrite/replace the deploy_docs action here. Now that the basic details are set up I don't think it would be too difficult.

Last note is that this requires a github token with access to the repo, which I've done on my fork.

psobolewskiPhD:
Outstanding! I love the small Melissa icon (docs 👸) and the clear instructions on the action to click!
❤️

jni:
> it needs a little more work because we would need to rewrite/replace the deploy_docs action here.

Why? Is it just to not have to maintain lots of different CI jobs? If so I reckon just leave it, we can cross that bridge later.

> This is a proof of concept, and there are no hard feelings if this doesn't get merged in the end

Erm I wanna merge the thing! 😂

melissawm:
Lol that is good with me - yeah I was just thinking we could have shorter CI and less use if resources but we can definitely do that later. Let me remove from draft then 😁

.github/workflows/circleci.yml
```diff
@@ -0,0 +1,23 @@
...
+
+on: [status]
+jobs:
+  circleci_artifacts_redirector_job:
+    runs-on: ubuntu-latest
+    if: "github.event.context == 'ci/circleci: build-docs'"
+    permissions:
+      statuses: write
+    name: Run CircleCI artifacts redirector
+    # if: github.repository == 'napari/docs'
```
psobolewskiPhD, line 15: Should this line be changed to be un-commented before merging? Based on the line at the top of the file `to enable this workflow on a fork comment out:`

melissawm: Yeah - so if this is commented the action also runs on forks (the case now that I'm running this on my fork). We should probably remove this before merging. If you folks are ready for that (i.e. no other reviews are necessary) I can do that.

psobolewskiPhD: For me, the `code` seems correct and it works on your fork...
This doesn't directly affect the website or any end users, so I'd be game to merge sooner rather than later and then if need be tweak something? Dunno if that makes sense in the grand scheme of things—will defer to others with more experience, like you and @jni

Czaki: Why? By default, a contributor will not have circleCI configured, so the workflow will not be triggered. But could a contributor allow testing without creating PR? I would rather see a limited list of branches.

psobolewskiPhD: Wait, am I missing something? 
I understood this action would still run on PRs in this repo, and make a clickable link like the numpy PRs... 
See https://github.com/napari/docs/issues/98#issuecomment-1407379347
The contributor shouldn't need to do anything on their fork I don't think once this is in main anyways.

Czaki: Sometimes people work on their Fort until get a satisfying solution to not depend on the test queue of the napari organization.

psobolewskiPhD: Hmm, fair enough. But I think docs contributors may be less savvy and more like end-users, rather than developers, so this PR is a nice quality of life improvement where they can see the rendered docs online.

melissawm: It is my understanding this will only run in their forks if they do sign up with CircleCI - otherwise they won't have the appropriate credentials.

jni: I don't have an opinion here. It's easy enough for a contributor to edit that line... Is this ready other than this discussion?

Czaki (approved):

Czaki:
I will merge this as it is to enable it in the repository and observe it in the wild.

PR #108, opened 2023-02-17
Change `CIRCLECI_TOKEN` to `GITHUB_TOKEN` in `circleci.yml`  workflow file. 

### LINKED ISSUE #107: Lack of link to documentation preview
The circle CI introduced in #104 works, but the job of posting the link to the documentation preview does not work. 

See, for example, https://github.com/napari/docs/pull/105

obraz

It should be fixed. Maybe @melissawm, you will have a fast idea of what is missed.

Czaki:
# Description

Close #107
Switches to use the default GITHUB_TOKEN rather than making adding a new CIRCLECI_TOKEN
This makes the workflow match the README of the action:
[larsoner/circleci-artifacts-redirector-action](https://github.com/larsoner/circleci-artifacts-redirector-action)
This is also used commonly in other repos, see for example numpy/numpy:
[numpy/numpy@1426570/.github/workflows/circleci.yml#L18](https://github.com/numpy/numpy/blob/14265704f6fb3b77ca2739329d35ff49293d26df/.github/workflows/circleci.yml#L18)

## Type of change

# References

## Final checklist:

### DIFF
--- a/.github/workflows/circleci.yml
+++ b/.github/workflows/circleci.yml
@@ -17,7 +17,7 @@ jobs:
       - name: GitHub Action step
         uses: larsoner/circleci-artifacts-redirector-action@master
         with:
-          repo-token: ${{ secrets.CIRCLECI_TOKEN }}
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
           artifact-path: 0/docs/_build/index.html
           circleci-jobs: build-docs
           job-title: Check the rendered docs here!


melissawm:
I don't believe this will change anything since the token needs to be created first.

Czaki:
> I don't believe this will change anything since the token needs to be created first.

it will change nothing as the action is executed using the workflow version from the main branch, not from PR.

psobolewskiPhD (approved):
@czaki As I posted in the issue, changing the token name should work—all the repo settings seem fine. 
I think you can revert the other changes and update the description, maybe something like:
>Close https://github.com/napari/docs/issues/107
Switches to use the default GITHUB_TOKEN rather than making adding a new CIRCLECI_TOKEN
This makes the workflow match the README of the action:
https://github.com/larsoner/circleci-artifacts-redirector-action
This is also used commonly in other repos, see for example numpy/numpy:
https://github.com/numpy/numpy/blob/14265704f6fb3b77ca2739329d35ff49293d26df/.github/workflows/circleci.yml#L18

 Then we need to merge to see any effect.

Czaki:
As you can see, the change of token leads to a properly working workflow. 

https://github.com/Czaki/napari-docs/pull/1

Also I notice some problem in build process https://github.com/napari/napari/issues/5572

Czaki:
It also works here. You can see in #110.

melissawm:
Thanks, @Czaki !

PR #109, opened 2023-02-18
Update version switcher json

jni:
I have to say I don't understand why anything in the site version switcher is working currently given the status of the version switcher json... @melissawm Any thoughts? Is this version not uploaded to the deployed repo? At any rate, this makes it correct. (I added the version number to the "stable" name as I think it's handy.)

### DIFF
--- a/docs/_static/version_switcher.json
+++ b/docs/_static/version_switcher.json
@@ -5,9 +5,14 @@
 		"url": "https://napari.org/dev/"
 	},
 	{
-		"name": "stable",
-		"version": "0.4.16",
+		"name": "stable (0.4.17)",
+		"version": "0.4.17",
 		"url": "https://napari.org/stable/"
+	},
+	{
+		"name": "0.4.16",
+		"version": "0.4.16",
+		"url": "https://napari.org/0.4.16/"
 	},
 	{
 		"name": "0.4.15",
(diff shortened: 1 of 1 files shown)


Czaki:
it looks working because of andy PR https://github.com/napari/napari.github.io/pull/375

melissawm:
Yes - in fact I think this json here should be removed, as the [json being used](https://github.com/napari/docs/blob/main/docs/conf.py#L86) is the one on the napari.github.io repo. I think this was an oversight when we made the final decision on how to deploy the website.

jni:
I've thought about this a bit longer, and I would like us to half-reverse this decision: I think we should use the json from the website, but I think the json from the website should be deployed from this repo. The website itself should be a repository of derived products, not the source of anything — it becomes confusing when it is source for some things mixed in with derived products for other things.

What do you think @melissawm @Czaki?

jni:
... Looking at it more closely, this would simply mean we start using the version switcher url at `napari.org/dev/_static/version_switcher.json`. We could make the other value a symlink?

melissawm:
Yes - the only requirement is that this needs to be a live link for the json to work (meaning: you can't build with a local json file). Other than that, as long as we use a permalink this should work just fine :+1:

jni:
Good-oh. Should we merge this PR then? 😅 Then we can update the one on the website.

melissawm (approved):
LGTM!

PR #113, opened 2023-02-22
Add skeleton for constructor manager documentation

goanpeca:
# Description

## Type of change

# References

## Final checklist:


docs/developers/packaging.md
```diff
@@ -312,6 +309,125 @@ a high-level list of the main changes introduced in the stack.
 * Add signing for Windows
 * Add notarization for macOS PKG
 
+#### `constructor-manager` stack for handling updates and recovery
+
+Constructor installers come with a `constructor-manager` stack that allows users to perform
+"updates in place", recover from a broken installation, rollback to a previous version or restore
+point and finally uninstall the entire application. The constructor manager stack is application
+agnostic in that it can be used for any othe application using the updated constructor stack
```
aganders3, line 317:
```suggestion
agnostic in that it can be used for any other application using the updated constructor stack
```

docs/developers/packaging.md
```diff
@@ -312,6 +309,125 @@ a high-level list of the main changes introduced in the stack.
...
+
+![Constructor manager stack](images/constructor_manager_stack.png)
+
+This stack is composed of three main packages [19]:
+
+* `constructor-manager-backend`: Provides a simplified command line interface on top of
+  `conda`, `mamba` and `conda-lock` to query for updates and perform the actual installation,
+  reset or uninstallation of the application.
+* `constructor-manager-client`: Provides a Python API to interact with the backend and
+  perform the same operations as the CLI.
```
aganders3, line 328: This is a little confusing to me. I guess I would consider the Python API the backend and the CLI to be the client? You know more about the implementation though so I defer to you on naming.

docs/developers/packaging.md
```diff
@@ -312,6 +309,125 @@ a high-level list of the main changes introduced in the stack.
...
+* `restore`: similar to revert but restore to a previously found state of the current version.
+  This command could become a single one by providing the specific restore file (which is created
+  by conda lock).
+* `lock_environment`: create a lock file of the current state of the application environment.
+  This can be called by the client applications (using constructor-manager-client) to check
+  for changes in the environment. If no changes are detected, no lock is created.
+* `uninstall`: to be implemented
+* `get_status`: get information on a currently running update/restore/revert in progress.
+* `open_application`: open a give application created with `conda` and `menuinst` un a cross
+  platform way
```
aganders3, line 408: I don't think you need to repeat all this here. Maybe just list any limitations of this API vs. the backend/CLI.

docs/developers/packaging.md
```diff
@@ -332,4 +448,6 @@ a high-level list of the main changes introduced in the stack.
 [16]: https://github.com/conda-forge/napari-feedstock/pulls?q=is%3Apr+sort%3Aupdated-desc+is%3Aclosed
 [17]: https://anaconda.org/napari
 [18]: https://github.com/napari/packaging/issues/15
+[19]: https://github.com/napari/packaging
+[20]: https://github.com/conda/menuinst/tree/cep-devel/menuinst
```
aganders3, line 452: Might be a typo but I'd just link to the root of the tree where the README is.
```suggestion
[20]: https://github.com/conda/menuinst/tree/cep-devel
```

docs/developers/packaging.md
```diff
@@ -312,6 +309,125 @@ a high-level list of the main changes introduced in the stack.
...
+This library is meant to be used in the application environment by the applications themselves so
+they can query the information to know about new updates and either trigger them directly in a
+detached process (to be implemented in a following version) or to open the `constructor-manager-ui`
+to handle those actions directly.
+
+More information on the [README](https://github.com/goanpeca/packaging/blob/constructor-cli/constructor-manager/README.md) of the package.
+
+##### `constructor-manager-ui`
+
+which in its present form is a pyqt/pyside based application. It provides a graphical interface
```
aganders3, line 422:
```suggestion
In its present form this is a pyqt/pyside based application. It provides a graphical interface
```

aganders3 (commented):
Great improvements - thanks for working on this!

docs/developers/packaging.md
goanpeca: Hi @aganders3 yes indeed this might be confusing. Let me explain and from there we can make adjustments

The backend right now is only a CLI, that under the hood calls mamba, conda, conda-lock, and does a bunch of admin work for us (update in place, restore points etc). This could be updated eventually to become something more of a cli, for example an actual server doing things. So in this case instead of pinging a local server, we ping an actual program CLI.

The client is what we use to communicate with this backend (currently a cli but maybe a server in the future). So that is why I went with that naming scheme. The client could be renamed API. I called it like that in the past actually. But in this case is one thing we use to communicate and do thins with the "server" (like a frontend backend thing).

But I am open to suggestions, not super important but I would like to get the naming "right" before publishing packages

docs/developers/packaging.md
goanpeca: Indeed, this was a lot of repetition, will update!

docs/developers/packaging.md
goanpeca: Thanks. Yes, I think I copied the wrong link whoops!

docs/developers/packaging.md
goanpeca: I used to have the names

`constructor-manager-cli` and `constructor-manager-api` which were pretty accurate in what they are now, the new ones are more generic, maybe I should stick to those old names @aganders3 ?

aganders3: Thanks - this makes sense.

I agree the names are not super important because the descriptions make things clear. I think I do prefer `constructor-manager-cli` and `constructor-manager-api` but not by a wide margin, and the new names leave things more open to changes down the line. Maybe split the difference and keep `constructor-manager-backend` and `constructor-manager-api`?

goanpeca: I will go with `constructor-manager` and `constructor-manager-api`, that way we leave the cli to be more open and add a specific entrypoint in there `constructor-manager-cli` and leave the api one as the api that it is :) 

Thanks for the feedback

melissawm:
Hi all! Is this something we are still planning to do? If so, can I help push it forward? Thanks!

melissawm:
Gentle re-ping just to get a status on this 😄

aganders3:
From my perspective I think this is back-burner for the time being. I think more effort is being put into the plugin manager compared to the constructor manager (this project) for the time being. @goanpeca probably has a better idea at this point, though!

DragaDoncila:
We haven't prioritised the in-place update for a while -- more recent discussion on in-place update is [here](https://github.com/napari/packaging/issues/281). @goanpeca we're going to close this PR for now, until/unless we decide to pick up the idea of a bundle updater again.

PR #119, opened 2023-03-10
Add support for sphinx-issues

### LINKED ISSUE #169: PR # on release notes should be links
It would be nice if the PR # on release notes were links - see https://napari.org/release/release_0_4_13.html

melissawm:
# Description
This is a redo of napari/napari#4920

There is one caveat left: the [sphinx-issues extension](https://github.com/sloria/sphinx-issues) only accepts one repo as source of PRs or issues, so for PRs coming from the docs repo, for example, this shortcut doesn't work. I haven't found another extension that has the same functionality.

An alternative is to post full links and let [linkify/PyData Sphinx theme](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/theme-elements.html#link-shortening-for-git-repository-services) beautify the links. 

The dream: 
- Add support for multiple repos to sphinx-issues
- Render PR links created with sphinx-issues using linkify

Maybe in a follow up?

## Type of change

# References
Closes napari/docs#169

## Final checklist:


Czaki:
In the documentation is a such an example 
```
See :issue:`sloria/konch#45`.
```
 so maybe this could be an option to provide docs pr/issues?

melissawm:
@Czaki maybe you are right, I haven't tested it because there were no such PRs/issues in this current change, but thanks for pointing it out!

Czaki:
You reminded me that the release note script should be updated to handle also this repository.

jni:
Yeah, I have to say I prefer the full links option. The full links will work when rendered in markdown on GitHub, which is not true of this `{pr}` syntax.

I also wonder how hard it would be to write our own linkifier that matches GitHub's, based on the existing linkifier code...?

Czaki:
> I also wonder how hard it would be to write our own linkifier that matches GitHub's, based on the existing linkifier code...?

It should be easy (base on my experience with sphinx-qt-documentation). The hard part could be write tests. 

Or do you mean pre-commit check that will convert links in pre-commit steep?


> The full links will work when rendered in markdown on GitHub, which is not true of this {pr} syntax.

Why use `{pr}` instead of `:pr:`?

melissawm:
> Why use {pr} instead of :pr:?

The first is the MyST Markdown syntax we use, the second is for .rst files. [source](https://myst-parser.readthedocs.io/en/latest/syntax/roles-and-directives.html#roles-an-in-line-extension-point)

melissawm:
Ok! Based on this feedback I think this should be closed, and the approach should be to use full links and take advantage of the linkify extension.

> I also wonder how hard it would be to write our own linkifier that matches GitHub's, based on the existing linkifier code...?

@jni can you clarify what you think is missing from the [PyData/linkify functionality](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/theme-elements.html#link-shortening-for-git-repository-services)?

jni:
> @jni can you clarify what you think is missing from the [PyData/linkify functionality](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/theme-elements.html#link-shortening-for-git-repository-services)?

Yes! I want to be able to write `#<num>` and have that point to `https://github.com/napari/napari/issues/<num>` based on a sphinx configuration variable. And `<org/user>/<repo>#<num>` should always work by configuring GH as the default repo provider.

But, full links is ok for now! I'm going to close this one, thank you for experimenting @melissawm and also being aware of what is out there 😅!

melissawm:
Sounds good, I will continue exploring 🤓

PR #128, opened 2023-03-29
WIP: Make PR numbers into links in release notes

### LINKED ISSUE #169: PR # on release notes should be links

melissawm:
# Description

This is a first attempt at a "linkifyer" (still needs extra functionality for style and for PRs in other repos) 

## Type of change

# References
closes napari/docs#169

## Final checklist:

### DIFF
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -3,3 +3,11 @@
     <script async defer data-domain="napari.org" src="https://plausible.io/js/plausible.js"></script>
     {{ super() }}
 {% endblock %}
+
+{% block docs_body %}
+    {% if pagename.startswith('release/release_') %}
+        {{ body|linkify_prs() }}
+    {%- endif %}
+    {{ super() }}
+{% endblock %}
+
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -155,6 +155,7 @@ myst_enable_extensions = [
     'dollarmath',
     'substitution',
     'tasklist',
+    'linkify',
 ]
 
 myst_heading_anchors = 3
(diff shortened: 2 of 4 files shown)


melissawm:
I am not sure that this is worth the effort. It seems that for more recent release notes the links have been added consistently so do we still need to do this? Maybe the best option would be to just manually apply it once to older release notes so that all of them have links instead of overengineering 😄

jni:
Oh, interesting, I hadn't looked at this yet and I assumed it was indeed a onetime thing.

Personally, I kinda like the idea of linkifying *all* the docs — it makes writing the markdown easier. But I don't know how this interacts with already-linked references...?

Anyway, I agree that it's not as important as it used to be, because our release scripts are doing this automatically now.

Note, some earlier release notes (at least [0.4.18](https://napari.org/dev/release/release_0_4_18.html)), some of the links were `napari/napari/#9301` (note the extra `/`) — it would be good as part of this effort to harmonise them to just `#num`.

None of this is urgent, though — I'd say it's fine to keep it on the backburner.

psobolewskiPhD:
Looks like https://napari.org/stable/release/release_0_4_17.html and older doesn't have links, while all the more recent ones do, which is probably all that anyone cares about? 🤷

willingc:
@melissawm Should we close this since I think the release note generator in `release-tools` repo seems to do this if I am understanding the PR's purpose correctly. :sunny:

melissawm:
I think so! Maybe a follow up is to go back and edit old release botes, but again - not sure if that's worth the effort.

willingc:
I don't think that it's work the effort to update the old notes. I recommend closing @melissawm.

PR #132, opened 2023-04-06
Using Dask and Napari to view and process 100µm resolution human brain dataset

melissawm:
# Description 

This is a redo of napari/napari.github.io#64. I have tried to keep the original commit history. I think most comments there were addressed, but it probably needs a new round of reviews to make sure it is current and correct.

This is a tutorial on using Dask and Napari to view and process 100µm resolution human brain dataset.

cc @DaliaAymanAmeen

## Type of change

# References
Redo of napari/napari.github.io#64

## Final checklist:


psobolewskiPhD:
Wonder why the circleCI bit didn't run for this one...

docs/tutorials/processing/brain_dataset_tutorial.md
```diff
@@ -0,0 +1,73 @@
+# Using Dask and napari to view and process a high-resolution human brain imaging dataset.
+We used napari to view and process a high-resolution human brain imaging dataset, and in the following tutorial we will break down how it all builds up!
+
+The dataset was obtained from [Dr. Brian Edlow's Lab](https://www.comarecoverylab.org/) for NeuroImaging of Coma and Consciousness, which is dedicated to promoting recovery of consciousness in people with severe brain injuries.
+This dataset is a 100 micron resolution magnetic resonance imaging (MRI) scan of an ex vivo human brain specimen. The brain specimen was donated by a [58-year-old woman](https://www.youtube.com/watch?v=Q-9jzBkoNuI) who had no history of neurological disease and died of non-neurological causes.
+
+#### So how can we lazily load an image using dask?
+First, we import delayed from dask, which is the main function responsible for handling __Lazy Loading__. Here we defined a function *load_image()* that will lazy load the image data when called given a path to the data.
```
psobolewskiPhD:
```suggestion
First, we import delayed from dask, which is the main function responsible for handling __Lazy Loading__. Here we defined a function `load_image()` that will lazy load the image data when called given a path to the data. 
```

docs/tutorials/processing/brain_dataset_tutorial.md
```diff
@@ -0,0 +1,73 @@
...
+First, we import delayed from dask, which is the main function responsible for handling __Lazy Loading__. Here we defined a function *load_image()* that will lazy load the image data when called given a path to the data. 
+```python
+
+from dask import delayed
+import dask.array as da
+def load_image(image_name):
+	delayed_image_load = delayed(io.imread)
+	return da.from_delayed(delayed_image_load(image_name), dtype='float64', shape=(1760, 1760))
+```
+And here's a link to napari's ```napari+dask``` tutorial.
```
psobolewskiPhD: should link to `./dask.html`

docs/tutorials/processing/brain_dataset_tutorial.md
```diff
@@ -0,0 +1,73 @@
...
+#### Using dask.stack() to concatenate images
+
+`dask.array.stack` is used to stack many existing dask arrays into a new array, creating a new dimension (along axis=0 by default):
+```python
+images = da.stack(load_image(image_name) for image_name in images_names) 
+```
+
+
+
+Here's the complete code:
```
psobolewskiPhD: Pretty sure this can all be simplified using:
```
from dask_image.imread import imread

stack = imread("/path/to/experiment/*.tif")
```
As described here:
https://napari.org/dev/tutorials/processing/dask.html#make-your-life-easier-with-dask-image

docs/tutorials/processing/brain_dataset_tutorial.md
```diff
@@ -0,0 +1,73 @@
...
+    images_names = glob.glob(
+        'SYNTHESIZED_TIFF_Images_Raw/Synthesized_FLASH25_100um_TIFF_Axial_Images/Synthesized_FLASH25_Axial_*.tiff')
+    images_names = sorted(images_names)
+    images = da.stack(load_image(image_name) for image_name in images_names).reshape(len(images_names), 1760, 1760)
+    with napari.gui_qt():
+        napari.view_image(images) 
+```
+
+__And here's the final output!__
+![Alt text](https://media.giphy.com/media/LO3gPbCs5AxApjAehW/giphy.gif)
```
psobolewskiPhD:
```suggestion
#### And here's the final output!
![Alt text](https://media.giphy.com/media/LO3gPbCs5AxApjAehW/giphy.gif)
```
Getting a wierd layout here:
image
Maybe just using a heading is better?

psobolewskiPhD (changes requested):
Some small comments, but importantly this isn't linked in the ToC!
Also maybe this should somehow be mentioned as an example in https://napari.org/dev/tutorials/processing/dask.html ?

docs/tutorials/processing/brain_dataset_tutorial.md
```diff
@@ -0,0 +1,73 @@
+# Using Dask and napari to view and process a high-resolution human brain imaging dataset.
+We used napari to view and process a high-resolution human brain imaging dataset, and in the following tutorial we will break down how it all builds up!
+
+The dataset was obtained from [Dr. Brian Edlow's Lab](https://www.comarecoverylab.org/) for NeuroImaging of Coma and Consciousness, which is dedicated to promoting recovery of consciousness in people with severe brain injuries.
+This dataset is a 100 micron resolution magnetic resonance imaging (MRI) scan of an ex vivo human brain specimen. The brain specimen was donated by a [58-year-old woman](https://www.youtube.com/watch?v=Q-9jzBkoNuI) who had no history of neurological disease and died of non-neurological causes.
```
psobolewskiPhD: Worth linking to the actual repository too—I think this is the correct link:
https://doi.org/10.5061/dryad.119f80q
and maybe the preprint where this is all described:
https://doi.org/10.1101/649822

docs/tutorials/processing/brain_dataset_tutorial.md
jni: @psobolewskiPhD just adding a blank line before the image will fix things.

docs/tutorials/processing/brain_dataset_tutorial.md
melissawm: I am wondering then if this tutorial should actually focus on using dask_image? Because testing locally it does seem that we don't need to use dask.delayed at all. What do you think?

Another option would be doing the same operations twice, once using the dask.delayed interface, and another using dask_image (similar to what is done in the Dask tutorial).

psobolewskiPhD: Well, the original text was written in 2020...
While rolling your own and using `dask.delayed` is useful, that's already described in the dask tutorial.
This one is more about a cool sample dataset, so I think the simpler `dask_image.imread` approach is the more up-to-date way of doing that.
CC: @GenevieveBuckley

jni: Yeah, 100%, let's use the easiest tool for the job, for me this demo is about getting cool datasets into napari!

I'm ok with merging this PR but I *actually* think it might be better as a gallery example; and if we could figure out how to use output from napari-animation in gallery examples, that would be even more amazing!

melissawm: I'd love to experiment with that! For this example though, I think the dataset may be too large to have as a dependency, so not sure how that could be handled. (we could always have it as a non-executable example, though - just as I've done here)

melissawm:
Ok folks - here's a simplified version that can maybe later be converted into an example. Or - alternatively, this could be a section under the dask tutorial itself. Let me know what you prefer, but I can also do that in a follow up.

psobolewskiPhD:
@melissawm 
Gah, after reading the `contribute examples`, I'm pretty sure the `with qt_gui()` thing is not a thing anymore? See also https://napari.org/stable/guides/event_loop.html#napari-s-qapplication 
It does work, but we recommend the:
```
if __name__ == '__main__':
    napari.run()
```
So i think the `imread` should be above the __main__ along with the napari.view_image
Sorry I missed this before, i mostly use napari from an interactive console!

But yeah, it's *really* simple now, isn't it—which is a good thing. Not sure it's really a tutorial at this point 👀 
Reading the other dask tutorial, it doesn't have any sample data. It's `your_experiment.tif` and such. So, what about throwing this *above*:
https://napari.org/stable/tutorials/processing/dask.html#processing-data-with-dask-array-map-blocks
(so at the end of using `.imread`)
So we explain everything and then have a concrete example.

melissawm:
Yes that does sound better! And thanks for helping me clean this up!

melissawm:
Here it is: https://output.circle-artifacts.com/output/job/be30165d-9b99-4fa1-8124-a51f51d4628f/artifacts/0/docs/_build/tutorials/processing/dask.html#a-real-example-high-resolution-human-brain-imaging-dataset

psobolewskiPhD (approved):
I think this is great, really improves the other tutorial and fits really well.

GenevieveBuckley (dismissed):
Thanks for working on this @melissawm!

1. This should live in a separate file. It's not good to stuff new content into the middle of a different tutorial, and makes the flow much more confusing. I think this should be a requirement before merging.
2. I'm willing to be overruled on this, but I don't like that we're adding *more* content that never executes or checks for errors in the code blocks. The napari API changes rapidly, and we've definitely had problems before where tutorials become outdated and do not work because of this... but nobody notices because the code is never executed. That means new users try them, fail, and maybe feel like they must not be very good at this and give up on napari. Like I said, I'm willing to be overruled on this because it is hard with external datasets, but I don't really want to encourage more content with this downside.

jni:
I'm kind of on board with @GenevieveBuckley's comments. I wouldn't say it's *exactly* out of place with the existing tutorial, but it certainly didn't need it, and there is such a thing as information overload for new learners. Ideally this would be a gallery example. @melissawm over on Zulip @psobolewskiPhD said you might have some pooch + caching approaches that could make it work? Personally I'd love if we explored that. (And the dask tutorial itself could be run since Talley submitted his dataset to bioimage archive. But that's a separate issue.)

melissawm:
Sounds good- let me try and explore that and I'll get back to this. In the meantime I will keep this as draft. Thanks!

melissawm:
At the documentation working group today we discussed that this could be turned into an example in the gallery, and we could add a link in the main dask tutorial to the example. Thoughts?

GenevieveBuckley:
> At the documentation working group today we discussed that this could be turned into an example in the gallery, and we could add a link in the main dask tutorial to the example. Thoughts?

My thoughts are [the same as last time](https://github.com/napari/docs/pull/132#pullrequestreview-1407258164), not much has changed since then.

I thought all the gallery examples had to be executed? That seems potentially problematic given the size of the dataset [like you mention here](https://github.com/napari/docs/pull/132#discussion_r1175561305).

psobolewskiPhD:
I'm going to bump this.
Recently I was helping someone who was trying https://napari.org/stable/tutorials/processing/dask.html#make-your-life-easier-with-dask-image

As that tutorial doesn't have any example data, they were using their own, but getting:
```
ValueError: Array chunk size or shape is unknown. Possible solution with x.compute_chunk_sizes()
```
I was worried that something in napari was out-dated or whatnot.
I recalled **this** PR and we tried the data from this PR and it worked! So clearly the issue was with the tiffs, maybe because they were multi channel?
Anyhow, there is real value to an example with data you can pull down and use. A 2Gb zip file isn't a big deal IMO.

So I will again advocate for adding this to the existing dask tutorial as an example.
If in the future we can make this stuff work on CI, great! But if not, let's not perfect get in the way of good.

docs/tutorials/processing/dask.md
```diff
@@ -171,9 +171,51 @@ channels = [imread(file_pattern.format(i)) for i in range(nchannels)]
...
+for NeuroImaging of Coma and Consciousness, which is dedicated to promoting
+recovery of consciousness in people with severe brain injuries. This dataset is
+a 100 micron resolution magnetic resonance imaging (MRI) scan of an ex vivo
+human brain specimen. The brain specimen was donated by a [58-year-old woman](https://www.youtube.com/watch?v=Q-9jzBkoNuI) who had no history of neurological disease and died of non-neurological causes.
+This dataset was originally described in [Edlow, B.L., Mareyam, A., Horn, A. _et al._ 7 Tesla MRI of the ex vivo human brain at 100 micron resolution. Sci Data 6, 244 (2019)](https://doi.org/10.1038/s41597-019-0254-8).
+
+```{admonition} Getting the data
+
+You can download the dataset from [here](https://datadryad.org/stash/dataset/doi:10.5061/dryad.119f80q).
+A `.zip` file will be downloaded, and after extracting the compressed dataset
```
psobolewskiPhD:
```suggestion
In this example we use the [FLASH25_100um_TIFF_Axial data](https://datadryad.org/stash/downloads/file_stream/182482). This link will download a ~2 Gb `.zip` file  and after extracting the compressed data set,
```

GenevieveBuckley: Is it possible to add the expected size of the download to this sentence as well? Maybe in brackets after the link or something. I think that's helpful for users to know whether or not they have bandwidth for it or not.

psobolewskiPhD: Previously the unzipped size was given, but I've now added the zipped size as well (~2 Gb)

melissawm:
Rebased and updated with alt-text. Thanks for the ping @psobolewskiPhD , happy to iterate further if needed.

psobolewskiPhD:
Thanks @melissawm !
I just read it over and ran it locally and it's great!
Let's get this in!

PR #137, opened 2023-04-12
Explain how to add new examples to the gallery

### LINKED ISSUE #92: Add instructions on how to include new gallery examples
## 📚 New content request

Currently, there are a few subtleties to make sure a new example added to the gallery is correctly displayed:
- It must have a "Title", written in reST format at the top of the document;
- To show up in the ToC sidebar, it must also be assigned to at least one tag.

Example:
```
"""
Add image
=========

Display one image using the :func:`view_image` API.

.. tags:: visualization-basic
"""
```

We should add a page (or a section, either in the contributing guide or the contributing documentation pages) about adding new examples, including these requirements.

### Outline

- Where are the files which generate the examples (napari/napari)
- How to include a new example: add a title and a tag
- How to make sure the example is being executed correctly

melissawm:
# Description
Adds some explanation on how to add new examples to the gallery.

## Type of change

# References
Closes #92

## Final checklist:


psobolewskiPhD (approved):
Awesome!

docs/developers/contributing.md
```diff
@@ -112,6 +112,50 @@ Now you are all set to start developing with napari.
...
+```
+
+The equals signs under "Add image" mark this as the example title. The text
+below is a short description of the example, and the tags are assigned by the
+use of the `.. tags::` directive. In this case, the only tag associated with
+this example is `visualization-basic`. Multiple tags can be separated with a
+comma.
+
+Note that the examples are the `.py` source files, and any outputs and images
+will be autogenerated when the documentation site is built.
```
jni: Do we also want to add a note that `napari.run()` *has* to be under the `__main__` guard?

jni (approved):
@melissawm super great, thank you! 😊 

Do you think we should also point out that `napari.run()` needs to be under a `__main__` guard or things won't work? That won't be obvious to most users.

docs/developers/contributing.md
psobolewskiPhD: Oh, yes. That's a good point I think. For people who use mostly notebooks and interactive environments the `__main__` and `napari.run()` could be a bit of a mystery?

melissawm: Good point! I will explain that.

melissawm:
Let me know if this is enough explanation - here's the relevant section: https://output.circle-artifacts.com/output/job/4f7e2cae-c069-4177-8bd7-506fe3230502/artifacts/0/docs/_build/developers/contributing.html#adding-examples-to-the-gallery

psobolewskiPhD:
Just to be clear: when contributing an example, for the example to be in our gallery though, the __main__ **is** required, right?
It's definitely required to be able to do `python example.py`

Also, I think ipython behaves same as jupiter, so copy paste the code just works, without the __main__

melissawm:
@psobolewskiPhD you are of course right - let me know if this rephrasing reads better?

docs/developers/contributing.md
```diff
@@ -112,6 +112,63 @@ Now you are all set to start developing with napari.
...
+use of the `.. tags::` directive. In this case, the only tag associated with
+this example is `visualization-basic`. Multiple tags can be separated with a
+comma.
+
+Note that the examples are `.py` source files, and any outputs and images will
+be autogenerated when the documentation site is built.
+
+````{note}
+If you are running any of these examples on a Jupyter notebook (or in the
+IPython console), the block
```
psobolewskiPhD:
```suggestion
If you are running any of these examples in an interactive environment (e.g. a Jupyter notebook or
IPython console), the block
```

docs/developers/contributing.md
```diff
@@ -112,6 +112,63 @@ Now you are all set to start developing with napari.
...
+
+```
+if __name__ == '__main__':
+    napari.run()
+```
+
+is not necessary, but it is required when running the examples as scripts.
```
psobolewskiPhD:
```suggestion
is not necessary—you can copy-paste the code above it. However, it is required when running examples as scripts, using for example `python add_image.py`.
```

docs/developers/contributing.md
```diff
@@ -112,6 +112,63 @@ Now you are all set to start developing with napari.
...
+Because our example gallery is built from Python scripts, you will see this
+block in all examples.
```
psobolewskiPhD:
```suggestion
Because our example gallery is built from Python scripts, you need to ensure this
block is present in all contributed examples.
```

psobolewskiPhD:
@melissawm I made some small suggestions, based on my understanding.

PR #143, opened 2023-04-20
Update links to napari hub docs

kandarpksk:
According to https://github.com/chanzuckerberg/napari-hub/wiki/Customizing-your-plugin's-listing#github, support for .napari/ will be deprecated at some point in the future, so updating this documentation.

PS: Also updated some links and made minor improvements on the test-and-deploy page.

PPS: The scope of this PR is now limited to what I've mentioned in the original postscript!


DragaDoncila:
Hi @kandarpksk, the core dev team would actually like to deprecate metadata outside of `napari.yaml` and standard python packaging metadata files, as per the discussion [here](https://github.com/chanzuckerberg/napari-hub/discussions/627). In the meantime we think it would be better to remove the `.napari` mention rather than update it to something that will probably need to be updated again.

kandarpksk:
Thanks @DragaDoncila, just read some of the discussion! Would you suggest...
- dropping this PR,
- revent the line about .napari for now and keep the rest of the minor changes proposed,
- or something else?

docs/plugins/test_deploy.md
```diff
@@ -104,13 +104,10 @@ Once your package is listed on [PyPI] (and includes the `Framework :: napari`
 [classifier]), it will also be visible on the [napari
 hub](https://napari-hub.org/). To ensure you are providing the relevant metadata and
 description for your plugin, see the following documentation in the [napari hub
-GitHub](https://github.com/chanzuckerberg/napari-hub/tree/main/docs)’s docs
```
DragaDoncila, line 107: I would suggest removing the entire sentence starting from `To ensure...` and references to the hub guide until we align on metadata sources and the hub guide is updated to reflect this.

docs/plugins/test_deploy.md
```diff
@@ -119,14 +116,15 @@ development process, see the [napari hub GitHub’s
 Wiki](https://github.com/chanzuckerberg/napari-hub/wiki).
 
 If you want your plugin to be available on PyPI, but not visible on the napari
-hub, you can add a `.napari/config.yml` file to the root of your repository with
-a visibility key. For details, see the [customization
+hub, you can add a `.napari-hub/config.yml` file to the root of your repository
```
DragaDoncila: Here I'm torn as currently the only option for updating visibility on the hub is to add this field (or remove your framework classifier). However, we really want to encourage adoption of this field from the manifest instead. We need to add this to `napari`, and the napari hub should also change their logic to read this field instead. We can open PRs to plugins using `config.yml` as needed, to avoid a lengthy deprecation step. I think for now this change is fine though.

DragaDoncila (commented):
@kandarpksk I've left a couple of comments suggesting changes to the page text. Let me know what you think. We're very eager to get the hub guide and logic updated to point to standard sources of metadata wherever possible - once that's done we can update this page again to point to the relevant resources.

docs/plugins/test_deploy.md
kandarpksk: FWIW, I was just learning more about plugin creation when I created this PR, and wouldn't want to be the one removing references to the hub guide (or in other words, I don't usually contribute to documentation, but am encouraged to submit small fixes when I notice any opportunities).

An alternative could be to mention that metadata sources are subject to change and that discussions to align them are ongoing. (And maybe that's something the hub guide could also be updated to say!)

docs/plugins/test_deploy.md
kandarpksk: I'm ambivalent about this change now, since regardless of what is said here, it's likely to change based on the discussion you referenced. The only advantage I guess is that it makes things consistent with the hub guide.

docs/plugins/test_deploy.md
DragaDoncila: @kandarpksk afaik the hub docs and preview page are currently being updated to highlight the preferred sources of metadata, so I think it's ok to update these links here, especially since the current links are invalid. I'm going to resolve this conversation.

docs/plugins/test_deploy.md
DragaDoncila: @kandarpksk I've opened [napari/#5879](https://github.com/napari/cookiecutter-napari-plugin/issues/160) to track the implementation of the `visibility` field in the plugin installer. I suggest we leave this bit of the docs as is until that's done, when we can update this section to refer to the manifest `visibility` field. If you update the PR to exclude the changes to this paragraph, I can go ahead and approve.

kandarpksk: I've excluded this change in the latest commit, and will update the PR title appropriately!

DragaDoncila (dismissed):
@psobolewskiPhD thanks for the ping. This looks good to go in for me. Question for @kandarpksk - is the preview page being deprecated? I remember there being some discussion about that.

DragaDoncila:
@psobolewskiPhD I've opened #336 to make these updates and also remove mention of the preview page.

PR #145, opened 2023-04-30
Update annotate_points.md

constantinpape:
# Description

This PR fixes issues with the point annotation tutorial:
- Make it work with latest napari version (tested with napari: 0.4.17, magicgui: 0.7.2)
- Add `refresh_colors` to see changes when labels are updated

## Type of change

# References

https://forum.image.sc/t/napari-point-annotation/80568/2


docs/tutorials/annotation/annotate_points.md
```diff
@@ -72,10 +72,14 @@ def create_label_menu(points_layer, labels):
 
     def label_changed(event):
         """Update the Points layer when the label menu selection changes"""
-        selected_label = event.value
+        try:
```
constantinpape: When I run the code `selected_label` is just a string (the label value), so calling `event.value` gives an `AttributeError`. I am not sure if there are cases where this is an `Event` type and calling `.value` is necessary, so I added the try-except.

docs/tutorials/annotation/annotate_points.md
```diff
@@ -72,10 +72,14 @@ def create_label_menu(points_layer, labels):
...
+            selected_label = event.value
+        except AttributeError:
+            selected_label = event
         current_properties = points_layer.current_properties
         current_properties['label'] = np.asarray([selected_label])
         points_layer.current_properties = current_properties
+        points_layer.refresh_colors()
```
constantinpape, line 78: Without this the point colors do not get updated on label change.

docs/tutorials/annotation/annotate_points.md
```diff
@@ -97,14 +101,17 @@ def point_annotator(
...
     viewer = napari.view_image(stack)
     points_layer = viewer.add_points(
+        data=fake_data,
         properties={'label': labels},
         edge_color='label',
         edge_color_cycle=COLOR_CYCLE,
         symbol='o',
         face_color='transparent',
-        edge_width=8,
+        edge_width=0.5,
```
constantinpape: The function signature of `add_points` has changed and now needs a relative instead of an absolute size for `edge_width`.

docs/tutorials/annotation/annotate_points.md
```diff
@@ -97,14 +101,17 @@ def point_annotator(
     """
     stack = imread(im_path)
 
+    fake_data = [[0.0, 0.0, 0.0]] * len(labels)
```
constantinpape: Without this fake data `add_points` fails for me with this error:
```
  File "/home/pape/software/conda/miniconda3/envs/sam/lib/python3.10/site-packages/napari/components/viewer_model.py", line 4, in add_points
    import itertools
  File "/home/pape/software/conda/miniconda3/envs/sam/lib/python3.10/site-packages/napari/layers/points/points.py", line 369, in __init__
    self._feature_table = _FeatureTable.from_layer(
  File "/home/pape/software/conda/miniconda3/envs/sam/lib/python3.10/site-packages/napari/layers/utils/layer_utils.py", line 956, in from_layer
    features = _features_from_properties(
  File "/home/pape/software/conda/miniconda3/envs/sam/lib/python3.10/site-packages/napari/layers/utils/layer_utils.py", line 1024, in _features_from_properties
    return _validate_features(properties, num_data=num_data)
  File "/home/pape/software/conda/miniconda3/envs/sam/lib/python3.10/site-packages/napari/layers/utils/layer_utils.py", line 999, in _validate_features
    return pd.DataFrame(data=features, index=index)
  File "/home/pape/software/conda/miniconda3/envs/sam/lib/python3.10/site-packages/pandas/core/frame.py", line 709, in __init__
    mgr = dict_to_mgr(data, index, columns, dtype=dtype, copy=copy, typ=manager)
  File "/home/pape/software/conda/miniconda3/envs/sam/lib/python3.10/site-packages/pandas/core/internals/construction.py", line 481, in dict_to_mgr
    return arrays_to_mgr(arrays, columns, index, dtype=dtype, typ=typ, consolidate=copy)
  File "/home/pape/software/conda/miniconda3/envs/sam/lib/python3.10/site-packages/pandas/core/internals/construction.py", line 120, in arrays_to_mgr
    arrays, refs = _homogenize(arrays, index, dtype)
  File "/home/pape/software/conda/miniconda3/envs/sam/lib/python3.10/site-packages/pandas/core/internals/construction.py", line 608, in _homogenize
    com.require_length_match(val, index)
  File "/home/pape/software/conda/miniconda3/envs/sam/lib/python3.10/site-packages/pandas/core/common.py", line 576, in require_length_match
    raise ValueError(
ValueError: Length of values (2) does not match length of index (0)
```
(This is for a case where I have 2 label values.)

docs/tutorials/annotation/annotate_points.md
```diff
@@ -114,6 +121,8 @@ def point_annotator(
     label_widget = create_label_menu(points_layer, labels)
     viewer.window.add_dock_widget(label_widget)
 
+    point_layer.data = []
```
constantinpape: Clear the fake data.

constantinpape (commented):
I have added comments to explain why the changes where necessary. Some of these are workarounds that probably should not be committed, in particular adding the fake data to the points layer.

docs/tutorials/annotation/annotate_points.md
jni: Ah, the issue is that `properties` is deprecated, and now you want to use `features` (when you have a table) and `feature_defaults` (when you just want to give options, as below). Can you try that? ie remove the fake data, change `properties=` to `feature_defaults=`?

docs/tutorials/annotation/annotate_points.md
jni: Well, optionally, you can pass `edge_width_is_relative=False` to use the old behaviour. tbh I suspect we might revisit this again in the future... CC @brisvag @andy-sweet. For now I would add a comment:

```suggestion
        edge_width=0.5,  # fraction of point size
```

Possibly it should be a bit smaller then, like 0.1?

docs/tutorials/annotation/annotate_points.md
jni: Ah, I think that magicgui was updated to use `psygnal`s rather than VisPy events (@tlambert03?), so that this is now indeed just the value of the thing. So rather than the try/except, just rename the parameter name:

```python
def label_changed(new_label: str):
```

docs/tutorials/annotation/annotate_points.md
jni: I *think* `current_properties` is deprecated and you should use `feature_defaults` instead. @andy-sweet?

docs/tutorials/annotation/annotate_points.md
constantinpape: I tried, but `feature_defaults` is currently not available via `add_points`, see https://github.com/napari/napari/issues/5789.

constantinpape:
Thanks for the comments @jni. It seems like `feature_defaults` is not yet supported in the latest release: https://github.com/napari/napari/issues/5789. I will wait with the changes here until this is resolved / we figure out what's going wrong there.

docs/tutorials/annotation/annotate_points.md
constantinpape: Done.

docs/tutorials/annotation/annotate_points.md
Czaki: They are available on main branch and this repository contains code for the current develop branch. We split docs from napari/napari after 0.4.17 release.

psobolewskiPhD:
@constantinpape I mentioned this in the napari issue, but crossposting here for the benefit of others potentially as well.
Our stable docs are synced with napari releases, so currently 0.4.17, but your example changes will be released with the *next version of napari*. 
This isn't very clear and to make it worse, we've made changes to the contributing guide that are not live yet. So you need to look here (latest in the version selector at the top):
[napari.org/dev/developers/documentation/index.html](https://napari.org/dev/developers/documentation/index.html)
Because examples and other docs will go with the next version of napari, we now recommend using development napari install along with development docs install, so everything is in sync.

docs/tutorials/annotation/annotate_points.md
constantinpape: > They are available on main branch and this repository contains code for the current develop branch. We split docs from napari/napari after 0.4.17 release.

I see. I was under the impression that the docs here reference the latest stable release. I have changed my set-up and I am now checking with the most up-to-date `main` branch of napari.

One more question though: what do you mean with 
> contains code for the current develop branch

I don't see a `develop` branch in https://github.com/napari/napari.

docs/tutorials/annotation/annotate_points.md
```diff
@@ -99,12 +99,12 @@ def point_annotator(
 
     viewer = napari.view_image(stack)
     points_layer = viewer.add_points(
-        properties={'label': labels},
+        feature_defaults={'label': labels},
```
constantinpape: @jni I have changed to `feature_defaults` in the constructor now. 
However, when running this code it now gives the following error:
```
  File "/home/pape/Work/my_projects/napari/doc-example/annotation.py", line 38, in main
    points = v.add_points(
  File "/home/pape/Work/my_projects/napari/napari/napari/components/viewer_model.py", line 4, in add_points
    import itertools
  File "/home/pape/Work/my_projects/napari/napari/napari/layers/points/points.py", line 440, in __init__
    self._feature_table = _FeatureTable.from_layer(
  File "/home/pape/Work/my_projects/napari/napari/napari/layers/utils/layer_utils.py", line 882, in from_layer
    return cls(features, defaults=feature_defaults, num_data=num_data)
  File "/home/pape/Work/my_projects/napari/napari/napari/layers/utils/layer_utils.py", line 714, in __init__
    self._defaults = _validate_feature_defaults(defaults, self._values)
  File "/home/pape/Work/my_projects/napari/napari/napari/layers/utils/layer_utils.py", line 947, in _validate_feature_defaults
    raise ValueError(
ValueError: Feature defaults contain some extra columns not in feature values: {'label'}
```
So it seems that it's currently not possible to pass the `feature_defaults` without also passing `features`. But in this case we don't have `features` (or points) yet, since they will be provided through the user annotations.
This error happens with the napari from the latest commit on `main`.

Is this a general issue with the `features` functionality (in that case I am happy to create an issue in `napari`), or am I still understanding or doing something wrong?

cc @andy-sweet

constantinpape:
Thanks for the clarification @psobolewskiPhD, I have updated my set-up accordingly and I am now testing my examples with the current napari `main` branch. (From my understanding this is the correct approach now.)

docs/tutorials/annotation/annotate_points.md
Czaki: It looks like a bug in the current napari. You should open an issue or PR.

constantinpape: See https://github.com/napari/napari/issues/5790

constantinpape: This is fixed by passing empty features.

docs/tutorials/annotation/annotate_points.md
constantinpape: I have now resolved the other issues, but simply replacing `current_properties` by `feature_defaults` everywhere does no fully work. The viewer starts, but changing the label (via menu or shortcut) does not change it for the currently selected point. It however changes for the next point that is placed.

I made two videos to demonstrate: first is the version with `current_properties` where everything works as expected (I am using a shortcut to flip the label in the video, but I also checked that changing it via the menu works):

https://user-images.githubusercontent.com/4263537/235505773-8242a255-d7fe-48aa-a89d-79de0bfb605e.mp4

And this is the version where I have replaced all `current_properties` with `feature_defaults`. Now switching the label for the currently selected point doesn't work anymore, but the next point that is placed has the correct label.

https://user-images.githubusercontent.com/4263537/235507388-6c13052c-a8ac-4fca-a7ad-c0a98b304a4b.mp4

I am not sure if I am doing something wrong or if there is still a bug.

cc @andy-sweet

brisvag: `current_properties` and `feature_defaults` are not equivalent; this is unfortunately a conflation we have with points that [we're currently trying to resolve](https://github.com/napari/napari/pull/5722#discussion_r1165758964).

In this case, this means that `current_properties` affects both the currently selected point, and the default values for the following point. However (due to a different issue which is [solved with `feature_defaults`](https://github.com/napari/napari/pull/5646)) it does not correctly update everything that needs to be updated.

On the other hand, `feature_defaults` *does* work as intended, but it intentionally does not affect the `current` features.

In short, I *think* you should probably change *both* here ^^' but maybe @andy-sweet knows better.

andy-sweet: @brisvag's summary is correct. They are intentionally not equivalent in an effort to simplify what `feature_defaults` means and what its setter does, compared to `current_properties`. I understand that's annoying when you need the extra functionality of `current_properties` - maybe we can just add a separate method to update the currently selected feature values (e.g. `Points.update_selected_features(...)`) to help real usage and keep things clear? Note that some similar discussion came up in https://github.com/napari/napari/pull/4218 where we updated a similar, simpler example.

Also, technically, `current_properties` is not deprecated yet (i.e. there is no deprecation warning). There is an intent to deprecate `properties`, `current_properties`, and `property_choices` (see https://github.com/napari/napari/pull/5477), but that intent has existed for some time, so I don't want to make predictions about when it will actually happen...

My instinct is to use `current_properties` for now since the whole example refers to the properties family and either workaround any bugs (i.e. by using `refresh_colors`) or file a bug related to the `current_properties` setter, so that can be fixed. Then the properties -> features swap should occur as part of the whole deprecation by whoever takes that me (i.e. probably me?).

And thanks for the effort here! If you're interested in trying to work through the nasty details here, I'm happy to help to make features work, but I also don't want people to miss out on a useful update to the docs.

constantinpape: Thanks for the explanations @brisvag and @andy-sweet. I would then suggest to move ahead and keep using `current_properties` for now; and we can revisit this once you actually deprecate it.
What do you think @jni? If you agree I can finish this up since the changes make the example run again.

melissawm:
Hi folks, is there anything we can do to push this PR forward? Thanks!

constantinpape:
Hi @melissawm,

I think this got stuck because it was unclear whether to move to the new `feature_defaults` or stick with `current_features` in the implementation.

See  this comment from above:
> Thanks for the explanations @brisvag and @andy-sweet. I would then suggest to move ahead and keep using current_properties for now; and we can revisit this once you actually deprecate it. What do you think @jni? If you agree I can finish this up since the changes make the example run again.

The problem at the time seemed to be that `feature_defaults` was not fully implemented yet and did not support the use-case yet. On the other hand `current_properties` was supposed to be deprecated.

I am not sure how these general features have evolved till then. I would potentially have time to finish this up in a few weeks (I am working on my own project that involves this functionality, so could then also update it here), but I would need some guidance on the state of the `feature_defaults` functionality.

andy-sweet:
> I am not sure how these general features have evolved till then. I would potentially have time to finish this up in a few weeks (I am working on my own project that involves this functionality, so could then also update it here), but I would need some guidance on the state of the `feature_defaults` functionality.

I don't think there have many major updates since your last look here (outside of a few related bug fixes). To summarize, here's my understanding of the state.

- The `Points.features_default` setter is available in a public release: [v0.4.18](https://github.com/napari/napari/blob/20cf5b2c7bcdad26adffa24c778b660cf89483e1/napari/layers/points/points.py#L640)
- `Points.current_properties` has not yet been deprecated on `main`, though we are in a good place to move on that for v0.5 (https://github.com/napari/napari/pull/5477).
- #5790 is still a bug on main, but since it has easy workarounds, is not a blocker for this documentation.

The documentation updates here are small, but valuable, so if it's not too much effort to finish the updates to use `properties` and have things working again for v0.4.17/18/19, that's still my vote. Especially as any changes to fix/change the behavior of `features` is going to have to wait for v0.5. 

If we can make it work with `features` as they are now (i.e. in v0.4.19) without too much effort, that's also viable - I just don't want to block a doc bug-fix indefinitely.

psobolewskiPhD:
So if I'm understanding correctly, it seems like it's either
- get merged ASAP and cherry picked into 0.4.19, using properties
OR
- update to features and milestone for 0.5.0
cc: @Czaki

andy-sweet:
> So if I'm understanding correctly, it seems like it's either
> 
> * get merged ASAP and cherry picked into 0.4.19, using properties
>   OR
> * update to features and milestone for 0.5.0
>   cc: @Czaki

Yes, I like this as a simple summary. If we want to go for the first option, I think we could revert to one of the prior commits to make the fix only use `properties` and then verify that actually works with `v0.4.19`. I can take that on if we need it in the next couple of days.

Czaki:
If you validate that this works with 0.4.19 I will be happy to cherrypick.

andy-sweet:
I made a PR to the repo that contains runnable code for this example: https://github.com/kevinyamauchi/PointAnnotator/pull/5

This includes the fixes here, but only uses `properties` (not `features`) and also makes a couple of other updates. Once that's approved, we'll mirror the changes in the docs here. How does that sound @constantinpape  and @kevinyamauchi ?

@Czaki : I don't think this should hold up a v0.4.19 release because things have been broken for a while. I.e. only wait for this if you're waiting for something else.

constantinpape:
Thanks @andy-sweet. I think we can then just revert to [f3542e8](https://github.com/napari/docs/pull/145/commits/f3542e847be43bda74c7fe61d702203fb5960ab2) here.

andy-sweet:
> Thanks @andy-sweet. I think we can then just revert to [f3542e8](https://github.com/napari/docs/pull/145/commits/f3542e847be43bda74c7fe61d702203fb5960ab2) here.

Instead of reverting, I created a PR (https://github.com/constantinpape/docs/pull/1) on top of this to make this consistent with some other fixes I picked up along the way (https://github.com/kevinyamauchi/PointAnnotator/pull/5)

Feel free to merge that PR in or copy and paste any desired changes over here.

constantinpape:
Thanks @andy-sweet! I went ahead and merged your PR. I wasn't aware of `property_choices` before, that's really neat!
I will try it in my plugins!

constantinpape:
And from what I can see this is ready to be cherry-picked. Just let me know if there is anything else needed from my side.

andy-sweet (approved):
Approved, though I've made many contributions here now, so ideally would have one other approver. I have verified this works well enough with v0.4.19rc3, v0.4.15-18, and the main branch.

@Czaki : I marked this with v0.4.19 milestone, so if you can bring it in for that release, please do. But if it's too late or too hard, that's OK too!

PR #124, opened 2023-03-21
Add pr dependency check

brisvag:
# Description
Followup on [these comments](https://github.com/napari/docs/pull/123#issuecomment-1476477928), adding a dependency check for PRs.

This will automatically detect stuff like:
- depends on #123
- blocked by https://github.com/napari/napari/issues/5650

and it also works cross-repo (which is especially useful here to ensure we merge docs only if the feature actually landed in napari core).

## Type of change

- [ ] Fixes or improves existing content
- [ ] Adds new content page(s)
- [x] Fixes or improves workflow, documentation build or deployment

# References

## Final checklist:
- [ ] My PR is the minimum possible work for the desired functionality
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added [alt text](https://webaim.org/techniques/alttext/) to new images included in this PR

PR #147, opened 2023-05-04
Update installation docs to remove briefcase bundle mentions

jaimergp:
# Description

- [X] Removes outdated documentation

# References

Depends on https://github.com/napari/napari/pull/5804

## Final checklist:


Czaki:
I have merged #124. It may be used here/

jaimergp:
This is ready for review but we still need https://github.com/napari/napari/pull/5804 merged first.

Czaki (approved):

psobolewskiPhD:
@jaimergp 
Should we get https://github.com/napari/docs/pull/74 over the line as well in parallel to this one?
In this one you already have a warning about the bundle being somewhat experimental...

jaimergp:
@psobolewskiPhD we had a little discussion over that PR in the packaging meeting and the consensus seemed to be that we will wait a little bit longer on that one until we clarify how this is going to be maintained in terms of teams and responsibilities.

psobolewskiPhD:
Hmm, OK.
Priority needs to be removing the broken bundle.
But we should prioritize re-documenting the new bundle, even if we need to make it clearly as experimental. This way we can get more feedback to deal with issues/bugs.

Czaki:
@napari/copy This also needs to be checked.

docs/tutorials/fundamentals/quick_start.md
```diff
@@ -91,7 +85,7 @@ napari
 - If you installed the bundled app:
 
     Click on the app icon to open it.<br>
-    *Note: for mac users, it may require [security setting changes](https://napari.org/tutorials/fundamentals/installation.html).*
+    *Note: macOS users might need to right click on the app icon and select "Open" to bypass the security check. You can also go to System Setting> Security and Privacy and click on "Open Anyway".*
```
kne42:
```suggestion
    *Note: macOS users might need to right click on the app icon and select "Open" to bypass the security check. You can also go to System Settings > Privacy & Security and click on "Open Anyway".*
```

docs/tutorials/fundamentals/installation.md
```diff
@@ -46,12 +46,6 @@ You may also want:
 This is the easiest way to install napari if you only wish to use it as a standalone GUI app.
 This installation method does not have any prerequisites.
 
-```{important}
-Note that the bundled app is still
-in active development, and may not be very stable. We strongly recommend
-[installing as a Python package instead](#install-as-python-package-recommended).
-```
-
```
psobolewskiPhD, line 54: Why is this removed? My understanding is that this is still the case for the conda bundle too, at least to some degree?

docs/tutorials/fundamentals/installation.md
```diff
@@ -230,90 +224,14 @@ page](https://github.com/napari/napari/releases) and scroll to the release you
...
-The very first time you launch napari the startup time will be fairly slow, but
-after that you will find that napari launches more quickly.
-
-### Installing the Linux bundle
+```{important}
+Note that the bundled app is still
+in active development, and may not be very stable. We strongly recommend
+[installing as a Python package instead](#install-as-python-package-recommended).
+```
```
psobolewskiPhD: Ah! I see now. I strongly prefer the warning at the top before the user pulls the trigger, than at the bottom afterwards.

docs/tutorials/fundamentals/quick_start.md
```diff
@@ -48,13 +48,7 @@ You will also see some examples of plugins. The core napari viewer focuses on do
...
 ### Installation
 
-- Download the napari {{ napari_version }} bundled app for simple installation:
-
-    {{ '[Linux installation](https://github.com/napari/napari/releases/download/vNAPARI_VER/napari-NAPARI_VER-Linux-x86_64.zip)'.replace('NAPARI_VER', napari_version) }}<br>
-    {{ '[macOS installation](https://github.com/napari/napari/releases/download/vNAPARI_VER/napari-NAPARI_VER-macOS-x86_64.zip)'.replace('NAPARI_VER', napari_version) }}<br>
-    {{ '[Windows installation](https://github.com/napari/napari/releases/download/vNAPARI_VER/napari-NAPARI_VER-Windows-x86_64.zip)'.replace('NAPARI_VER', napari_version) }}<br>
-
-    *Note: for all other releases, please visit [here](https://github.com/napari/napari/releases) and look for Assets.*
+- Download the napari {{ napari_version }} bundled app for a simple installation. The step-by-step installers are available for macOS, Linux, and Windows in {{ '[the Releases page](https://github.com/napari/napari/releases/vNAPARI_VER)'.replace('NAPARI_VER', napari_version) }}, in the Assets section at the bottom.
```
psobolewskiPhD: I kinda like the direct links, can we modify them so they work with the new bundle? Perhaps with adding a new subsitution if needed?

psobolewskiPhD (changes requested):
I'd prefer the warning for the bundle being in dev being at the top.
Also I'd like the direct links to work, instead of just directing to the assets.
Can add a substitution if needed to make it simpler.
(Copy is fine IMO)

docs/tutorials/fundamentals/installation.md
jaimergp: I moved it because some parts of the documentation link to this section directly, instead of the (weird imo) forwarder above. But I don't have strong feelings; I just didn't want users to miss the warning.

docs/tutorials/fundamentals/quick_start.md
jaimergp: We have seen some releases with botched version names due to branch/tags shenanigans.  See https://github.com/napari/napari/releases/v0.4.16 for an example. That was the rationale here.

psobolewskiPhD: Hmm. I think I'd rather cross fingers we don't mess up tags and have direct links. If we mess up we can patch on the website repo?

docs/tutorials/fundamentals/installation.md
psobolewskiPhD: OK I see now. How about moving the warning up a bit though, to like new line 221, so after the bundle is introduced but before the directions on where to find the installers?

docs/tutorials/fundamentals/quick_start.md
jaimergp: Added links now :)

psobolewskiPhD (approved):
@jaimergp 
Looks good now, but I noticed a tiny typo in unrelated part of the installation.md:
new line 247:
```
    - include the output of `napari -info`
```
should be:
```
    - include the output of `napari --info``
```
(two dashes)
I can't suggest it, can you make that change? else it's another PR...

PR #177, opened 2023-06-12
Add documentation for #5806

melonora:
# Description
Documentation for the new polygon tool in the label layer

- depends on https://github.com/napari/napari/pull/5806

## Type of change

# References


psobolewskiPhD:
@melonora 
FYI, we have a `depends on` checker action, so you can add:
`Depends on https://github.com/napari/napari/pull/5806`
See: https://github.com/napari/docs/pull/124

And this way it will not be accidentally merged before the actual feature is in.
We probably need to consider milestones too, because a 0.5.0 feature shouldn't be documented in `stable` till 0.5.0

docs/howtos/layers/labels.md
```diff
@@ -263,6 +263,37 @@ pixels depending on if the contiguous property is disabled or not.
...
 layer is selected.
 
+## Drawing using polygons in the labels layer
+
+Another tool that can be used to quickly add or edit image segmentations is the 
+`polygon` tool. It combines functionality of the `paintbrush` and `fill bucket` 
+tool. The `polygon` tool can be activated by clicking on the icon resembling a 
+polygon in the layer control panel or by pressing `3`. Once activated, click 
+anywhere on the canvas to start drawing the polygon. After clicking on the canvas 
+the mouse. A vertex to be added will track your mouse cursor and a line with the
```
psobolewskiPhD:
```suggestion
polygon in the layer control panel or by pressing `3`. Once activated, left-click 
anywhere on the canvas to start drawing the polygon. A vertex to be added will track your mouse cursor and a line with the 
```

docs/howtos/layers/labels.md
```diff
@@ -263,6 +263,37 @@ pixels depending on if the contiguous property is disabled or not.
...
+added to the label image upon finishing the drawing process. The drawing process can be
+cancelled any time by pressing `escape`. This will delete the polygon being added and will 
+ensure the label layer is not changed.
+While the polygon overlay may be visible outside the canvas space during drawing, upon 
+finishing drawing the polygon will be cut off so that the part outside the canvas space is
+removed. This ensures that the dimensions of the label image are not larger than the image
+for which you are segmenting of for which you are editing the segmentations.
+
+If using the `polygon` tool for adding or editing segmentations of 3D image data, this only 
+adjusts the label layer in one plane, with the exception when viewing the image data as RGB.
```
psobolewskiPhD:
```suggestion
Note: if you use the `polygon` tool for adding or editing segmentations of 3D image data, you can only 
adjust labels in one plane, with the exception when viewing the image data as RGB. 
```

psobolewskiPhD (changes requested):
Thanks for this!
Made a few tiny suggestions—only the first is important, because something is missing or duplicated.

psobolewskiPhD (approved):

ksofiyuk:
@melonora Thank you for preparing the docs for the polygon tool. I've looked through it and I think that it would be better to change the document structure a little bit:

- Put the brush tool and the polygon tool in the same section called "Painting in the labels layer" as subsections as the polygon tool is just another way of painting in the labels layer. Both tools are different ways of painting, so I think they should be in one place.
- Replace paragraphs describing the polygon actions and their shortcuts with a corresponding bullet list. I find it not very convenient to navigate through paragraphs of texts if you just need to have a cursory view of the tool and its shortcuts.

> It combines functionality of the paintbrush and fill bucket tool. 

I think this sentence is a bit confusing as the polygon tool does not actually combine them. In that sense, the brush is a combination of the circle and the fill bucket tool, which does not make sense to me :smile:. I don't think that there should be any links between the fill bucket tool and the polygon tool.

melonora:
> @melonora Thank you for preparing the docs for the polygon tool. I've looked through it and I think that it would be better to change the document structure a little bit:
> 
>     * Put the brush tool and the polygon tool in the same section called "Painting in the labels layer" as subsections as the polygon tool is just another way of painting in the labels layer. Both tools are different ways of painting, so I think they should be in one place.
> 
>     * Replace paragraphs describing the polygon actions and their shortcuts with a corresponding bullet list. I find it not very convenient to navigate through paragraphs of texts if you just need to have a cursory view of the tool and its shortcuts.
> 
> 
> > It combines functionality of the paintbrush and fill bucket tool.
> 
> I think this sentence is a bit confusing as the polygon tool does not actually combine them. In that sense, the brush is a combination of the circle and the fill bucket tool, which does not make sense to me 😄. I don't think that there should be any links between the fill bucket tool and the polygon tool.

I changed it to bullet point format to provide a clearer overview. As for putting both in a section `Painting labels in layer`, I did not opt for that as the same could also be said for the fill bucket. Regarding the combined functionality, I now explicitly state that it does so by allowing for readily drawing enclosed instance segmentations

PR #178, opened 2023-06-13
Fix broken gallery examples with Qt error

### LINKED ISSUE #174: [Maint] The Gallery is broken in CircleCI previews
## 🧰 Task

Figure out what the issue is, the CircleCI action is different than the build_docs action, which works correctly—the gallery is fine in the .zip.
There is a difference in the order of the installs, which results in different ninja version.
There is a difference in python version
There is a difference in Qt backend.

I took some stabs at the first and second part, but that resulted in worse outcomes, with one of the examples failing totally and the build failing as as result, see:
https://github.com/napari/docs/pull/173

melissawm:
# Description
Closes #174 

Also updates the Python image on CircleCI and removes a spurious requirements.txt file that (I believe) is not in use anymore.

## Type of change

# References
Closes #174

## Final checklist:


psobolewskiPhD:
@melissawm I did some googling for the error and found some suggestions that this could a resource/RAM issue:
https://support.circleci.com/hc/en-us/articles/115014359648-Exit-Code-137-Out-of-Memory

melissawm:
Superseded by #207

PR #187, opened 2023-06-23
Try to use herbstluftwm in CircleCI to fix gallery

### LINKED ISSUE #174: [Maint] The Gallery is broken in CircleCI previews

psobolewskiPhD:
# Description
In CircleCI previews, the Gallery is broken. The errors appear to be related to pyside.
The normal build on Github works, but it uses a different xvfb setup and herbstluftwm.
Some googling suggests that maybe herbstluftwm is the key, so trying here.

## Type of change

# References
Closes https://github.com/napari/docs/issues/174

## Final checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,9 +52,12 @@ jobs:
           name: Build docs
           command: |
             . venv/bin/activate
-            herbstluftwm &
+            export DISPLAY=:99.0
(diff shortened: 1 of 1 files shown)


psobolewskiPhD:
no errors from the herbsluftwm but still doesn't work 😢

PR #194, opened 2023-07-03
Use python version range instead of min version

dstansby:
# Description
Where `python_minimum_version` was being used (one place), it was originaly saying that Python >= 3.8 was supported. This isn't correct, so I replaced this one instance with `python_range` and removed the `python_minimum_version` variable to avoid this happening in the future.

## Type of change

# References

## Final checklist:


psobolewskiPhD:
@Czaki do you think we should get this in 4.18? It basically just makes it that 3.11 isn't advertised as supported (>3.8 vs 3.8-3.10). 4.18 has experimental support for 3.11, so maybe it's moot?

psobolewskiPhD (approved):

PR #200, opened 2023-07-04
NAP 7: Key Binding Dispatch

kne42:
NAP for how key bindings will be handled

specific areas I would appreciate feedback on:
- do you agree with the justifications of what should/should not be considered a valid key binding?
- what are your thoughts on blocking all key bindings attached to the first part of a two-part key binding (e.g. if `Ctrl+A S` is active, `Ctrl+A` would not activate)
- which approach do you think is better, the mapping one or the prefix tree one? is there another approach you can suggest?


docs/naps/7-key-binding-dispatch.md
```diff
@@ -0,0 +1,427 @@
...
+
+In case (B), the corresponding command will never be triggered so long as it indirectly conflicts with another key binding. In this sense, multi-part key bindings will always take priority over single-part key bindings.
+
+### Finding a match
+
+When checking if an active key binding matches the entered key sequence, the resolver will fetch the pre-sorted list of direct conflicts and check if the last entry is active using its `when` property, moving to the next entry if it is not. When it encounters a blocking rule, it will return no match, and for a negate rule, it will store the affected command in an ignore list and continue to the next entry. If no special rules are present, it will return a match if the command is not in an ignore list, otherwise continuing to the next entry, and so on, until no more entries remain.
+
+In psuedo-code this reads as:
```
jni:
```suggestion
In pseudo-code this reads as:
```

docs/naps/7-key-binding-dispatch.md
```diff
@@ -0,0 +1,427 @@
...
+
+In case (B), the corresponding command will never be triggered so long as it indirectly conflicts with another key binding. In this sense, multi-part key bindings will always take priority over single-part key bindings.
+
+### Finding a match
+
+When checking if an active key binding matches the entered key sequence, the resolver will fetch the pre-sorted list of direct conflicts and check if the last entry is active using its `when` property, moving to the next entry if it is not. When it encounters a blocking rule, it will return no match, and for a negate rule, it will store the affected command in an ignore list and continue to the next entry. If no special rules are present, it will return a match if the command is not in an ignore list, otherwise continuing to the next entry, and so on, until no more entries remain.
```
jni, line 178: I don't know what blocking or negate rules are... Could you add a paragraph on these?

docs/naps/7-key-binding-dispatch.md
```diff
@@ -0,0 +1,427 @@
...
+    KeyChord(<KeyCombo.CtrlCmd|KeyX: 2089>, <KeyCode.KeyC: 20>),
+    KeyChord(<KeyCombo.CtrlCmd|KeyX: 2089>, <KeyCode.KeyV: 39>),
+]
+
+> list(filter(multi_part, keymap))
+[
+    KeyChord(<KeyCombo.CtrlCmd|KeyX: 2089>, <KeyCode.KeyC: 20>),
+    KeyChord(<KeyCombo.CtrlCmd|KeyX: 2089>, <KeyCode.KeyV: 39>),
+]
+```
```
jni: The formatting here is a bit weird... I suggest not mixing pure Python cells and cells containing prompts and outputs.

docs/naps/7-key-binding-dispatch.md
```diff
@@ -0,0 +1,427 @@
...
+
+Note that because of the spec, querying for modifiers will only check the first part:
```
jni: "because of the spec" — maybe be more precise about *which part of the spec*?

docs/naps/7-key-binding-dispatch.md
```diff
@@ -0,0 +1,427 @@
...
+    for _, entries in filter(conflict_filter, keymap.items()):
+        if find_active_match(entries):
+            return True
+    
+    return False
+```
+
+### Completing the dispatch
+
+Putting everything together, the following psuedo code represents the logic of key binding dispatch:
```
jni:
```suggestion
Putting everything together, the following pseudo-code represents the logic of key binding dispatch:
```

jni (approved):
Wow this is very nicely written @kne42! I've left a couple of minor suggestions, but as per the NAP process, I think this can be merged very soon.

melissawm:
The docs build is erroring out because MyST (and reST) tries to add a "transition" (a horizontal bar) before the footnotes, but you also have a section name for it (`## References and Footnotes`). To prevent this error, you can add

```
myst_footnote_transition = False
```

to `conf.py`. Cheers!

docs/naps/7-key-binding-dispatch.md
```diff
@@ -0,0 +1,427 @@
...
+- `f1-f12`, `a-z`, `0-9`
+- `` ` ``, `-`, `=`, `[`, `]`, `\`, `;`, `'`, `,`, `.`, `/`
+- `left`, `up`, `right`, `down`, `pageup`, `pagedown`, `end`, `home`
+- `tab`, `enter`, `escape`, `space`, `backspace`, `delete`
+- `pausebreak`, `capslock`, `insert`, `numlock`, `printscreen`
+- `numpad0-numpad9`, `numpad_decimal`, `numpad_multiply`, `numpad_divide`, `numpad_add`, `numpad_subtract`
+
+a **key combination** is a base key pressed with one or more modifier keys, e.g. `ctrl+c` or `ctrl+shift+z`
+
+a **key chord** consists of two parts, of which each can be either a base key or a key combination, e.g, `ctrl+x v`
```
andy-sweet, line 45: What do you mean by two parts? A key combo represents simultaneously pressing several keys, but a chord represents a sequence? Must each part be pressed separately? Do they need to within a certain time gap?

docs/naps/7-key-binding-dispatch.md
```diff
@@ -0,0 +1,427 @@
...
+    USER = 500
+```
+
+### Indirect conflicts
+
+When a key sequence matches a key binding and is also a sub-sequence of a key sequence used by another key binding, it is considered an indirect conflict.
+
+There are two ways indirect conflicts can exist:
+
+A. the provided key sequence is a single modifier that is a modifier in another key binding's key combination or is a modifier in the first key combination of a key binding's key chord
```
andy-sweet: Maybe I don't have the right kind of brain to think about key bindings, but an example for each case would help a lot!

andy-sweet:
Overall, the document is beautiful.

I haven't kept up with the key binding discussion, so some of the details are a little lost on me, but I got the gist. I had some (hopefully) simple clarifying questions. Will leave more detailed discussion and questions for those with more context and stronger opinions here.

> * which approach do you think is better, the mapping one or the prefix tree one? is there another approach you can suggest?

The mapping for similar reasons as given in the NAP itself - because it's simpler and the prefix tree doesn't perform much better in a fairly large/complex case.

docs/naps/7-key-binding-dispatch.md
```diff
@@ -0,0 +1,427 @@
...
+Some users want to use traditional modifier keys as a base key in key binding for convenience purposes [^id2]. However, this can lead to conflicts since many key bindings may include the modifier key in their key sequence and thus cause confusion and cost extra engineering effort.
+
+The proposed restrictions on what key sequences can be used in a key binding aim to allow for the simplest user need while cutting down on any unnecessary complexities:
+
+- key combinations containing any modifier keys as a base key are invalid
+- key chords cannot contain a base key that is a modifier key (aka a single modifier)
+
+Here are some examples:
+- `alt` is **valid** as a base key because it contains no other modifiers and no other parts
+- `alt-meta` is **invalid** because it is a key combination comprised of only modifiers
```
Carreau: Is the `-` as a joiner on purpose ? You use + in other places.

docs/naps/7-key-binding-dispatch.md
```diff
@@ -0,0 +1,427 @@
...
+
+Conditional evaluation allows plugin developers to tie their keybinding to a specific viewer state.
+
+### Definitions
+
+**modifier keys** refer to `ctrl`, `shift`, `alt`, and `meta`. `meta` is also known as `cmd`, `win`, or `super` on osx, windows, or linux, respectively
+
+a **base key** is one of the following supported keys:
+- `f1-f12`, `a-z`, `0-9`
+- `` ` ``, `-`, `=`, `[`, `]`, `\`, `;`, `'`, `,`, `.`, `/`
```
Carreau: That is a US centric view. Many of those are characters, not keys, that might be present on the same key. `6` and `-` are the same key on french keyboard for example.

I'm fine having a US centric implementation, but I just want to underline that the nomenclature is important at otherwise it can lead to confusion.

I would suggest pulling a-z0-9 into their own category, and saying something along "keys that support one or more of the following character are base keys".

As far as I understand for the other ones, they may be on the same physical button, (eg, pageupdown, and arrow keys), but needs the fn button button to be pressed, so at keyboard level they register as a different key.

I believe vocabulary and definitions are important otherwise we will talk pass each other, and defining say button/key/characters may help us express things we couldn't do otherwise.

In practice keys/character distinction might be important for the physical layout of  where the shortcuts are. Typically games can use `wasd` for moving character, those are _key_ bindings and not character binding, as on other non-english keyboard, you want the positions of the button to be the same, but the character ends up being `zqsd` on say french keyboard.

brisvag (approved):
Great work! Should merge asap as per NAP rules :)

psobolewskiPhD:
This is really clear, but I have just one comment:
napari isn't so clear between keybindings and mouse modifiers—they're sort of handled the same and keyboard centric? (e.g. can't rebind scroll wheel). 
But keyboard modifiers for mouse actions are a pretty standard UI interaction (alt click, control click, shift click/drag etc).
It's pretty intuitive to bind the left side modifier keys for different interactions, like preserve labels or fill, etc.
I think some carve-out for that sort of behavior (e.g. press & hold a modifier alone vs. press and release a key (with optional modifier)) would be worthwhile in the long run.

docs/naps/7-key-binding-dispatch.md
kne42: no, will fix

docs/naps/7-key-binding-dispatch.md
kne42: yes, each part needs to be pressed separately. they do not need to be within a certain time gap, although that was a consideration for indirect conflicts (e.g. `ctrl+x` is both its own kb and also part of the chord `ctrl+x m`)

kne42:
> It's pretty intuitive to bind the left side modifier keys for different interactions, like preserve labels or fill, etc. I think some carve-out for that sort of behavior (e.g. press & hold a modifier alone vs. press and release a key (with optional modifier)) would be worthwhile in the long run.

So have special casing for if a modifier is pressed and held vs pressed and immediately released?

I think this adds too much complexity for my liking. Advanced users could start a timer when the key is first "pressed" (this signal can be delayed due to dispatching) and then check on release if they should execute different logic based on the time. I don't really want to support/encourage the usage of modifier keys as keybindings outside of changing the GUI (e.g. press `alt` to change the cursor to a crosshair to make it clear that you're doing something special while adding a point)

psobolewskiPhD:
Sorry if I wasn't clear, I meant a (keyboard) modifier (held) for a mouse action—exactly like your  case of alt to modify a select/cursor!

kne42:
> Sorry if I wasn't clear, I meant a (keyboard) modifier (held) for a mouse action—exactly like your case of alt to modify a select/cursor!

I still don't quite follow. Could you give me an example of the two different use cases someone might have and how we want to handle each?

My current understanding is that you want to have an additional registration system (outside of key bindings) for mouse bindings that similarly captures conditionals, modifiers, etc.

Some examples with most selection tools include click to change selection to an element, shift+click to add multiple to selection, and control+click to toggle selection of an element.

docs/naps/7-key-binding-dispatch.md
kne42: Oh, that's interesting. I find that explanation confusing as well, so I put down "a **base key** is a key that when pressed without a modifier key, produces one of the following [key codes](https://w3c.github.io/uievents-code/#keyboard-101)".

Do you find that satisfactory?

kne42:
> The docs build is erroring out because MyST (and reST) tries to add a "transition" (a horizontal bar) before the footnotes, but you also have a section name for it (`## References and Footnotes`). To prevent this error, you can add
> 
> ```
> myst_footnote_transition = False
> ```
> 
> to `conf.py`. Cheers!

hi @melissawm im still getting this error, any ideas?

docs/naps/7-key-binding-dispatch.md
Carreau: I think that' better, and thank for your link, it is interesting.

kne42: I think this excerpt is particularly interesting

Screenshot 2023-07-15 at 2 37 24 PM

e.g. `KeyCode.KeyQ` produces `'a'` on a French keyboard layout which is not immediately intuitive

kne42:
ok, i found the issue, for some reason, it initially correctly installs `markdown-it==2.2.0`, but later in ["Install napari dev"](https://app.circleci.com/pipelines/github/napari/docs/277/workflows/93530fa6-6f66-4510-b00f-8aef0b39b909/jobs/267?invite=true#step-105-248), it overwrites it to `markdown-it==3.0.0`, which is incompatible with `mdit-py-plugins==0.3.5`

kne42:
hi @jni could you give a re-review if possible? starting a merge countdown if no one has additional feedback at this time

PR #201, opened 2023-07-04
Add extra note for Windows and Make to avoid running `rm -rf C:\Users`

seankmartin:
# Description
We've added instructions to build the docs on Windows, but if you skip straight to that section and follow the guides for Git Bash, you could end up a bit frustrated.
The default Windows path naming for user profiles with spaces is like `C:\Users\First Last\...`, so if you cloned napari to a subdirectory of that and ran `make docs`, there is a chance that (as happened to me), the command `rm -rf C:\Users` is run by the makefile.
This is just because the handling of spaces is really weird in Make.

I've added a duplicate of the note nearer the top of the page about Make and spaces to reiterate this for Windows users, and I've tried to stress the severity of it.
I'm open to other thoughts here, but I'd really feel bad if someone went straight to the Windows build docs and ended up running `rm -rf C:\Users`.

## Type of change

### DIFF
--- a/docs/developers/documentation/index.md
+++ b/docs/developers/documentation/index.md
@@ -407,6 +407,15 @@ help you edit your document and find the right spot!
 
 ## Building the documentation on Windows
 
+```{note}
+It is very important that you clone the `napari/docs` repository to a path that does not contain spaces.
+For example, `C:\Users\myusername\Documents\GitHub\napari-docs` is a valid path, but \
+`C:\Users\my username\Documents\GitHub\napari-docs` is not.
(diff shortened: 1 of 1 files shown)


melissawm (approved):
Oh no, that is very dangerous! Thanks for catching this!

brisvag:
> The default Windows path naming for user profiles with spaces is like `C:\Users\First Last\...`, so if you cloned napari to a subdirectory of that and ran `make docs`, there is a chance that (as happened to me), the command `rm -rf C:\Users` is run by the makefile.

This is both hilarous and terrifying :facepalm:

psobolewskiPhD (approved):

Issue #153, opened 2023-05-10
Reorganising and improving plugin documentation


seankmartin:
## 📚 New content request

Hello everyone!

The napari plugin ecosystem and the content at [napari.org/plugins](https://napari.org/dev/plugins) have been growing rapidly, and it would be beneficial to reorganize the plugin documentation to improve navigation. Here is a proposal for restructuring the existing plugin-related content and introducing new valuable information. This layout has already been shared with @DragaDoncila and the initial round of feedback has been incorporated into this proposal. Please add any suggestions or comments regarding this proposal - I'd love to hear your thoughts!

### Mock-up of the plugins main page

Here is a mock-up of the proposed site layout at a top level:

![plugin-site-mock](https://github.com/napari/docs/assets/21295664/4666f4ca-2dff-45b5-98d5-e2fcb730f1ee)

### Proposed plugin documentation structure

<details>
<summary style="font-size:14px">Click to view the full plugin site layout proposal</summary>
<p>

![napari-plugin-proposal](https://github.com/napari/docs/assets/21295664/3e0b37b2-a776-4eb2-9ef2-0b72cc67f35a)

</p></details>

---

<details>
<summary style="font-size:14px">Click to view the proposed table of contents</summary>
<p>

* Getting started
  * Finding and installing plugins
  * What plugins can provide
  * How to use plugins
  * Getting support for plugins
* Tutorials
  * Filtering plugins on napari hub
  * Using plugins for cell segmentation
  * Converting a napari script into a plugin
  * Building a full plugin
* Building a plugin
  * Your first plugin
  * Plugin anatomy
  * Plugin contributions
  * Key concepts for plugins (e.g. multithreading)
  * Debugging during plugin development
  * Best practices
* User experience guidelines
  * Overview of user experience
  * Consistency with napari
  * Menus and tooltips
  * Containers and tabs
* Testing and publishing
  * Testing guidelines
  * Describing and categorising your plugin
  * Setting up the napari hub plugin preview page
  * Enhance your plugin's reach and demonstrate quality
  * In depth-guide to plugin testing (converted workshop content)
* Advanced topics
  * Migration to npe2
  * First generation plugins
* Technical references
  * Plugin contributions reference
  * Plugin manifest reference
  * PyPI and napari hub metadata linking reference

</p></details>

### How this proposal was developed

[...]

PR #203, opened 2023-07-05
Plugin docs restructure of content

seankmartin:
# Description
I've reorganised the content of /plugins following #153. I'm interested to see what you think!
I was very heavy handed with the use of grids for the index pages, because they are the best I could think of. Open to other options as always.

I also fixed a small number links on other pages outside of /plugins because I broke a lot of links with this restructure and I wasn't exactly sure what I caused and what was already broken - thought it was safer to just fix them.

Here is what the homepage of plugins would look like:
image

## Type of change

## Final checklist:

## Addendum

A small number of my commits got associated with my bot account (I set the wrong email in my git config because I deleted my config before creating #201 🙈). I genuinely don't know how to fix that. If it is going to be an issue I could make a new branch and squash all the commits into one big commit from my seankmartin account.


docs/plugins/getting_started/finding_and_installing_plugins.md
```diff
@@ -1,5 +1,5 @@
 (find-and-install-plugins)=
-# Finding and installing a napari plugin
+# Finding and installing plugins
```
melissawm: I can't comment on lines below, but because you have already disambiguated unpublished plugins vs. from the hub, I'd remove the section below on "Finding plugins on the napari hub".

melissawm (commented):
This looks great! Very well done and so much easier to follow - thanks @seankmartin !

A couple of very minor comments that could be done in a follow-up because they are more content-related:
- In building_a_plugin/best_practices.md, there is a section called "Don’t import heavy dependencies at the top of your module" - explicitly, it says "This point will be less relevant when we move to the second generation manifest-based plugin declaration" and if I'm not mistaken that's npe2, so just want to make sure we don't forget to update this.
- In the same best_practices.md file, I think it would be better to replace
```
See our [Tips for testing napari plugins](testing_and_publishing/test.html#plugin-testing-tips).
```
with
```
See [Tips for testing napari plugins](testing_and_publishing/index.html).
```

So far that's what I can observe, but overall looks awesome. Cheers!

seankmartin:
> * In building_a_plugin/best_practices.md, there is a section called "Don’t import heavy dependencies at the top of your module" - explicitly, it says "This point will be less relevant when we move to the second generation manifest-based plugin declaration" and if I'm not mistaken that's npe2, so just want to make sure we don't forget to update this.

Good catch!

docs/plugins/getting_started/finding_and_installing_plugins.md
seankmartin: I've removed that section ("Finding plugins on the napari hub")

melissawm:
For the docs build failure, @seankmartin can you try this: https://github.com/napari/docs/pull/200#issuecomment-1636898740

seankmartin:
> For the docs build failure, @seankmartin can you try this: [#200 (comment)](https://github.com/napari/docs/pull/200#issuecomment-1636898740)

Seems that worked, thanks @melissawm!

melissawm (approved):
I see conflicts now - otherwise this is good to go on my side!

docs/plugins/start_using_plugins/finding_and_installing_plugins.md
```diff
@@ -10,12 +10,6 @@ metadata, to produce a more user friendly way to find napari plugins.
 Similarly, plugins annotated on PyPI with `Framework :: napari` are listed in
 the `Plugins > Install/Uninstall Plugins` menu within napari.
 
-## Finding plugins on the napari hub
-
-The [napari hub](https://napari-hub.org) hosts information about all plugins.
-You can browse, search, and filter to find plugins that fit your needs.
-You can also share links to specific search results and plugins.
-
```
psobolewskiPhD, line 18: Caught my eye and I can't suggest on the right line, but below in `Installing plugins` new line 22:
Could you mention:
- scrolling through the list of plugins or filtering plugins using the text entry box at the top
- just clicking `install` in the tile of a plugin
This to not just focus on the tricker manual install as is currently.

psobolewskiPhD (approved):
This is just awesome, I love the tiles on the root page and everywhere else! 😍 
Makes it much easier to navigate. Super job.
I left a minor comment, but it could be done in a new PR.

docs/plugins/start_using_plugins/finding_and_installing_plugins.md
seankmartin: Good catch! I'm going to write a simple description of how to do this for this PR, but I'll keep in mind that this section could probably use a new screenshot ahead of 0.5.0 since the plugin installer menu looks a little different.

seankmartin:
> I see conflicts now - otherwise this is good to go on my side!

Hopefully got those conflicts sorted out

PR #210, opened 2023-07-26
Add npe2 and magicgui to requirements

dstansby:
# Description
Add npe2 and magicgui to the requirements file.

If found that when running `make prep-docs` it needs `npe2` and `magicgui`, so add to requirements.

## Type of change

# References

## Final checklist:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,3 +13,4 @@ napari-sphinx-theme
 matplotlib
 lxml
 imageio-ffmpeg
+npe2
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,3 +14,4 @@ matplotlib
 lxml
 imageio-ffmpeg
 npe2
+magicgui


melissawm:
Sounds good to me, although because I always build with a development version of napari already built, these will be included (for example, if you use `pip install napari[dev,pyqt]` before building.)

Czaki:
As @melissawm said. It looks like the requirements contain only this, which is not in the napari requirements.

dstansby:
In a fresh environment with nothing installed shouldn't `make prep-docs` do everything needed to allow `make html` to run succesfully though?

Czaki:
> In a fresh environment with nothing installed shouldn't `make prep-docs` do everything needed to allow `make html` to run succesfully though?

No. It od not clone the napari repository that is required to build examples.

dstansby:
Hmm okay, so what is the point/definition of what should go in `requirements.txt` in this repository? And what is the scope of `make prep-docs`?

Czaki:
> Hmm okay, so what is the point/definition of what should go in `requirements.txt` in this repository?

As I understand requirements.txt contains old section `docs` from napari requirements.


> And what is the scope of `make prep-docs`?

clone the repositories that are required for build docs (npe2) and reset it it a proper version of npe2, to embed selected part in build docs.

dstansby:
Was the conclusion that this PR is not needed?

brisvag:
I'm still confused about this whole docs split, hopefully @melissawm can clarify :)

GenevieveBuckley:
I'm with @dstansby - it seems clear cut that if these requirements are necessary to build the docs, they should be included in the docs requirements list. It's irrelevant that other people have development environments where they have been installed in an indirect manner, since clearly they are required no matter how someone chooses to install them.

GenevieveBuckley (approved):

jni:
I would say that we should remove requirements.txt altogether rather than update it.

Czaki:
> I'm with @dstansby - it seems clear cut that if these requirements are necessary to build the docs,

whole napari repository and all its dependencies are required to build docs, as without it the API and examples section will fail.

Czaki:
> I would say that we should remove requirements.txt altogether rather than update it.

The whole point of having requirements here is to not need to make separate PR to the main repository to change dependency. but we could rename this file to something more specific.

jni:
> The whole point of having requirements here is to not need to make separate PR to the main repository to change dependency. but we could rename this file to something more specific.

Ah, got it. Ech. napari/napari#6049 😜

jni:
But in the meantime though I guess we should merge this? But it would be good for `napari[docs]` to contain all the requirements, and then that means keeping two files in sync...

psobolewskiPhD:
> But in the meantime though I guess we should merge this?

If we're merging this, don't we also need to add `napari`? Or really just add napari? But building docs with released napari will make stuff out of date, so you really do need a cloned napari, like the guide says--but not the guide on `stable`, the up-to-date guide on `latest`.

I think instead of merging this, which doesn't accomplish much, maybe we should put a brief step-by-step of building locally in the README of this repo. This would target someone who is basically a dev or contributor and pretty savvy, just needing to know what is needed. Of course, with a link to the Docs Contribution guide for further details and other approaches.

PR #216, opened 2023-08-01
Run PR depenecy check on every comment added to PR

Czaki:
# Description

For the current state, the core dev needs to restart the dependency check, or a dummy commit needs to be done. 

As the PR dependency check is cheap, I think it is worth triggering it on every comment, so if a user write comment containing `merge` the check will be restarted

## Type of change

# References

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_comment-use-issue_comment

### DIFF
--- a/.github/workflows/pr_dependencies.yml
+++ b/.github/workflows/pr_dependencies.yml
@@ -1,7 +1,9 @@
 # https://github.com/marketplace/actions/pr-dependency-check
 name: "PR Dependency Check"
 on:
-  - pull_request
+  pull_request:
+  issue_comment:
+    types: [ created ]
 
 jobs:
   check_dependencies:
--- a/.github/workflows/pr_dependencies.yml
+++ b/.github/workflows/pr_dependencies.yml
@@ -8,6 +8,7 @@ on:
 jobs:
   check_dependencies:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


brisvag (approved):
Nice :+1:

brisvag:
Though maybe *every* comment is still a lot, since it will spin up a VM every time? shall we have a trigger word instead?

Czaki:
We may, like in https://github.com/napari/napari/pull/6101. What phase did you suggest?

brisvag:
Yeah better. Something like "dependency closed"?

Czaki:
Maybe just "dependency"?

brisvag:
Though it will fire if we talk about library dependencies xD

Czaki:
hm. Maybe? I do not know. but it is now easy to suggest proper phase by "request changes"

psobolewskiPhD:
would `pr merged` be too easy to accidentally trigger? would accidental trigger have too high a cost?

brisvag:
deps can also be issues in principle I think... I think we're being overzealous here, let's just merge and if problems arise we can solve them :P

brisvag:
@Czaki please update description first!

Czaki:
@brisvag updated description

PR #217, opened 2023-08-01
Fix getting started in napari linking to the unittest getting started page

seankmartin:
Clicking the Getting started grid in the below image pointed to https://docs.python.org/3/library/unittest.mock-examples.html#getting-started.
image

I've changed the link, so it *should* now link to https://napari.org/dev/tutorials/fundamentals/getting_started.html, which I assume was the intention here.

### DIFF
--- a/docs/tutorials/start_index.md
+++ b/docs/tutorials/start_index.md
@@ -22,7 +22,7 @@ How to do a clean install of napari and launch the viewer.
 :::
 
 :::{grid-item-card} Getting started
-:link: getting-started
+:link: getting_started
 :link-type: ref
 
 This tutorial will teach you all the different ways to launch napari.


melissawm (approved):
Good catch, thank you!

psobolewskiPhD (approved):
wow, how'd that slip through!
Thanks

psobolewskiPhD:
@melissawm should we consider making this fix as a patch in the live docs?

melissawm:
Happy to do that - once this is merged I can backport the changes to stable at least.

melissawm:
(I hope it's ok that I created a new label to keep track of what we want to backport to stable)

psobolewskiPhD:
awesome idea!

PR #226, opened 2023-08-11
Add nap for telemetry

Czaki:
# Description

Add NAP with the proposition of a telemetry system.

## Type of change

### DIFF
--- a/docs/naps/8-telemetry.md
+++ b/docs/naps/8-telemetry.md
@@ -50,7 +50,7 @@ The user should be able to adjust the telemetry level of detail. The following l
 
 1. `none` - no telemetry is collected
 2. `basic` - information about the napari version, python version, OS, and CPU architecture is collected and if it is the first report by the user. There is also a user identifier created based on computer details that will be rerendered each week to prevent tracking the user, but allow to not count a user multiple times. 
-3. `middle` - same as in `basic` but also information about the list of installed plugins and their versions is collected. We take care to not collect data about plugins that are not intended to be public. 
+3. `middle` - same as in `basic` but also information about the list of installed plugins and their versions is collected. We take care to not collect data about plugins that are not intended to be public, so we will not collect information about plugins searchable as napri plugin using plugin dialog or napri-hub. We also will not collect information about plugins that are installed in non stable version.
 4. `full` - same as in `middle` but also collects information about plugin usage by binding to app-model and logging plugin commands used. Also basic information about data like type (`np.ndarray`, `dask.array`, `zarr.Array`, etc.) and its size is collected.
 
 There should be a visible indicator that telemetry is enabled (for example on the status bar). 
--- a/docs/naps/8-telemetry.md
+++ b/docs/naps/8-telemetry.md
@@ -16,17 +16,17 @@
   
  ## Abstract 
   
- This NAP is describing why Telemetry is helpful to the Napari project and describes the architecture and solutions selected to maximize the privacy of our users. 
+ This NAP is describes why telemetry is helpful to the napari project and the architecture and solutions selected to maximize the privacy of our users. 
  
  ## Motivation and Scope
 
 With the growth of napari, the standard feedback loop through napari community meetings and napari-related events at conferences has reached its capacity. Also, we collect many feature requests for which we cannot find volunteers for implementation. 
 
-To have the possibility of sustainable development of the project we need either have funds to pay contractors or have some company that donates their worker times manageable by core-devs.
+To have the possibility of sustainable development of the project we will either need to rely on paid contractors or on companies donating employee time managed by the core devs.
 
-Both these scenarios require our side the ability to provide some information about the estimated number of users to prove to potential founders that their donation/grant will be used in a valuable way. 
+Both scenarios require us to provide some information about the estimated number of users to prove to potential funders that their donation/grant will be used in a valuable way. 
 
-Adding the option for monitoring plugin usage allows us to identify the most important plugins and try to establish cooperation with their maintainers to reduce the probability that the plugin will not be ready for a new release. Such monitoring could contain not only the list of installed plugins but also which commands contributions are used most often.
+Adding the option for monitoring plugin usage allows us to identify the most important plugins and try to establish cooperation with their maintainers to reduce the probability that the plugin will not be ready for a new napari release. Such monitoring could contain not only the list of installed plugins but also which commands and contributions are used most often.
 
 Also collecting information about data types and their size will provide valuable information about the typical use cases of napari.
 
@@ -35,11 +35,11 @@ Still, a user need to be able to opt out of such monitoring. And adjust the leve
   
  ## Detailed Description
 
-`napari-telemetry` will be a package responsible for collecting and sending telemetry data to the napari server. It will be installed after user confirmation. It will contain callbacks for collection data, and utils for its storage and sending. Also, this package will contain utils for validation if the user has agreed to telemetry. 
+`napari-telemetry` will be a package responsible for collecting and sending telemetry data to the napari server. It will be installed after user confirmation. It will contain callbacks for data collection, and utils for storage and sending. Also, this package will contain utils for validating if the user has agreed to telemetry. 
 
-In the main package, there is a need to add code to ask users if they want to enable telemetry. This 1code should be executed only once per environment.
+In the main package, there is a need to add code to ask users if they want to enable telemetry. This code should be executed only once per environment.
 
-Telemetry should contain following way to disable it:
+Telemetry should contain following ways to disable it:
 
 1. uninstall `napari-telemetry` package
 2. Environment variable `NAPARI_TELEMETRY=0`
@@ -49,12 +49,12 @@ The user should be able to adjust the telemetry level of detail. The following l
 
 1. `none` - no telemetry is collected
 2. `basic` - information about the napari version, python version, OS, and CPU architecture is collected and if it is the first report by the user. There is also a user identifier created based on computer details that will be rerendered each week to prevent tracking the user, but allow to not count a user multiple times. 
-3. `middle` - same as in `basic` but also information about the list of installed plugins and their versions is collected. We take care to not collect data about plugins that are not indented to be public. 
-4. `full` - same as in `middle` but also information about plugin usage by binding to app-model and collect information about called plugins command. Also basic information about data like type (`np.ndarray`, `dask.array`, `zarr.Array`, etc.) and its size is collected.
+3. `middle` - same as in `basic` but also information about the list of installed plugins and their versions is collected. We take care to not collect data about plugins that are not intended to be public. 
+4. `full` - same as in `middle` but also collects information about plugin usage by binding to app-model and logging plugin commands used. Also basic information about data like type (`np.ndarray`, `dask.array`, `zarr.Array`, etc.) and its size is collected.
 
 There should be a visible indicator that telemetry is enabled (for example on the status bar). 
 
-The second part of this work should be setup the server to collect telemetry data. Next to collecting data, it should provide a basic public dashboard that will allow a community to see aggregated information.
+The second part of this work should be to setup the server to collect telemetry data. After collecting data, it should provide a basic public dashboard that will allow the community to see aggregated information.
 
 I propose to have the following data retention policy:
 
@@ -65,9 +65,9 @@ I propose to have the following data retention policy:
 ## Privacy assessment
 
 During the preparation of this NAP we assume that none of the collected data will be presented in 
-a form that allows to identify a single user or identify a research area of user. We also select a set of data that will be collected to minimize the possibility of reval of fragile data, but it is impossible to guarantee that it will not be possible to identify a single user (for example checking plugins combination).
+a form that allows to identify a single user or identify a research area of user. We also select a set of data that will be collected to minimize the possibility of revealing fragile data, but it is impossible to guarantee that it will not be possible to identify a single user (for example by checking installed plugin combinations).
 
-Because of this, we decided to not publish raw data and only show aggregated results. The aggregation will be performed using scripts. Napari core-devs will access raw data only when there will be some errors in the aggregation process.
+Because of this, we propose to not publish raw data and only show aggregated results. The aggregation will be performed using scripts. Napari core devs will access raw data only if there are errors in the aggregation process.
 
 We also will publish a list of endpoints for each level of telemetry, so the given level of telemetry could be blocked on the organization level (for example by the rule on the firewall).
 
@@ -92,27 +92,27 @@ https://github.com/grafana/grafana
 
 The main thing for implementation should be the low cost of maintenance. So the solution should be as simple as possible. We could either use existing solutions on the server side or implement our own.
 
-The benefit of the existing solution is that most of the work is already done. The downside is that it may require additional cost of maintenance. This cost may be caused by many features that are not needed for napari and could increase the risk of leaking data.  As I check they are implemented in techniques that are not familiar to napari core-devs. So if there will be a decision to use them we should select an SAS solution that will be maintained by the company.
+The benefit of existing solutions is that most of the work is already done. The downside is that it may require additional cost of maintenance. This cost may be caused by many features that are not needed for napari and could increase the risk of leaking data.  Quick checks of their code revealed they are implemented in techniques that are not familiar to napari core devs. So, if we decide to use them, we should select an SAS solution that will be maintained by the company.
 
 
-For the current, I suggest creating a simple REST API server for collecting the data. 
+For now, I suggest creating a simple REST API server for collecting the data. 
 It could be a simple Python FastAPI server that will store data in the SQLite database.
 
 Data for aggregation should be extracted from the database using a script running on the same machine.
 
 The output of the aggregation script should be loaded to some existing visualization tool, like grafana.
 
-It may be nice to host them on separate servers, then even if the data presented on the dashboard will be compromised, the raw data will be not exposed to the world.
+It may be nice to host raw and aggregate data on separate servers - then even if the data presented on the dashboard is compromised, the raw data will be not exposed to the world.
 
-Having both server and aggregation scripts in Python will reduce maintenance costs for napari core-devs.
+Having both server and aggregation scripts in Python will reduce maintenance costs for napari core devs.
 
-We should register the `telemetry.napari.org` domain and use it for the server. The main page contains this NAP and a link to the summary dashboard.
+We should register the `telemetry.napari.org` domain and use it for the server. The main page will contain this NAP and a link to the summary dashboard.
 
 
 The main part of the application side should be implemented in `napari-telemetry` package. 
 The package should not report in stream mode, but collect data on the disk and send it in batches. This will reduce the risk of leaking data. The package should implement a utility to allow users to preview collected data before sending it to the server.
 
-In the napari itself following changes should be implemented:
+In napari itself, the following changes should be implemented:
 
 1) The indicator that shows the telemetry status
 2) The dialog that asks a user if they want to enable telemetry
@@ -152,7 +152,7 @@ A nice extension may be the ability for the steering council to create a certifi
  But based on talks that happen during the discussion we may think about more frequent checks for updates to inform users that they could update their Napari or plugin version. For such a change we need to update our update server to provide information per Python version (as some plugins could drop old Python earlier).
 
 
-The second alternative is use a third-party solution like [plausable.io](https://plausible.io/). But from my perspective, it is harder to adjust a set of data that is collected as these services are designed to monitor webpages. 
+The second alternative is use a third-party solution like [plausible.io](https://plausible.io/). But from my perspective, it is harder to adjust a set of data that is collected as these services are designed to monitor webpages. 
 
   
  ## Discussion 
--- a/docs/naps/8-telemetry.md
+++ b/docs/naps/8-telemetry.md
@@ -30,7 +30,7 @@ Adding the option for monitoring plugin usage allows us to identify the most imp
 
 Also collecting information about data types and their size will provide valuable information about the typical use cases of napari.
 
-Still, a user need to be able to opt out of such monitoring. And adjust the level of detail of the information that is sent to the napari server.
+Still, users need to be able to opt out of such monitoring, and adjust the level of detail of the information that is sent to the napari server.
  
   
  ## Detailed Description
--- a/docs/naps/8-telemetry.md
+++ b/docs/naps/8-telemetry.md
@@ -1,0 +1,181 @@
+:orphan: 
+  
+ (nap-8)= 
+  
+ # NAP-8 — Telemetry
+  
+ ```{eval-rst} 
+ :Author: Grzegorz Bokota 
+ :Created: <date created on, in yyyy-mm-dd format> 
+ :Resolution: <url> (required for Accepted | Rejected | Withdrawn) 
+ :Resolved: <date resolved, in yyyy-mm-dd format> 
+ :Status: Draft
+ :Type: Standards Track 
+ :Version effective: <version-number> (for accepted NAPs) 
+ ``` 
+  
+ ## Abstract 
+  
+ This NAP is describing why Telemetry is helpful to the Napari project and describes the architecture and solutions selected to maximize the privacy of our users. 
+ 
+ ## Motivation and Scope
+
+With the growth of napari, the standard feedback loop through napari community meetings and napari-related events at conferences has reached its capacity. Also, we collect many feature requests for which we cannot find volunteers for implementation. 
+
+To have the possibility of sustainable development of the project we need either have funds to pay contractors or have some company that donates their worker times manageable by core-devs.
+
+Both these scenarios require our side the ability to provide some information about the estimated number of users to prove to potential founders that their donation/grant will be used in a valuable way. 
+
+Adding the option for monitoring plugin usage allows us to identify the most important plugins and try to establish cooperation with their maintainers to reduce the probability that the plugin will not be ready for a new release. Such monitoring could contain not only the list of installed plugins but also which commands contributions are used most often.
+
+Also collecting information about data types and their size will provide valuable information about the typical use cases of napari.
+
+Still, a user need to be able to opt out of such monitoring. And adjust the level of detail of the information that is sent to the napari server.
+ 
+  
+ ## Detailed Description
+
+`napari-telemetry` will be a package responsible for collecting and sending telemetry data to the napari server. It will be installed after user confirmation. It will contain callbacks for collection data, and utils for its storage and sending. Also, this package will contain utils for validation if the user has agreed to telemetry. 
+
+In the main package, there is a need to add code to ask users if they want to enable telemetry. This 1code should be executed only once per environment.
+
+Telemetry should contain following way to disable it:
+
+1. uninstall `napari-telemetry` package
+2. Environment variable `NAPARI_TELEMETRY=0`
+3. Full list of endpoints used for collecting telemetry, that could be filtered on the firewall level.
+
+The user should be able to adjust the telemetry level of detail. The following levels are proposed:
+
+1. `none` - no telemetry is collected
+2. `basic` - information about the napari version, python version, OS, and CPU architecture is collected and if it is the first report by the user. There is also a user identifier created based on computer details that will be rerendered each week to prevent tracking the user, but allow to not count a user multiple times. 
+3. `middle` - same as in `basic` but also information about the list of installed plugins and their versions is collected. We take care to not collect data about plugins that are not indented to be public. 
+4. `full` - same as in `middle` but also information about plugin usage by binding to app-model and collect information about called plugins command. Also basic information about data like type (`np.ndarray`, `dask.array`, `zarr.Array`, etc.) and its size is collected.
+
+There should be a visible indicator that telemetry is enabled (for example on the status bar). 
+
+The second part of this work should be setup the server to collect telemetry data. Next to collecting data, it should provide a basic public dashboard that will allow a community to see aggregated information.
+
+I propose to have the following data retention policy:
+
+1) Up to 2 weeks for logs.
+2) up 2 months of raw data (1 month of collection, then aggregation and time to validate aggregated data),
+3) infinite of aggregated data.
+  
+## Privacy assessment
+
+During the preparation of this NAP we assume that none of the collected data will be presented in 
+a form that allows to identify a single user or identify a research area of user. We also select a set of data that will be collected to minimize the possibility of reval of fragile data, but it is impossible to guarantee that it will not be possible to identify a single user (for example checking plugins combination).
+
+Because of this, we decided to not publish raw data and only show aggregated results. The aggregation will be performed using scripts. Napari core-devs will access raw data only when there will be some errors in the aggregation process.
+
+We also will publish a list of endpoints for each level of telemetry, so the given level of telemetry could be blocked on the organization level (for example by the rule on the firewall).
+
+
+If someone found that we are publishing some problematic data we will remove them and update the aggregation process to prevent such a situation in the future.
+This NAP will be updated to reflect the current state of telemetry. 
+
+ 
+## Related Work 
+
+Total systems:
+https://plausible.io/
+https://sentry.io/
+https://opentelemetry.io/
+
+Visualizations:
+https://github.com/grafana/grafana
+  
+
+  
+## Implementation 
+
+The main thing for implementation should be the low cost of maintenance. So the solution should be as simple as possible. We could either use existing solutions on the server side or implement our own.
+
+The benefit of the existing solution is that most of the work is already done. The downside is that it may require additional cost of maintenance. This cost may be caused by many features that are not needed for napari and could increase the risk of leaking data.  As I check they are implemented in techniques that are not familiar to napari core-devs. So if there will be a decision to use them we should select an SAS solution that will be maintained by the company.
+
+
+For the current, I suggest creating a simple REST API server for collecting the data. 
+It could be a simple Python FastAPI server that will store data in the SQLite database.
+
+Data for aggregation should be extracted from the database using a script running on the same machine.
+
+The output of the aggregation script should be loaded to some existing visualization tool, like grafana.
+
+It may be nice to host them on separate servers, then even if the data presented on the dashboard will be compromised, the raw data will be not exposed to the world.
+
+Having both server and aggregation scripts in Python will reduce maintenance costs for napari core-devs.
+
+We should register the `telemetry.napari.org` domain and use it for the server. The main page contains this NAP and a link to the summary dashboard.
+
+
+The main part of the application side should be implemented in `napari-telemetry` package. 
+The package should not report in stream mode, but collect data on the disk and send it in batches. This will reduce the risk of leaking data. The package should implement a utility to allow users to preview collected data before sending it to the server.
+
+In the napari itself following changes should be implemented:
+
+1) The indicator that shows the telemetry status
+2) The dialog that asks a user if they want to enable telemetry
+3) code to check if telemetry is enabled (to not load the `napari-telemetry` package if it is disabled)
+4) code required to init `napari-telemetry` package
+
+
+## Potential problems 
+
+There is a risk that someone may try to highjack the telemetry module name to have code executed at every napari start. 
+
+I do not expect that it is a high risk, but exists. We could address it by code signing. This will require additional procedures to protect private cryptographic keys.
+
+Another option is to scan public plugins and their dependencies. This is simpler but will require establishing additional communication channels to be able to warn users about the potential problem. 
+
+
+  
+## Backward Compatibility 
+  
+ Not relevant
+  
+## Future Work 
+
+A nice extension may be the ability for the steering council to create a certificate of telemetry output that could be given to plugin maintainers to prove to supervisors that their plugin is used by the community. 
+
+  
+ ## Alternatives 
+
+ During the discussion, there is a proposal to use the same approach as used in ImageJ. 
+ 
+ Mean that instead of implementing telemetry on the client side we could implement it on the update server side. The advantage and disadvantage of such a solution is that no user could opt out of telemetry. Also, such a method could potentially provide information about the Python version, napari version and list of installed plugins. All others will require a mechanism from this NAP.
+
+ It will also require updates on the Napari side as currently we only communicate with the update server when a user opens the plugin manager. Also, to have proper information about installed plugins we will need to send information about the list of installed plugins instead of just downloading the information about all plugins from the server. 
+
+ As this solution provides less information, does not allow for opt-out and could cause blacklisting of the update server IP address, I do not recommend it.
+
+ But based on talks that happen during the discussion we may think about more frequent checks for updates to inform users that they could update their Napari or plugin version. For such a change we need to update our update server to provide information per Python version (as some plugins could drop old Python earlier).
+
+
+The second alternative is use a third-party solution like [plausable.io](https://plausible.io/). But from my perspective, it is harder to adjust a set of data that is collected as these services are designed to monitor webpages. 
+
+  
+ ## Discussion 
+  
+ This section may just be a bullet list including links to any discussions 
+ regarding the NAP, but could also contain additional comments about that 
+ discussion: 
+  
+ - This includes links to discussion forum threads or relevant GitHub discussions. 
+  
+ ## References and Footnotes 
+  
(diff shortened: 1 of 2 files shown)


docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+ :Resolution: <url> (required for Accepted | Rejected | Withdrawn) 
+ :Resolved: <date resolved, in yyyy-mm-dd format> 
+ :Status: Draft
+ :Type: Standards Track 
+ :Version effective: <version-number> (for accepted NAPs) 
+ ``` 
+  
+ ## Abstract 
+  
+ This NAP is describing why Telemetry is helpful to the Napari project and describes the architecture and solutions selected to maximize the privacy of our users.
```
DragaDoncila:
```suggestion
 This NAP is describes why telemetry is helpful to the napari project and the architecture and solutions selected to maximize the privacy of our users. 
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+  
+ ## Abstract 
+  
+ This NAP is describing why Telemetry is helpful to the Napari project and describes the architecture and solutions selected to maximize the privacy of our users. 
+ 
+ ## Motivation and Scope
+
+With the growth of napari, the standard feedback loop through napari community meetings and napari-related events at conferences has reached its capacity. Also, we collect many feature requests for which we cannot find volunteers for implementation. 
+
+To have the possibility of sustainable development of the project we need either have funds to pay contractors or have some company that donates their worker times manageable by core-devs.
```
DragaDoncila:
```suggestion
To have the possibility of sustainable development of the project we will either need to rely on paid contractors or on companies donating employee time managed by the core devs.
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+
+Both these scenarios require our side the ability to provide some information about the estimated number of users to prove to potential founders that their donation/grant will be used in a valuable way.
```
DragaDoncila:
```suggestion
Both scenarios require us to provide some information about the estimated number of users to prove to potential funders that their donation/grant will be used in a valuable way. 
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+ 
+ ## Motivation and Scope
+
+With the growth of napari, the standard feedback loop through napari community meetings and napari-related events at conferences has reached its capacity. Also, we collect many feature requests for which we cannot find volunteers for implementation. 
+
+To have the possibility of sustainable development of the project we need either have funds to pay contractors or have some company that donates their worker times manageable by core-devs.
+
+Both these scenarios require our side the ability to provide some information about the estimated number of users to prove to potential founders that their donation/grant will be used in a valuable way. 
+
+Adding the option for monitoring plugin usage allows us to identify the most important plugins and try to establish cooperation with their maintainers to reduce the probability that the plugin will not be ready for a new release. Such monitoring could contain not only the list of installed plugins but also which commands contributions are used most often.
```
DragaDoncila:
```suggestion
Adding the option for monitoring plugin usage allows us to identify the most important plugins and try to establish cooperation with their maintainers to reduce the probability that the plugin will not be ready for a new napari release. Such monitoring could contain not only the list of installed plugins but also which commands and contributions are used most often.
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+
+Also collecting information about data types and their size will provide valuable information about the typical use cases of napari.
+
+Still, a user need to be able to opt out of such monitoring. And adjust the level of detail of the information that is sent to the napari server.
```
DragaDoncila:
```suggestion
Still, users need to be able to opt out of such monitoring, and adjust the level of detail of the information that is sent to the napari server.
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+ 
+  
+ ## Detailed Description
+
+`napari-telemetry` will be a package responsible for collecting and sending telemetry data to the napari server. It will be installed after user confirmation. It will contain callbacks for collection data, and utils for its storage and sending. Also, this package will contain utils for validation if the user has agreed to telemetry.
```
DragaDoncila:
```suggestion
`napari-telemetry` will be a package responsible for collecting and sending telemetry data to the napari server. It will be installed after user confirmation. It will contain callbacks for data collection, and utils for storage and sending. Also, this package will contain utils for validating if the user has agreed to telemetry. 
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+Also collecting information about data types and their size will provide valuable information about the typical use cases of napari.
+
+Still, a user need to be able to opt out of such monitoring. And adjust the level of detail of the information that is sent to the napari server.
+ 
+  
+ ## Detailed Description
+
+`napari-telemetry` will be a package responsible for collecting and sending telemetry data to the napari server. It will be installed after user confirmation. It will contain callbacks for collection data, and utils for its storage and sending. Also, this package will contain utils for validation if the user has agreed to telemetry. 
+
+In the main package, there is a need to add code to ask users if they want to enable telemetry. This 1code should be executed only once per environment.
```
DragaDoncila:
```suggestion
In the main package, there is a need to add code to ask users if they want to enable telemetry. This code should be executed only once per environment.
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+
+Telemetry should contain following way to disable it:
```
DragaDoncila:
```suggestion
Telemetry should contain following ways to disable it:
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+
+1. uninstall `napari-telemetry` package
+2. Environment variable `NAPARI_TELEMETRY=0`
+3. Full list of endpoints used for collecting telemetry, that could be filtered on the firewall level.
+
+The user should be able to adjust the telemetry level of detail. The following levels are proposed:
+
+1. `none` - no telemetry is collected
+2. `basic` - information about the napari version, python version, OS, and CPU architecture is collected and if it is the first report by the user. There is also a user identifier created based on computer details that will be rerendered each week to prevent tracking the user, but allow to not count a user multiple times. 
+3. `middle` - same as in `basic` but also information about the list of installed plugins and their versions is collected. We take care to not collect data about plugins that are not indented to be public.
```
DragaDoncila:
```suggestion
3. `middle` - same as in `basic` but also information about the list of installed plugins and their versions is collected. We take care to not collect data about plugins that are not intended to be public. 
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+1. uninstall `napari-telemetry` package
+2. Environment variable `NAPARI_TELEMETRY=0`
+3. Full list of endpoints used for collecting telemetry, that could be filtered on the firewall level.
+
+The user should be able to adjust the telemetry level of detail. The following levels are proposed:
+
+1. `none` - no telemetry is collected
+2. `basic` - information about the napari version, python version, OS, and CPU architecture is collected and if it is the first report by the user. There is also a user identifier created based on computer details that will be rerendered each week to prevent tracking the user, but allow to not count a user multiple times. 
+3. `middle` - same as in `basic` but also information about the list of installed plugins and their versions is collected. We take care to not collect data about plugins that are not indented to be public. 
+4. `full` - same as in `middle` but also information about plugin usage by binding to app-model and collect information about called plugins command. Also basic information about data like type (`np.ndarray`, `dask.array`, `zarr.Array`, etc.) and its size is collected.
```
DragaDoncila:
```suggestion
4. `full` - same as in `middle` but also collects information about plugin usage by binding to app-model and logging plugin commands used. Also basic information about data like type (`np.ndarray`, `dask.array`, `zarr.Array`, etc.) and its size is collected.
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+
+There should be a visible indicator that telemetry is enabled (for example on the status bar). 
+
+The second part of this work should be setup the server to collect telemetry data. Next to collecting data, it should provide a basic public dashboard that will allow a community to see aggregated information.
```
DragaDoncila:
```suggestion
The second part of this work should be to setup the server to collect telemetry data. After collecting data, it should provide a basic public dashboard that will allow the community to see aggregated information.
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+I propose to have the following data retention policy:
+
+1) Up to 2 weeks for logs.
+2) up 2 months of raw data (1 month of collection, then aggregation and time to validate aggregated data),
+3) infinite of aggregated data.
+  
+## Privacy assessment
+
+During the preparation of this NAP we assume that none of the collected data will be presented in 
+a form that allows to identify a single user or identify a research area of user. We also select a set of data that will be collected to minimize the possibility of reval of fragile data, but it is impossible to guarantee that it will not be possible to identify a single user (for example checking plugins combination).
```
DragaDoncila:
```suggestion
a form that allows to identify a single user or identify a research area of user. We also select a set of data that will be collected to minimize the possibility of revealing fragile data, but it is impossible to guarantee that it will not be possible to identify a single user (for example by checking installed plugin combinations).
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+
+Because of this, we decided to not publish raw data and only show aggregated results. The aggregation will be performed using scripts. Napari core-devs will access raw data only when there will be some errors in the aggregation process.
```
DragaDoncila:
```suggestion
Because of this, we propose to not publish raw data and only show aggregated results. The aggregation will be performed using scripts. Napari core devs will access raw data only if there are errors in the aggregation process.
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+Visualizations:
+https://github.com/grafana/grafana
+  
+
+  
+## Implementation 
+
+The main thing for implementation should be the low cost of maintenance. So the solution should be as simple as possible. We could either use existing solutions on the server side or implement our own.
+
+The benefit of the existing solution is that most of the work is already done. The downside is that it may require additional cost of maintenance. This cost may be caused by many features that are not needed for napari and could increase the risk of leaking data.  As I check they are implemented in techniques that are not familiar to napari core-devs. So if there will be a decision to use them we should select an SAS solution that will be maintained by the company.
```
DragaDoncila:
```suggestion
The benefit of existing solutions is that most of the work is already done. The downside is that it may require additional cost of maintenance. This cost may be caused by many features that are not needed for napari and could increase the risk of leaking data.  Quick checks of their code revealed they are implemented in techniques that are not familiar to napari core devs. So, if we decide to use them, we should select an SAS solution that will be maintained by the company.
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+
+
+For the current, I suggest creating a simple REST API server for collecting the data.
```
DragaDoncila:
```suggestion
For now, I suggest creating a simple REST API server for collecting the data. 
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+
+
+For the current, I suggest creating a simple REST API server for collecting the data. 
+It could be a simple Python FastAPI server that will store data in the SQLite database.
+
+Data for aggregation should be extracted from the database using a script running on the same machine.
+
+The output of the aggregation script should be loaded to some existing visualization tool, like grafana.
+
+It may be nice to host them on separate servers, then even if the data presented on the dashboard will be compromised, the raw data will be not exposed to the world.
```
DragaDoncila:
```suggestion
It may be nice to host raw and aggregate data on separate servers - then even if the data presented on the dashboard is compromised, the raw data will be not exposed to the world.
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+
+Having both server and aggregation scripts in Python will reduce maintenance costs for napari core-devs.
```
DragaDoncila:
```suggestion
Having both server and aggregation scripts in Python will reduce maintenance costs for napari core devs.
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+
+We should register the `telemetry.napari.org` domain and use it for the server. The main page contains this NAP and a link to the summary dashboard.
```
DragaDoncila:
```suggestion
We should register the `telemetry.napari.org` domain and use it for the server. The main page will contain this NAP and a link to the summary dashboard.
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+
+
+The main part of the application side should be implemented in `napari-telemetry` package. 
+The package should not report in stream mode, but collect data on the disk and send it in batches. This will reduce the risk of leaking data. The package should implement a utility to allow users to preview collected data before sending it to the server.
+
+In the napari itself following changes should be implemented:
```
DragaDoncila:
```suggestion
In napari itself, the following changes should be implemented:
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+
+1) The indicator that shows the telemetry status
+2) The dialog that asks a user if they want to enable telemetry
+3) code to check if telemetry is enabled (to not load the `napari-telemetry` package if it is disabled)
+4) code required to init `napari-telemetry` package
+
+
+## Potential problems 
+
+There is a risk that someone may try to highjack the telemetry module name to have code executed at every napari start.
```
DragaDoncila: Can we expand on this please? I don't understand how this would happen and how examining plugin dependencies could avoid it.

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,181 @@
...
+
+ It will also require updates on the Napari side as currently we only communicate with the update server when a user opens the plugin manager. Also, to have proper information about installed plugins we will need to send information about the list of installed plugins instead of just downloading the information about all plugins from the server. 
+
+ As this solution provides less information, does not allow for opt-out and could cause blacklisting of the update server IP address, I do not recommend it.
+
+ But based on talks that happen during the discussion we may think about more frequent checks for updates to inform users that they could update their Napari or plugin version. For such a change we need to update our update server to provide information per Python version (as some plugins could drop old Python earlier).
+
+
+The second alternative is use a third-party solution like [plausable.io](https://plausible.io/). But from my perspective, it is harder to adjust a set of data that is collected as these services are designed to monitor webpages.
```
DragaDoncila:
```suggestion
The second alternative is use a third-party solution like [plausible.io](https://plausible.io/). But from my perspective, it is harder to adjust a set of data that is collected as these services are designed to monitor webpages. 
```

DragaDoncila (commented):
@Czaki this is looking great, thank you for all your work on this 🎉 ❤️ ! I've done a pass of suggestions for wording etc. Once those updates are made and I've made sure the formatting is correct, I'll be happy to approve the draft.

In terms of comments on the proposal itself these are my general thoughts:

- I would like to see us include in the NAP a process for getting community feedback and concerns before implementation. While the NAP is usually the way to publicize big changes like this, I think they're mostly watched/discussed by core devs. Since telemetry can be a controversial topic, I would like us to include a process here that ensures the community is aware of and on board with the proposal
- I think before accepting the NAP, I would like to see a complete list (as complete as possible) of exactly what is collected with each level of telemetry, and potentially also a sample report, and what would show up on the public dashboards
-  We mention collecting information on what plugin contributions are used, but I think it would be great to collect information on what internal commands are used as well. It would give us great insight into popular features, potentially underused/underpublicized features and "danger zones" for API deprecations. Not sure to what extent this is possible, but once the app-model conversions are done there should be a few things we can learn?
- I would like us to expand a little on the motivation for telemetry beyond "gauging user numbers" which imo is the weakest motivation. For me it's really more about learning what plugins and parts of the codebase are seeing heavy usage, predicting the impact of changing APIs and gaining insight into what could be deprecated, changed, better documented, etc.

docs/naps/8-telemetry.md
Czaki: In general, there is no need for the Python package name will match the module name `scikit-image` vs `skimage`. So some package could contain `napari_telemetry` module that during installation will overwrite/create `napari_telemetry` module. 

Simple things that could be used for hardering:

1) check if `napari-telemetry` package is installed using `importlib.metadata` before try to import
2) use sha256 from package metadata to validate file before import. 

But same problem may happen with `napari_builtins` or even someone could try to edit and glue to `napari/__init__.py`, So I'm not sure. Maybe I should remove this part?

Czaki:
> I would like us to expand a little on the motivation for telemetry beyond "gauging user numbers" which imo is the weakest motivation. For me it's really more about learning what plugins and parts of the codebase are seeing heavy usage, predicting the impact of changing APIs and gaining insight into what could be deprecated, changed, better documented, etc.

This is mentioned in the motivation and scope paragraph. Should I rephase it? 



> * We mention collecting information on what plugin contributions are used, but I think it would be great to collect information on what internal commands are used as well

great idea. I will add this in the following days. 

I also agree with polishing the list of collected data. Opening this PR is way to open it on the community. I think, that after some polishing we should make a public announcement.

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,182 @@
...
+`napari-telemetry` will be a package responsible for collecting and sending telemetry data to the napari server. It will be installed after user confirmation. It will contain callbacks for data collection, and utils for storage and sending. Also, this package will contain utils for validating if the user has agreed to telemetry. 
+
+In the main package, there is a need to add code to ask users if they want to enable telemetry. This code should be executed only once per environment.
+
+Telemetry should contain following ways to disable it:
+
+1. Disable in settings
+2. uninstall `napari-telemetry` package
+3. Environment variable `NAPARI_TELEMETRY=0`
+4. Full list of endpoints used for collecting telemetry, that could be filtered on the firewall level.
```
Carreau:
```suggestion
5 . Disable at system-wide level in a way that is not user-controlable (Hpc/ other deployments).
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,182 @@
...
+
+There should be a visible indicator that telemetry is enabled (for example on the status bar). 
+
+The second part of this work should be to setup the server to collect telemetry data. After collecting data, it should provide a basic public dashboard that will allow the community to see aggregated information.
+
+I propose to have the following data retention policy:
+
+1) Up to 2 weeks for logs.
+2) up 2 months of raw data (1 month of collection, then aggregation and time to validate aggregated data),
+3) infinite of aggregated data.
```
Carreau: ~~Do you send data as they are created, or aggregate locally and send once ?~~

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,182 @@
...
+
+To have the possibility of sustainable development of the project we will either need to rely on paid contractors or on companies donating employee time managed by the core devs.
+
+Both scenarios require us to provide some information about the estimated number of users to prove to potential funders that their donation/grant will be used in a valuable way. 
+
+Adding the option for monitoring plugin usage allows us to identify the most important plugins and try to establish cooperation with their maintainers to reduce the probability that the plugin will not be ready for a new napari release. Such monitoring could contain not only the list of installed plugins but also which commands and contributions are used most often.
+
+Also collecting information about data types and their size will provide valuable information about the typical use cases of napari.
+
+Still, users need to be able to opt out of such monitoring, and adjust the level of detail of the information that is sent to the napari server.
```
Carreau: Detail, I would store the date when someone agreed, and maybe reconfirm every few month or so.

You might need a way for the remote to reply with a specific error code/version number that void the user consent, and retrigger asking consent as well.

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,182 @@
...
+3) infinite of aggregated data.
+  
+## Privacy assessment
+
+During the preparation of this NAP we assume that none of the collected data will be presented in 
+a form that allows to identify a single user or identify a research area of user. We also select a set of data that will be collected to minimize the possibility of revealing fragile data, but it is impossible to guarantee that it will not be possible to identify a single user (for example by checking installed plugin combinations).
+
+Because of this, we propose to not publish raw data and only show aggregated results. The aggregation will be performed using scripts. Napari core devs will access raw data only if there are errors in the aggregation process.
+
+We also will publish a list of endpoints for each level of telemetry, so the given level of telemetry could be blocked on the organization level (for example by the rule on the firewall).
```
Carreau: For everything that is plugin-related, you likely want to filter-out everything that is outside of a explicit-exclusion list. Local plugin names may leak internal information.

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Potential problems 
+
+There is a risk that someone may try to highjack the telemetry module name to have code executed at every napari start. 
+
+I do not expect that it is a high risk, but exists. We could address it by code signing. This will require additional procedures to protect private cryptographic keys.
+
+Another option is to scan public plugins and their dependencies. This is simpler but will require establishing additional communication channels to be able to warn users about the potential problem. 
+
+
```
Carreau, line 184: You will need to describe the way people can contact napari for their data to be removed. 

You have to be careful, login IP in the machine that collect logs may be considered collecting PII in some regions.

docs/naps/8-telemetry.md
Carreau: I don't think that's a problem. Someone could make a fake numpy...

docs/naps/8-telemetry.md
Czaki: I have clarified that plugin need to be uploaded to PyPI with the napari classifier (to be searchable in the napari plugin manager) and be in the stable version.

docs/naps/8-telemetry.md
Czaki: My Idea is to at the first phase, not to log IP. I do not see the benefit of such information and only risk. 

I ha e added GDPR section

docs/naps/8-telemetry.md
Czaki: Why do you prefer longer storage?

docs/naps/8-telemetry.md
Carreau: Sorry I was unclear, if for example you run the collection service behind nginx, the nginx logs may store IP. So you have to be careful that not only the service you write but part of the stack does not store data and/or that you can delete it.

Czaki: Some storage is required (Identifying spamming users), but it may be short with fast removal. 
And such things will not be propagated to the aggregator.

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,190 @@
...
+1. Disable in settings
+2. uninstall `napari-telemetry` package
+3. Environment variable `NAPARI_TELEMETRY=0`
+4. Full list of endpoints used for collecting telemetry, that could be filtered on the firewall level.
+
+The user should be able to adjust the telemetry level of detail. The following levels are proposed:
+
+1. `none` - no telemetry is collected
+2. `basic` - information about the napari version, python version, OS, and CPU architecture is collected and if it is the first report by the user. There is also a user identifier created based on computer details that will be rerendered each week to prevent tracking the user, but allow to not count a user multiple times. 
+3. `middle` - same as in `basic` but also information about the list of installed plugins and their versions is collected. We take care to not collect data about plugins that are not intended to be public, so we will not collect information about plugins searchable as napri plugin using plugin dialog or napri-hub. We also will not collect information about plugins that are installed in non stable version.
```
soupault: I believe `basic` and `middle` could be combined into a single level. From the dev's perspective, `basic` does not (arguably) bring much actionable information. From the user's perspective, these two levels are likely to be percieved as providing similar levels of anonymity.

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,190 @@
...
+I propose to have the following data retention policy:
+
+1) Up to 2 weeks for logs.
+2) up 2 months of raw data (1 month of collection, then aggregation and time to validate aggregated data),
+3) infinite of aggregated data.
+  
+## Privacy assessment
+
+During the preparation of this NAP we assume that none of the collected data will be presented in 
+a form that allows to identify a single user or identify a research area of user. We also select a set of data that will be collected to minimize the possibility of revealing fragile data, but it is impossible to guarantee that it will not be possible to identify a single user (for example by checking installed plugin combinations).
```
soupault: fragile -> sensitive

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,190 @@
...
+  
+## Implementation 
+
+The main thing for implementation should be the low cost of maintenance. So the solution should be as simple as possible. We could either use existing solutions on the server side or implement our own.
+
+The benefit of existing solutions is that most of the work is already done. The downside is that it may require additional cost of maintenance. This cost may be caused by many features that are not needed for napari and could increase the risk of leaking data.  Quick checks of their code revealed they are implemented in techniques that are not familiar to napari core devs. So, if we decide to use them, we should select an SAS solution that will be maintained by the company.
+
+
+For now, I suggest creating a simple REST API server for collecting the data. 
+It could be a simple Python FastAPI server that will store data in the SQLite database.
```
soupault, line 145: Any communication with the server needs to be encrypted to avoid MITM. I would specify explicitly the protocol indended to be used for that, i.e. SSH/HTTPS/other

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,190 @@
...
+## Potential problems 
+
+There is a risk that someone may try to highjack the telemetry module name to have code executed at every napari start. 
+
+I do not expect that it is a high risk, but exists. We could address it by code signing. This will require additional procedures to protect private cryptographic keys.
+
+Another option is to scan public plugins and their dependencies. This is simpler but will require establishing additional communication channels to be able to warn users about the potential problem. 
+
+
+## GDPR compatybility
```
soupault: -> compliance

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,190 @@
...
+
+I'm almost sure that we will not collect data that are covered by GDPR. But to get better atmosphere 
+we need to add instruction how user could retrive his unique identifier and setup a process 
+for requests to remove data from the server. It is not high propability of usage as life span of data is short,
+but we need to be prepared for such a situation. I suggest to use e-mail for that.
```
soupault, line 182: I fully agree on this. As it is currently written, the procedure does not collect any sensitive data in GDPR sense, yet it is better to be fully transparent and accountable with the users.

soupault (approved):

docs/naps/8-telemetry.md
DragaDoncila: agreed, I think this is overkill and not really related to the telemetry package specifically.

docs/naps/8-telemetry.md
Czaki: In theory middle allow to identification of user base on a unique set of plugins. Where base still allows to have basic knowledge about os, new python adaptation etc.

docs/naps/8-telemetry.md
Czaki: You mean to add new point?

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,191 @@
...
+ :Resolution: <url> (required for Accepted | Rejected | Withdrawn) 
+ :Resolved: <date resolved, in yyyy-mm-dd format> 
+ :Status: Draft
+ :Type: Standards Track 
+ :Version effective: <version-number> (for accepted NAPs) 
+ ``` 
+  
+ ## Abstract 
+  
+ This NAP is describes why telemetry is helpful to the napari project and the architecture and solutions selected to maximize the privacy of our users.
```
DragaDoncila:
```suggestion
 This NAP describes how telemetry would be used by the napari project and the architecture and solutions proposed to maximize the privacy of our users. 
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,191 @@
...
+ 
+ ## Motivation and Scope
+
+With the growth of napari, the standard feedback loop through napari community meetings and napari-related events at conferences has reached its capacity. Also, we collect many feature requests for which we cannot find volunteers for implementation. 
+
+To have the possibility of sustainable development of the project we will either need to rely on paid contractors or on companies donating employee time managed by the core devs.
+
+Both scenarios require us to provide some information about the estimated number of users to prove to potential funders that their donation/grant will be used in a valuable way. 
+
+Adding the option for monitoring plugin usage allows us to identify the most important plugins and try to establish cooperation with their maintainers to reduce the probability that the plugin will not be ready for a new napari release. Such monitoring could contain not only the list of installed plugins but also which commands and contributions are used most often.
```
DragaDoncila:
```suggestion
Adding the option for monitoring plugin usage allows us to identify heavily used plugins and try to establish cooperation with their maintainers to reduce the probability that the plugin will not be ready for a new napari release. Such monitoring could contain not only the list of installed plugins but also which commands and contributions are used most often.
```

docs/naps/8-telemetry.md
DragaDoncila: Possibly we should merge the two sentences:

4. System-wide disablement e.g. via firewall filtering for hpc or other environments

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,191 @@
...
+ ## Detailed Description
+
+`napari-telemetry` will be a package responsible for collecting and sending telemetry data to the napari server. It will be installed after user confirmation. It will contain callbacks for data collection, and utils for storage and sending. Also, this package will contain utils for validating if the user has agreed to telemetry. 
+
+In the main package, there is a need to add code to ask users if they want to enable telemetry. This code should be executed only once per environment.
+
+Telemetry should contain following ways to disable it:
+
+1. Disable in settings
+2. uninstall `napari-telemetry` package
```
DragaDoncila:
```suggestion
2. Uninstall `napari-telemetry` package
```

docs/naps/8-telemetry.md
DragaDoncila: I think the proposal is that we reconfirm opt-in every few months or so. I think this is a good idea and worth adding.

docs/naps/8-telemetry.md
DragaDoncila: I agree @Czaki I would like to keep the distinction there.

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,191 @@
...
+3. Environment variable `NAPARI_TELEMETRY=0`
+4. Full list of endpoints used for collecting telemetry, that could be filtered on the firewall level.
+
+The user should be able to adjust the telemetry level of detail. The following levels are proposed:
+
+1. `none` - no telemetry is collected
+2. `basic` - information about the napari version, python version, OS, and CPU architecture is collected and if it is the first report by the user. There is also a user identifier created based on computer details that will be rerendered each week to prevent tracking the user, but allow to not count a user multiple times.
```
DragaDoncila:
```suggestion
2. `basic` - information about the napari version, python version, OS, and CPU architecture is collected and if it is the first report by the user. There is also a user identifier created based on computer details that will be regenerated each week to prevent tracking the user, but allow us to accurately gauge individual user numbers. 
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,191 @@
...
+1. Disable in settings
+2. uninstall `napari-telemetry` package
+3. Environment variable `NAPARI_TELEMETRY=0`
+4. Full list of endpoints used for collecting telemetry, that could be filtered on the firewall level.
+
+The user should be able to adjust the telemetry level of detail. The following levels are proposed:
+
+1. `none` - no telemetry is collected
+2. `basic` - information about the napari version, python version, OS, and CPU architecture is collected and if it is the first report by the user. There is also a user identifier created based on computer details that will be rerendered each week to prevent tracking the user, but allow to not count a user multiple times. 
+3. `middle` - same as in `basic` but also information about the list of installed plugins and their versions is collected. We take care to not collect data about plugins that are not intended to be public, so we will not collect information about plugins searchable as napri plugin using plugin dialog or napri-hub. We also will not collect information about plugins that are installed in non stable version.
```
DragaDoncila:
```suggestion
3. `middle` - same as in `basic` but information about the list of installed plugins and their versions is also collected. We take care to not collect data about plugins that are not intended to be public, so we will only collect information about plugins searchable as using plugin dialog or napari hub. We also will not collect information about plugins that are installed in non stable version.
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,191 @@
...
+4. `full` - same as in `middle` but also collects information about plugin usage by binding to app-model and logging plugin commands used. Also basic information about data like type (`np.ndarray`, `dask.array`, `zarr.Array`, etc.) and its size is collected.
+
+There should be a visible indicator that telemetry is enabled (for example on the status bar). 
+
+The second part of this work should be to setup the server to collect telemetry data. After collecting data, it should provide a basic public dashboard that will allow the community to see aggregated information.
+
+I propose to have the following data retention policy:
```
DragaDoncila:
```suggestion
We propose the following data retention policy:
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,191 @@
...
+
+1) Up to 2 weeks for logs.
+2) up 2 months of raw data (1 month of collection, then aggregation and time to validate aggregated data),
```
DragaDoncila:
```suggestion
2) Up 2 months of raw data (1 month of collection, then aggregation and time to validate aggregated data).
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,191 @@
...
+3) infinite of aggregated data.
```
DragaDoncila:
```suggestion
3) Infinite of aggregated data.
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,191 @@
...
+https://opentelemetry.io/
+
+Visualizations:
+https://github.com/grafana/grafana
+  
+
+  
+## Implementation 
+
+The main thing for implementation should be the low cost of maintenance. So the solution should be as simple as possible. We could either use existing solutions on the server side or implement our own.
```
DragaDoncila:
```suggestion
The key consideration for implementation should be the low cost of maintenance. So the solution should be as simple as possible. We could either use existing solutions on the server side or implement our own.
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,191 @@
...
+
+I do not expect that it is a high risk, but exists. We could address it by code signing. This will require additional procedures to protect private cryptographic keys.
+
+Another option is to scan public plugins and their dependencies. This is simpler but will require establishing additional communication channels to be able to warn users about the potential problem. 
+
+
+## GDPR compliance
+
+I'm almost sure that we will not collect data that are covered by GDPR. But to get better atmosphere 
+we need to add instruction how user could retrive his unique identifier and setup a process
```
DragaDoncila:
```suggestion
we need to add instruction how user could retrieve his unique identifier and setup a process 
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,191 @@
...
+I do not expect that it is a high risk, but exists. We could address it by code signing. This will require additional procedures to protect private cryptographic keys.
+
+Another option is to scan public plugins and their dependencies. This is simpler but will require establishing additional communication channels to be able to warn users about the potential problem. 
+
+
+## GDPR compliance
+
+I'm almost sure that we will not collect data that are covered by GDPR. But to get better atmosphere 
+we need to add instruction how user could retrive his unique identifier and setup a process 
+for requests to remove data from the server. It is not high propability of usage as life span of data is short,
```
DragaDoncila:
```suggestion
for requests to remove data from the server. It is not high probability of usage as life span of data is short,
```

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,191 @@
...
+## Future Work 
+
+A nice extension may be the ability for the steering council to create a certificate of telemetry output that could be given to plugin maintainers to prove to supervisors that their plugin is used by the community. 
+
+  
+ ## Alternatives 
+
+ During the discussion, there is a proposal to use the same approach as used in ImageJ. 
+ 
+ Mean that instead of implementing telemetry on the client side we could implement it on the update server side. The advantage and disadvantage of such a solution is that no user could opt out of telemetry. Also, such a method could potentially provide information about the Python version, napari version and list of installed plugins. All others will require a mechanism from this NAP.
```
DragaDoncila:
```suggestion
This would mean instead of implementing telemetry on the client side we could implement it on the update server side. The advantage and disadvantage of such a solution is that no user could opt out of telemetry. Also, such a method could potentially provide information about the Python version, napari version and list of installed plugins. All others will require a mechanism from this NAP.
```

DragaDoncila (commented):
@Czaki I have done a final pass and once that's done I think this should go in in draft mode. Once we have some sample reports to look at, we should be able to add any other necessary detail. It would be nice to have a basic architecture diagram as well.

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,191 @@
...
+`napari-telemetry` will be a package responsible for collecting and sending telemetry data to the napari server. It will be installed after user confirmation. It will contain callbacks for data collection, and utils for storage and sending. Also, this package will contain utils for validating if the user has agreed to telemetry. 
+
+In the main package, there is a need to add code to ask users if they want to enable telemetry. This code should be executed only once per environment.
+
+Telemetry should contain following ways to disable it:
+
+1. Disable in settings
+2. uninstall `napari-telemetry` package
+3. Environment variable `NAPARI_TELEMETRY=0`
+4. Full list of endpoints used for collecting telemetry, that could be filtered on the firewall level.
```
Czaki:
```suggestion
4. System-wide disablement e.g. via firewall filtering for hpc or other environments.
```

@DragaDoncila This?

docs/naps/8-telemetry.md
Czaki: I still do not follow. 
Did we want to reconfirm every three months (for example), or when we update the telemetry code, so change the amount of collected information?

Czaki:
dependency check is failing because of bug in main repo https://github.com/gregsdennis/dependencies-action/issues/24

docs/naps/8-telemetry.md
DragaDoncila: I think it should be `min(3_months, package_update)` for re-confirming, personally. 

The dialog for whether you want to continue sending telemetry could show an example of a recent report that had been sent, so the user can gauge their participation based on real data. We could always add a `Don't ask me again` option as well of course.

docs/naps/8-telemetry.md
DragaDoncila: @Czaki yep!

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,191 @@
+:orphan: 
+  
+ (nap-8)= 
+  
+ # NAP-8 — Telemetry
+  
+ ```{eval-rst} 
+ :Author: Grzegorz Bokota 
+ :Created: <date created on, in yyyy-mm-dd format>
```
DragaDoncila: @Czaki you should update this, I think it's fine to date it as the day you opened the PR.

docs/naps/8-telemetry.md
```diff
@@ -0,0 +1,265 @@
...
+There is a risk
+that someone may try to high-jack the telemetry module name to have code executed at every napari start. 
+
+I do not expect that it is a high risk, but exists.
+We could address it by code signing.
+This will require additional procedures to protect private cryptographic keys.
+
+Another option is to scan public plugins and their dependencies. 
+This is simpler,
+but will require establishing additional communication channels to be able to warn users about the potential problem.
```
Czaki: I started thinking that this section is obsolete and could be removed from NAP.

DragaDoncila: @Czaki agreed, I don't think this is necessary.

DragaDoncila (approved):
@Czaki I think this is fine to go in whenever in draft mode. Thanks for all your work! Any further updates can be made in future PRs as is standard for NAPs

brisvag (approved):
Agreed, let's get this in and iterate in a PR.

PR #230, opened 2023-08-16
Unversioned pages: overwrite home and developer docs on stable with dev version

### LINKED ISSUE #86: Unversioned (permanent) home page
## 📚 New content request
During today's Documentation working group meeting we have discussed using a permanent, unversioned home page to avoid versioning issues.

### Outline
Can be done together with napari/napari#4275

melissawm:
# Description
Creates pages that are always synced to latest by copying over index.html and the contents of the developers/ folder over their stable version.

There are maybe easier versions of this, happy to change strategy in that case 😄 

I've tested this successfully on my fork at melissawm/napari.github.io 

## Type of change

# References
Fixes #86 

## Final checklist:
- [?] My PR is the minimum possible work for the desired functionality

### DIFF
--- a/.github/workflows/deploy_docs.yml
+++ b/.github/workflows/deploy_docs.yml
@@ -15,9 +15,11 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+
   build-and-deploy:
     name: Build & Deploy
     runs-on: ubuntu-latest
+
     steps:
       - name: Clone docs repo
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy_docs.yml
```diff
@@ -72,3 +74,43 @@ jobs:
...
+
+    - name: Download artifact
+      uses: actions/download-artifact@master
+      with:
+        name: permanent_docs
+
+    - name: Copy permanent files to napari.github.io/stable
+      uses: dmnemec/copy_file_to_another_repo_action@main
+      env:
+        API_TOKEN_GITHUB: ${{ secrets.API_TOKEN_GITHUB }}
```
melissawm, line 109: This token needs to be created on the napari.github.io repo

.github/workflows/deploy_docs.yml
```diff
@@ -72,3 +74,43 @@ jobs:
...
+      with:
+        source_file: ${{ matrix.source_files }}
+        destination_repo: 'napari/napari.github.io'
+        destination_folder: 'stable'
+        destination_branch: 'gh-pages'
+        user_email: 'melissawm@gmail.com'
+        user_name: 'melissawm'
```
melissawm, line 116: This needs to be replaced by the appropriate credentials

.github/workflows/deploy_docs.yml
psobolewskiPhD: I can make a secret within *this* repo (napari/docs) but not napari.github.io
Any chance the existing token used in the last deploy step: `deploy_key: ${{ secrets.ACTIONS_DEPLOY_KEY }}` could work?

psobolewskiPhD:
Melissa walked me through this PR and there are some notes in the docs agenda:
https://hackmd.io/10DKrIMhREmk9sa_QRirlg?view#2023-08-16

It's quite clever and pretty safe: the unmodified stable and dev are both built and deployed, we're just adding a copy after, so if we wanted to abandon this strategy, it would just require commenting that out of the action and next merge to either repo would return the current status quo.

Unless someone has some other suggestion, I'd be in favor of merging and then we can figure out next steps?

.github/workflows/deploy_docs.yml
melissawm: we can try, I'm not sure! As long as it has the same repo permissions I think that should work?

melissawm:
To clarify: the stable docs are not built every time - they were built once and now are in git history. So we could bring stable back by checking out the last commit before the overwrite. Still, it's there!

melissawm:
After discussion, @Czaki suggested having two new jobs:
- One that builds the latest docs, and copies index.html and the developer documentation files over stable, as discussed previously;
- A second one that checks out the docs for the specific stable version (including those sources) and builds them too.
This is to prevent creating broken links and inconsistencies between versions.

Another tought was that we should have an extra check on docs PRs about broken links (either internal, using the sphinx linkchecker builder, or external, using something like the Lychee tool) to prevent merging PRs that break the docs.

Please let me know if I forgot anything cc @psobolewskiPhD @DragaDoncila

melissawm:
I am closing this PR in favor of https://github.com/napari/napari.github.io/pull/393 - if we decide to reevaluate I can always reopen.

PR #232, opened 2023-08-18
Update Kyle's tag on the core devs page

seankmartin:
# Description
I noticed that Kyle's tag on the core devs page was linking to the right place, but the display name was wrong. So I've updated that here.


Czaki (approved):

jni:
Oops! Thanks! 😅

PR #236, opened 2023-09-01
Remove sub-sub section heading from rendering guide

dstansby:
# Description
The block of text under this heading doesn't refer to time-series data. While it's applicable to time-series data, it's also applicable to 3D volume data, so it seemed easiest just to remove the heading.

## Type of change

# References

## Final checklist:

### DIFF
--- a/docs/guides/rendering.md
+++ b/docs/guides/rendering.md
@@ -27,8 +27,6 @@ it will call {meth}`~napari.layers.Image.on_chunk_loaded` with
 the loaded data. The next frame {class}`~napari.layers.Image`
 can display the new data.
 
-### Time-series data
-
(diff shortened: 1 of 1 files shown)


brisvag (approved):

psobolewskiPhD (approved):
Huh, that heading is kinda oddly out of place isn't it.
Thanks for catching this.

Small aside--not blocking, just came to mind: @andy-sweet do we need to update any of this for the new async? If so, should we make an issue to track?

andy-sweet:
> Small aside--not blocking, just came to mind: @andy-sweet do we need to update any of this for the new async? If so, should we make an issue to track?

Yes, you're right. I added an issue to describe some thoughts and ideas there: https://github.com/napari/docs/issues/237

PR #238, opened 2023-09-18
Update selection instructions in Points tutorial for Shift-A keybinding

### LINKED ISSUE #1760: [Feature] Create a keyboard shortcut or a button to select all points loaded in napari GUI
## 🚀 Feature
Create a keyboard shortcut to select all points loaded in napari GUI

## Motivation
I am using cellfinder-napari which allows visualization and annotation of cells within 2d brain images. Using napari GUI, I can only select specific points within a specific plane. I cannot generalize my modifications to all the points in all the planes. 
Following snapshot shows few points selected after clicking on "select points"

image

But notice at the bottom right corner that this is for 131st (out of 549) plane. Any modifications I make to these points will not be reflected in the points that show up in any other plane (once I start going from left to right). So It can be frustration to individually select points again and again and make modification for each selection. 

## Pitch
A keyboard shortcut or a button that selects every single point so that the modifications (increase/decrease point size, change color, change shape) can be applied to everything and not just few selected points. 

## Alternatives

@sofroniewn has mentioned [here](https://github.com/brainglobe/cellfinder/issues/103#issuecomment-713703809) that this can be done via CLI using: 

```layer.selected_data = list(range(len(layer.data)))```

But I am not sure if I can use this technique as I am using an existing plugin. I haven't yet looked into writing a plugin myself but if this feature is hard to implement at the moment, then I will definitely look into it.

psobolewskiPhD:
Closes: https://github.com/napari/napari/issues/1760

# Description
This PR updates the Points tutorial to include the select-all-points-in-layer (Shift-A) keybinding. Also, both keybindings work in `(+)` mode, not just select mode.

### DIFF
--- a/docs/howtos/layers/points.md
+++ b/docs/howtos/layers/points.md
@@ -153,8 +153,10 @@ additional points, or by dragging a bounding box around the points you want to
 select. You can quickly select the select points tool by pressing the `S` key
 when the points layer is selected.
 
-You can select all the points in the currently viewed slice by clicking the `A`
-key if you are in select mode.
+Additionally, you can select all the points in the currently viewed slice by pressing 
+the `A` key and all the points in the layer (across all slices) using `Shift-A`. 
+Note: Pressing either keybinding again will toggle the selection, so you can select 
+all points in a layer and the *deselect* points from a slice. 
 
 Once selected you can delete the selected points by clicking on the delete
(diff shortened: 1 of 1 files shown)


melissawm (approved):
We discussed this in the docs working group meeting today, looks good to me. Thanks @psobolewskiPhD !

PR #241, opened 2023-09-19
fix outdated dimension sliders documentation

### LINKED ISSUE #240: Setting slider position of a multidimensional images
## 🐛 Bug
In the current [documentation](https://napari.org/stable/tutorials/fundamentals/viewer.html#dimension-sliders) it is stated that `current_step` of the dim slider returns a single integer and that it can be set by assigning a single integer to it. This is actually not the case as it returns a tuple of `int` and it can only be set by assigning a tuple of int.

## Additional context

The issue was first reported [here](https://github.com/napari/napari/issues/6236)

melonora:
Closes #240 

# Description
This PR updates the documentation under the section `Dimension Sliders`. Here it was still shown that `current step` would return
a single integer and could be set by assigning a single integer. I have checked whether this was also the case elsewhere in the documentation, but that is not the case.

### DIFF
--- a/docs/tutorials/fundamentals/viewer.md
+++ b/docs/tutorials/fundamentals/viewer.md
@@ -301,7 +301,7 @@ viewer.dims.current_step
 # to change the current position of this example to step 3
 viewer.dims.current_step = (3, 255, 255)
 ```
-Note that in this case two dimensions are displayed and thus the last two elements of the tuple have no effect.
+The length of the `current_step` tuple corresponds to the number of dimensions. Note that in this example, the last two dimensions are displayed (don't have a slider) and thus changing the last two elements of the tuple will have no effect until the axes order is changed.
 
 `viewer.dims.point` contains the position in world coordinates (i.e., including
 scale and translate transformations).
--- a/docs/tutorials/fundamentals/viewer.md
+++ b/docs/tutorials/fundamentals/viewer.md
@@ -295,15 +295,38 @@ nbscreenshot(viewer, alt_text="A 2d view of the moon on top of which is overlaid
 In this example there are three dimensions. In order to get or update the current position of the slider, use:
 
 ```python
-# to get the current position returned as tuple of length 3
+# To get the current position returned as tuple of length 3
 viewer.dims.current_step 
 
-# to change the current position of this example to step 3
+# To change the current position of this example to step 3
 viewer.dims.current_step = (3, 255, 255)
 ```
 The length of the `current_step` tuple corresponds to the number of dimensions. Note that in this example, the last two dimensions are displayed (don't have a slider) and thus changing the last two elements of the tuple will have no effect until the axes order is changed.
+The dimension order can be changed as follows:
 
-`viewer.dims.point` contains the position in world coordinates (i.e., including
+```python
+# To get the current dimension order as tuple of int
+viewer.dims.order
+
+# To change the current dimension order
+viewer.dims.order = (2, 1, 0)
+```
+In this case the third dimension will be controlled by the slider and the first and second dimension will be visible.
+Note that this has no effect on the order of `viewer.dims.current_step`. The first element still corresponds to the first dimension for example.
+
+The dimensions can also have labels. These can be set in the following manner:
+
+```python
+# To get the dimension labels
+viewer.dims.axis_labels
+
+# To set new axis labels
+viewer.dims.axis_labels = ("label_2", "label_1", "label_0")
+```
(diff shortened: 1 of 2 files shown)


docs/tutorials/fundamentals/viewer.md
```diff
@@ -259,6 +259,7 @@ viewer.layers.pop(i)
 One of the main strengths of **napari** is that it has been designed from the beginning to handle n-dimensional data. While much consumer photography is 2D and `RGB`, scientific image data can often be volumetric (i.e. 3D), volumetric timeseries (i.e. 4D), or even higher dimensional. **napari** places no limits on the dimensionality of its input data for all its layer types.
 
 Adding data with a dimensionality greater than 2D will cause dimension sliders to appear directly underneath the canvas and above the status bar. As many sliders as needed will appear to ensure the data can be fully browsed. For example, a 3D dataset needs one slider, a 4D dataset needs two sliders, and so on. The widths of the scroll bars of the dimension sliders are directly related to how many slices are in each dimension.
+To the left of each slider will be an integer indicating which dimension is being controlled by that slider. These integers are automatically updated when changing which dimensions are to be displayed.
```
psobolewskiPhD: Strictly speaking, there isn't always an int, because the axis slider could have a label?
Not sure if it adds confusion, but maybe it's worth adding mention of:
`viewer.dims.order` and `viewer.dims.axis_labels`
A bit of scope creep, but it fits this section of the tutorial and those arn't documented outside the API as far as I can tell.

docs/tutorials/fundamentals/viewer.md
```diff
@@ -291,14 +292,16 @@ viewer.add_image(blobs, name='blobs', opacity=0.5, colormap='red')
...
-viewer.dims.current_step
-# to change the current position
-viewer.dims.current_step = 3
+# to get the current position returned as tuple of length 3
+viewer.dims.current_step 
+
+# to change the current position of this example to step 3
+viewer.dims.current_step = (3, 255, 255)
 ```
+Note that in this case two dimensions are displayed and thus the last two elements of the tuple have no effect.
```
psobolewskiPhD:
```suggestion
The length of the `current_step` tuple corresponds to the number of dimensions. Note that in this example, the last two dimensions are displayed (don't have a slider) and thus changing the last two elements of the tuple will have no effect until the axes order is changed.
```

docs/tutorials/fundamentals/viewer.md
melonora: Will adjust when I get back home, but then we really need a method for changing 'current_step'.

psobolewskiPhD (approved):
Thanks! Sniped me @melonora 
What do you think of adding a line or two about about `viewer.dims.order` and `viewer.dims.axis_labels`? 
Can punt to another PR. Or I can make some suggestions later today.

melonora:
> Thanks! Sniped me @melonora What do you think of adding a line or two about about `viewer.dims.order` and `viewer.dims.axis_labels`? Can punt to another PR. Or I can make some suggestions later today.

Thanks for the quick review! I agree and with your earlier suggestion I think it would connect well.

melonora:
@psobolewskiPhD What do you think after the last change?

psobolewskiPhD:
Everything reads well! 
But now, given the fact that `viewer.dims.current_step` is always fixed order, maybe the `dims.order` stuff is confusing--it's not actually helpful--and spurious here? Maybe here just show `axis_labels` here--which are also always fixed order and helpful I think--and add a link to refer down to *Roll dimensions* for the 
> until axes order is changed

to https://napari.org/stable/tutorials/fundamentals/viewer.html#roll-dimensions
and then *here* add the order stuff as a programmatic way?

edit: Sorry if i'm being pedantic or waffling, I didn't read through well the first time I think.

melonora:
No problen:) Would line 315 (sorry github on phone since I won't open my laptop anymore this night) not be clear enough regarding the fixed order? We could separate it but it won't change the implementation and here it now expicitly tells the user at least.

psobolewskiPhD:
How about after diner I make a PR to your branch and we'll get a better sense of what we're looking at?

psobolewskiPhD:
Sorry! I'm not sure how I accidentally merged the PR to your branch!!
:grimacing:

melonora:
> Sorry! I'm not sure how I accidentally merged the PR to your branch!! 😬

haha well that is a surprise in the morning. No worries though, I think this indeed makes it more clear

PR #246, opened 2023-10-02
Update docs to suggest python 3.10 install

psobolewskiPhD:
In discussion about python version support, Matthias mentioned that more napari downloads are 3.9 than 3.10
https://github.com/napari/napari/pull/5768#issuecomment-1742691809

# Description
This PR updates the `python_version` variable in conf.py to `3.10` that is used for the myst substitution used in installation docs and contributing (developer env) guide.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -163,7 +163,7 @@ myst_enable_extensions = [
 myst_heading_anchors = 4
 
 version_string = '.'.join(str(x) for x in __version_tuple__[:3])
-python_version = '3.9'
+python_version = '3.10'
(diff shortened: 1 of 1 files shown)


psobolewskiPhD:
Should we also update the range to include 3.11, here:
https://github.com/napari/docs/blob/d9d061a646bec5024518cfd605a458c1b1a081c6/docs/conf.py#L167

DragaDoncila (approved):
Thanks @psobolewskiPhD! Re. your question about updating the range - I thought 3.11 was still experimental?

PR #249, opened 2023-10-13
Add NAP for multicanvas

aganders3:
Particular thanks to @brisvag and @melonora for previous work and discussions on this, and for agreeing to be co-authors. I am happy to absorb any criticism and direct praise their way 😄.

Thanks also to @andy-sweet and @Czaki, both of whom provided helpful discussion while I worked on this NAP.

Of course I expect many changes to this and look forward to feedback.


docs/naps/8-multiple-canvases.md
```diff
@@ -0,0 +1,261 @@
...
+
+### Out of Scope
+* Improvements to VisPy to support multiple views of the same `SceneGraph` (sharing data, saving VRAM) - for relevant discussion start with [vispy/#1992](https://github.com/vispy/vispy/issues/1992).
+* For now, canvas arrangement (for example: tiling behavior) will be handled in the view only (left to Qt or custom Qt widgets). Making this state (de)serializable is out of scope for this project, but may be relevant when implementing a “savable viewer state” feature.
+* Normalizing slice data for different layer types, though this may benefit in the course of this work.
+* Specific UI implementations will be explored as part of this work, but I expect UX and UI will be formalized later (possibly in a separate NAP).
+* Supporting alternative frontend (Qt) and backend (Vispy) frameworks. While this work should not make such tasks more difficult in the future, explicit consideration is out-of-scope until further progress is made in these areas.
+* [Non-goals also in NAP-3](https://napari.org/dev/naps/3-spaces.html#non-goals) are related but also considered out-of-scope here
+    * Separation of rendering information from the Layer(Data) model
```
andy-sweet: Would this benefit the work here or do you think of it as mostly orthogonal? Would a simultaneous effort help or hinder progress here?

docs/naps/8-multiple-canvases.md
```diff
@@ -0,0 +1,261 @@
...
+* Slice state will be moved off the layer as necessary, into new `LayerSlice` classes for each layer type
+    * This will be different for each Layer type - unifying the structure in the process may be a secondary benefit but is not the goal
+* Each VispyCanvas will hold a reference to a dedicated model class (`CanvasModel`)
+    * dims will move from `ViewerModel` -> `CanvasModel`
+    * camera will move from `ViewerModel` -> `CanvasModel`
+    * `layer_to_slice` will map each Layer (global list) to a `LayerSlice` (one per `CanvasModel`)
+* `ViewerModel` will own a *list* of `CanvasModel` objects
+* `QtViewer` will own a *list* of `VispyCanvas` objects
+* `VispyLayer` subclasses will hold references to their `Layer` (for rendering information) as well as a `LayerSlice` (for data)
+* The LayerSlicer will need to update the sync and async callbacks[^async-only] to pass a canvas parameter where the resulting sliced data will be stored, rather than storing it on the layer itself. Slice task cancellation logic will need to be revisited accordingly.
```
andy-sweet: One alternative here that came to mind is that each canvas (or maybe vispy canvas) could own its own layer slicer - that way you don't need to pass in the extra canvas parameter and you can do cancellation per canvas (which I think is desirable).

With lots of canvas, you might introduce too much concurrency, but in most cases, maybe that won't be too bad. And we may be able to serialize it with a shared resource/controller if needed.

docs/naps/8-multiple-canvases.md
```diff
@@ -0,0 +1,261 @@
...
+```{image} _static/multicanvas-napari-architecture-tomorrow.png
+---
+name: fig-2
+---
+Fig. 2: napari architecture tomorrow, with proposed changes from this NAP highlighted in orange.
+```
+
+> Note: in both diagrams, the `on_draw()` method on the `VispyCanvas` breaks MVC convention (view layer talks directly to the model layer). This is a separate/known issue and I believe is mostly only true for multiscale image layers right now. Changing this is considered out of scope for this project at this time.
+
+Slice state for each layer is currently stored on the Layer model. Again, see [NAP-4 for previous discussion](https://napari.org/stable/naps/4-async-slicing.html#existing-slice-state). This NAP proposes to move this state off the Layer instance, into a specific Layer Slice instance. This is what will allow multiple slices of a single layer to be visualized simultaneously. *Table 1* lists the attributes related to slice state that will be moved in this work from each Layer class into corresponding Layer Slice classes.
```
andy-sweet: Maybe I misread the out-of-scope section, but my impression was that moving slice state off the layer was a non-goal - is that wrong?

Or maybe this is the proposed approach that incidentally achieves that non-goal?

Anyway, I'm generally in favor of that move, but just wanted to clarify the proposal and intent here.

docs/naps/8-multiple-canvases.md
```diff
@@ -0,0 +1,261 @@
...
+* [OHIF/Cornerstone.js](https://www.cornerstonejs.org/live-examples/crosshairs)
+* [neuroglancer](https://neuroglancer-demo.appspot.com/#!%7B%22dimensions%22:%7B%22x%22:%5B8e-9%2C%22m%22%5D%2C%22y%22:%5B8e-9%2C%22m%22%5D%2C%22z%22:%5B8e-9%2C%22m%22%5D%7D%2C%22position%22:%5B2914.500732421875%2C3088.243408203125%2C4045%5D%2C%22crossSectionScale%22:3.762185354999915%2C%22projectionOrientation%22:%5B0.31435418128967285%2C0.8142172694206238%2C0.4843378961086273%2C-0.06040274351835251%5D%2C%22projectionScale%22:4593.980956070107%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%22precomputed://gs://neuroglancer-public-data/flyem_fib-25/image%22%2C%22tab%22:%22source%22%2C%22name%22:%22image%22%7D%2C%7B%22type%22:%22segmentation%22%2C%22source%22:%22precomputed://gs://neuroglancer-public-data/flyem_fib-25/ground_truth%22%2C%22tab%22:%22source%22%2C%22segments%22:%5B%2221894%22%2C%2222060%22%2C%22158571%22%2C%2224436%22%2C%222515%22%5D%2C%22name%22:%22ground-truth%22%7D%5D%2C%22showSlices%22:false%2C%22layout%22:%224panel%22%7D)
+
+## Implementation
+
+1. Introduce minimally disruptive `_canvases` attribute on the ViewerModel
```
andy-sweet: I'm fully in support of this ordering where implementation of architecture changes happen first to support new functionality later.

One mistake on the async slicing project is that I didn't do enough to prioritize things that way, which meant that things didn't end up quite as I'd hoped.

docs/naps/8-multiple-canvases.md
```diff
@@ -0,0 +1,261 @@
...
+> [name=Ashley A]
+
+### UI Design and Architecture
+Specific UI design and architecture remains to be determined. This will be explored as part of step 4 in the [Implementation Plan](#Implementation). UI design needs additional refinement and exploration, and this is expected to continue after basic/core implementation propsed in this NAP is complete. UI changes may also be described in a separate NAP along with a discussion of convenience functions and affordances for common operations. Some placeholder or experimental code will be used in the meantime as a prototype implementation.
+
+Some open questions here are (for example):
+* Should each canvas also have visible dims sliders, or can we keep one set of dims sliders that changes based on the active (selected) canvas?
+* What kind of cross-reference displays or tools should there be?
+    * though-plane slice indicators
```
andy-sweet: Typo?

```suggestion
    * through-plane slice indicators
```

docs/naps/8-multiple-canvases.md
```diff
@@ -0,0 +1,261 @@
...
+        * `VispyCanvas` will subscribe to relevant events, set data if corresponding to its own `CanvasModel`. Other canvases may also be interested in this event for example to update cross-reference overlays.
+
+3. Update `QtViewer` and `VispyCanvas` to support multiple canvases
+    * Still only displaying one canvas at a time in the main widget
+    * Update main widget as `ViewerModel` “active” canvas changes, storing additional canvases and swapping them out as necessary
+
+4. Update `QtViewer` to show multiple canvases simultaneously
+    * This is exploratory work at the moment, see [UI Architecture](#UI-Design-and-Architecture) section below
+
+[^async-only]: Depending on the timeline and prototype implemetiaton, it may be acceptable/prefereable for mutli-canvas feature to rely on (currently experimental) async slicing ([see NAP-4](https://napari.org/stable/naps/4-async-slicing.html)).
```
andy-sweet: I think that shouldn't be needed and if it is, I'm happy to help make that need go away.

Also, all the typos :D

```suggestion
[^async-only]: Depending on the timeline and prototype implementation, it may be acceptable/preferable for multi-canvas feature to rely on (currently experimental) async slicing ([see NAP-4](https://napari.org/stable/naps/4-async-slicing.html)).
```

docs/naps/8-multiple-canvases.md
```diff
@@ -0,0 +1,261 @@
...
+|| `_drag_start`
+| Shapes | `_data_dict`
+|| `_data_view`
+
+> TODO: add rough class definition(s) for Layer Slice(s)
+> Also - the various `_<Layer>SliceResponse` classes introduced by async slicing may already fill much of this role. Another option to consider here is to codify a protocol ([`typing.Protocol`](https://docs.python.org/3/library/typing.html#typing.Protocol)) for these classes. Even this protocol may not be necessary - early prototypes use these classes as-is with minimal modifications.
+> [name=Ashley A]
+
+### UI Design and Architecture
+Specific UI design and architecture remains to be determined. This will be explored as part of step 4 in the [Implementation Plan](#Implementation). UI design needs additional refinement and exploration, and this is expected to continue after basic/core implementation propsed in this NAP is complete. UI changes may also be described in a separate NAP along with a discussion of convenience functions and affordances for common operations. Some placeholder or experimental code will be used in the meantime as a prototype implementation.
```
andy-sweet: Is there any prototype code available to look at and play with? And/or some videos showing some of the proposals here at least kind-of in action?

Both could help to ground some of the proposals and discussion here.

docs/naps/8-multiple-canvases.md
```diff
@@ -0,0 +1,261 @@
...
+
+Slice state for each layer is currently stored on the Layer model. Again, see [NAP-4 for previous discussion](https://napari.org/stable/naps/4-async-slicing.html#existing-slice-state). This NAP proposes to move this state off the Layer instance, into a specific Layer Slice instance. This is what will allow multiple slices of a single layer to be visualized simultaneously. *Table 1* lists the attributes related to slice state that will be moved in this work from each Layer class into corresponding Layer Slice classes.
+
+> Note: some layers include "seleciton" information (Points, Shapes, and Tracks). How selection is handled in multicanvas view is TBD at this point.
+
+***Table 1** - Layer attributes that hold slice data. These attributes will be moved from the Layer onto individual Layer Slice objects (one Layer Slice per Layer per Canvas).*
+> TODO: make this a list or figure out a way to center it [name=Ashley A]
```
andy-sweet: I regretted adding something like this to the async slicing NAP, because it's just too much detail, so another option might be to remove it from the main NAP document and maybe link to it or shove it into an appendix.

andy-sweet (commented):
I took a first pass through this and left a few comments. Overall, I'm very excited to see something like this get approval and get implemented!

docs/naps/8-multiple-canvases.md
melonora: I think layergroups here could potentially provide a cleaner API, but overall the API is similar to current grid mode at least for multichannel view, thus could be an orthogonal effort.

docs/naps/8-multiple-canvases.md
aganders3: Yeesh those typos 🙈

I agree I don't think this is necessary. This was speculation before I started prototyping but so far it hasn't been an issue. I can remove this note.

docs/naps/8-multiple-canvases.md
aganders3: I don't think it's entirely orthogonal, but perhaps mostly. I think it depends on the scope of features we want. For example just below I propose for all canvases to share a common layer list, with independent slicing, camera, dimensionality, and layer visibility. I think this would be more relevant if supporting (for example) different colormaps for the same data in two canvases.

docs/naps/8-multiple-canvases.md
aganders3: Thanks - that's an interesting idea I hadn't considered. I was initially thinking the current LayerSlicer may expand to use a pool instead of a single thread.

I think you're right passing the full canvas is probably more than necessary, and trying to pass a canvas ID or something feels a bit hacky to me. One benefit of just passing the canvas object is that it would bring with it the camera, which might be useful for finding `corner_pixels` without waiting for the draw.

docs/naps/8-multiple-canvases.md
aganders3: Do you mean this non-goal?
> Normalizing slice data for different layer types, though this may benefit in the course of this work.

That's not really what I meant there, but also I think it's ambiguous enough warrant changing or removing. I was trying to say it is a non-goal to create a uniform/generic concept (or Protocol) of a "slice" - it's okay for the slices to be quite specific for each layer type.

docs/naps/8-multiple-canvases.md
aganders3: I found it helpful to write this out, but if it's too detailed I'm happy to move or remove it. It might just fit better in a PR with actual code changes.

aganders3:
> I took a first pass through this and left a few comments. Overall, I'm very excited to see something like this get approval and get implemented!

Thanks! I appreciate the support and feedback.

Czaki (dismissed):
The number 8 is for telemetry (#226), also please add a new NAP to the table of contents.

docs/naps/8-multiple-canvases.md
aganders3: My prototype/experimentation code is pretty early, but is here:
https://github.com/aganders3/napari/pull/13

docs/naps/9-multiple-canvases.md
```diff
@@ -0,0 +1,261 @@
+```{eval-rst}
+:Authors: Ashley Anderson <aandersoniii@chanzuckerberg.com>, Wouter-Michiel Vierdag, Lorenzo Gaifas
+:Created: 2023-08-04
+:Status: Draft
+:Type: Standards Track
+```
+
+## Abstract
+
+Current napari architecture supports a single canvas (viewbox) per viewer (window). Simultaneously showing multiple views of the same data generally necessitates opening an entirely new napari viewer window or low-level work with Qt widgets and private napari APIs. This wastes resources (primarily memory) and complicates interaction.
```
brisvag, line 30: I would make an extra effort from the get-go to define terminology:
- viewbox: specific concept in vispy, different from canvas
- viewer: I would specify "currently almost 1to1 matching to a window"

Opening multiple viewers is not only necessary for seeing the same data, but also for seeing different data, which does not incur the same memory complications and should be even easier, but is currently also impossible inside a single viewer.

docs/naps/9-multiple-canvases.md
```diff
@@ -0,0 +1,261 @@
...
+
+## Motivation and Scope
+The ability to view n-D data from multiple perspectives is a common feature request, and has proved useful in many other tools for data exploration and analysis. Here is a sampling of issues requesting support and discussing potential implementations:
```
brisvag: I would add the use cases with non-identical data, like viewing different segmentations overlaid to the same image, side by side.

docs/naps/9-multiple-canvases.md
```diff
@@ -0,0 +1,261 @@
...
+* [#760](https:////github.com/napari/napari/issues/760) Linked multicanvas support
+* [#662](https:////github.com/napari/napari/issues/662) Linked 2D views
+* [#561](https:////github.com/napari/napari/issues/561) multicanvas grid display for layers in Napari
+* [#1478](https:////github.com/napari/napari/issues/1478) Orthogonal viewer plugin
+
+Several plugins and examples have been created to address these limitations, for example:
+* [napari-3d-ortho-viewer](https://github.com/gatoniel/napari-3d-ortho-viewer/tree/main)
+* [multiple viewer widgets example](https://napari.org/stable/gallery/multiple_viewer_widget.html#sphx-glr-gallery-multiple-viewer-widget-py)
+
+This document is intended to cover what [#5348](https://github.com/napari/napari/issues/5348) refers to as "True Multicanvas". The changes to "Grid Mode" as described in that issue may be an important step along the way. This should be considered as part of the [implementation plan](#Implementation-Plan).
```
brisvag: Grid mode is orthogonal to this effort, IMO, at least as we intended it in that issue; it should be drop-in with the current grid mode, with no public API changes, at least at first. What do you think would be useful from the grid mode efforts to the "true multicanvas"?

docs/naps/9-multiple-canvases.md
```diff
@@ -0,0 +1,261 @@
...
+    * Separation of rendering information from the Layer(Data) model
+    * Window state restoration
+
+## Detailed Description
+
+### Requirements
+* The application data model (`ViewerModel` + Layers) shall support multiple canvases.
+* The application shall natively display multiple canvases simultaneously.
+    * There shall be a minimum of one canvas (current status) per viewer.
+    * There shall be a maximum of 8 canvases (tentative) per viewer.
```
brisvag: I guess these are numbers just to have some goal? I don't see why we would go for specificallty 8 other than "arbitrary"...

docs/naps/9-multiple-canvases.md
```diff
@@ -0,0 +1,261 @@
...
+* All canvases shall share a common layer list and (unsliced) layer data.
```
brisvag, line 66: This is my only disagreement with the current form of the NAP, I wrote more details in the general review comment.

docs/naps/9-multiple-canvases.md
```diff
@@ -0,0 +1,261 @@
...
+
+*Viewer* - The napari application main window, including canvas(es), dims sliders, layer list, layer controls, and dock widgets. Related is the `ViewerModel`, a class in napari that maintains the state of the Viewer.
+
+*Canvas* - The main napari data view (2D rectangle) where 2D or 3D slice data is displayed, and additional data views displayed within a single Viewer window.
+
+*Layer* - The base unit of the napari image data model. A ViewerModel maintains an ordered list of Layers that it may display on its Canvas.
+
+*Layer Slice* - A subset of data from a Layer, reduced to 2D or 3D (via slicing) for visualization.
+
+*Visual* - The corresponding visual representation of a Layer Slice displayed on a Canvas - currently all Visuals are implemented with [VisPy](https://vispy.org/).
```
brisvag: As mentioned above, I would move this higher up for clarity. I would also make sure to distinguish what's the *current* definition versus this NAP's *preferred* definition. I know it's semantics, but this stuff is confusing (as evidenced from previous conversations, such as the one in https://github.com/napari/napari/pull/4684.)

For example, we really need to clarify that when this nap says "Canvas", it's *not* a vispy/opengl/other-backend `Canvas`, but a more complex object that's much closer to what the `ViewerModel` is now.

docs/naps/9-multiple-canvases.md
```diff
@@ -0,0 +1,261 @@
...
+
+> TODO: add rough CanvasModel definition here [name=Ashley A]
+
+[^napari-lite]: A lightweight "canvas" might be relevant to the implementation of ["napari-lite"](https://github.com/napari/napari/issues/5940).
+
+An important consideration is to minimize breaking changes to the public napari API. While napari is still pre-1.0, there is already a healthy developing ecosystem of plugins, scripts, and users. Changes to the API may be necessary and should be made if they constitute improvements, but should be minimized and well documented.
+
+The concept of an "active" canvas will work in service of minimizing API changes. This will allow existing APIs on the main Viewer/ViewerModel to remain and simply delegate to the active canvas.
+
+In addition to maintaining the MVC architecture of napari, this proposal aims to maintain or improve decoupling of the UI framework (currently Qt), the visualization library (currently VisPy), and the napari core code.
```
brisvag: For us non-IT people, please use the full name once for "MVC" before the acronym :P

docs/naps/9-multiple-canvases.md
```diff
@@ -0,0 +1,261 @@
...
+Fig. 1: napari architecture today.
+```
+
+Figure 2 shows proposed changes (in orange) to the architecture to support multiple canvases. The new architecture is still following the MVC pattern. Again, this diagram only includes the Image layer type. Here is a summary of the planned changes:
+* Slice state will be moved off the layer as necessary, into new `LayerSlice` classes for each layer type
+    * This will be different for each Layer type - unifying the structure in the process may be a secondary benefit but is not the goal
+* Each VispyCanvas will hold a reference to a dedicated model class (`CanvasModel`)
+    * dims will move from `ViewerModel` -> `CanvasModel`
+    * camera will move from `ViewerModel` -> `CanvasModel`
+    * `layer_to_slice` will map each Layer (global list) to a `LayerSlice` (one per `CanvasModel`)
```
brisvag, line 112: Not sure I get this: `layer_to_slice` is supposed to be just a dict to keep track of which layer each slice in the canvasmodel comes from? If so, why is this necessary as opposed to the current event system? Or is it just a way to hold state about the sliced data *somehwere*? (in which case, my first instinct would be to hold sliced data *not* in the canvas, but higher up in the viewer (since the same layer might be present in different canvases).

EDIT: after reading below about the LayerSlicer updates, I'm more and more convinced that these slices should be held higher up for canvases to *take* from when necessary.

docs/naps/9-multiple-canvases.md
```diff
@@ -0,0 +1,261 @@
...
+name: fig-2
+---
+Fig. 2: napari architecture tomorrow, with proposed changes from this NAP highlighted in orange.
+```
+
+> Note: in both diagrams, the `on_draw()` method on the `VispyCanvas` breaks MVC convention (view layer talks directly to the model layer). This is a separate/known issue and I believe is mostly only true for multiscale image layers right now. Changing this is considered out of scope for this project at this time.
+
+Slice state for each layer is currently stored on the Layer model. Again, see [NAP-4 for previous discussion](https://napari.org/stable/naps/4-async-slicing.html#existing-slice-state). This NAP proposes to move this state off the Layer instance, into a specific Layer Slice instance. This is what will allow multiple slices of a single layer to be visualized simultaneously. *Table 1* lists the attributes related to slice state that will be moved in this work from each Layer class into corresponding Layer Slice classes.
+
+> Note: some layers include "seleciton" information (Points, Shapes, and Tracks). How selection is handled in multicanvas view is TBD at this point.
```
brisvag: I don't see any way/reason to have selection be different from one canvas to the other, so that seems like a pretty layer-tied concept rather than slice-tied. If one needs different selections at the same time on the same layer, they probably need two layers anyways?

docs/naps/9-multiple-canvases.md
```diff
@@ -0,0 +1,261 @@
...
+> [name=Ashley A]
+
+### UI Design and Architecture
+Specific UI design and architecture remains to be determined. This will be explored as part of step 4 in the [Implementation Plan](#Implementation). UI design needs additional refinement and exploration, and this is expected to continue after basic/core implementation propsed in this NAP is complete. UI changes may also be described in a separate NAP along with a discussion of convenience functions and affordances for common operations. Some placeholder or experimental code will be used in the meantime as a prototype implementation.
+
+Some open questions here are (for example):
+* Should each canvas also have visible dims sliders, or can we keep one set of dims sliders that changes based on the active (selected) canvas?
+* What kind of cross-reference displays or tools should there be?
+    * through-plane slice indicators
+    * three-point slice definition
```
brisvag, line 174: This is hard to do unless we assume a shared space between canvases, which I'm not sure we should do. Initially, I think we should leave this to the user.

Creating an orthoview example as part of this would would be a great way to test these ideas.

docs/naps/9-multiple-canvases.md
```diff
@@ -0,0 +1,261 @@
...
+
+1. Introduce minimally disruptive `_canvases` attribute on the ViewerModel
+    * Just a list (`EventedList`) of `CanvasModel` objects
+    * One canvas is “active”, relevant properties on ViewerModel (`camera`, `dims`) are delegated to the active canvas
+        * "active" canvas is just index 0 (alternatively: use `SelectableEventedList`)
+    * Only the active canvas is shown (minimal modifications to `QtViewer` and `VispyCanvas`)
+    * A single `QtDims` view (slicing sliders) is shown, updates depending on the active canvas
+    * Adjust event callbacks such that interactions (slicing, camera movement, ndisplay toggle) only apply to the active canvas
+    * Add public APIs to add/remove/rotate (change active) `_canvases`
+    * *Note: An improved version of "Grid Mode" may be able to be implemented with just these changes.*
```
brisvag: Which changes? I don't think grid mode is related to any of the above bullet points (at least as we intended it in the referenced issue: multiple viewboxes inside a single vispy canvas, sharing dims/layerlist/..., linked cameras, and just spreading the layers across viewboxes with the same logic as the current grid). If you mean a different kind of Grid Mode, please expand on it here for clarity.

docs/naps/9-multiple-canvases.md
```diff
@@ -0,0 +1,261 @@
...
+    * Update main widget as `ViewerModel` “active” canvas changes, storing additional canvases and swapping them out as necessary
+
+4. Update `QtViewer` to show multiple canvases simultaneously
+    * This is exploratory work at the moment, see [UI Architecture](#UI-Design-and-Architecture) section below
+
+[^async-only]: Depending on the timeline and prototype implementation, it may be acceptable/preferable for multi-canvas feature to rely on (currently experimental) async slicing ([see NAP-4](https://napari.org/stable/naps/4-async-slicing.html)).
+
+## Backward Compatibility
+
+Maintaining the proxy API on the viewer via the concepts of a main and/or active canvas will make this work mostly backward-compatible, though there will inevitably be some breaking changes. There will like be significant breaking changes to private APIs. For example if plugins are attempting to access slice data directly from a layer instance, it may no longer be as expected. If this is a large burden, it too may be mitigated by delegating from the layer to the slice corresponding to the main or active canvas.
```
brisvag:
```suggestion
Maintaining the proxy API on the viewer via the concepts of a main and/or active canvas will make this work mostly backward-compatible, though there will inevitably be some breaking changes. There will likely be significant breaking changes to private APIs. For example if plugins are attempting to access slice data directly from a layer instance, it may no longer be as expected. If this is a large burden, it too may be mitigated by delegating from the layer to the slice corresponding to the main or active canvas.
```

brisvag (approved):
Generally, great work on the NAP! Other than comments and clarifications here and there, I have one main disagreement, and it seems it comes from a misunderstanding when we discussed before, because I thing this is quite important. You say:

```
* All canvases shall share a common layer list and (unsliced) layer data.
```

I think we need to ensure from the get-go that canvases can hold different layer lists, but can also share some layers. This is key for example for applications like highly multiplexed data (@melonora can add some more specifics here if needed). I fear that leaving this out will make it harder to support it afterwards.

Anyways, approving as draft as per normal NAP procedure :)

aganders3:
Thanks for the review and great comments @brisvag! I will try to respond to each and update accordingly, but generally your comments cleared up a few things for me. We're not too far apart on anything here :).

docs/naps/8-multiple-canvases.md
andy-sweet: No, I meant this one.

> Separation of rendering information from the Layer(Data) model

But maybe you meant that only with respect to NAP-3.

Anyway, I also agree that a generic slice class/protocol should be a non-goal. Though we might consider during/after.

docs/naps/9-multiple-canvases.md
aganders3: I took a look back at the discussion of improved grid mode, and I agree. It _could_ be implemented with multiple canvases, but I think it makes sense to implement with vispy viewboxes. I've just removed the references here.

docs/naps/9-multiple-canvases.md
aganders3: Thanks - this makes sense. Removed!

docs/naps/9-multiple-canvases.md
aganders3: Yes sorry!

aganders3:
I updated a bit but not ready for your re-review @brisvag. I should be able to finish addressing your comments next week, but you can ignore GitHub notifications for now :)

docs/naps/8-multiple-canvases.md
aganders3: I see - I will try to make this more clear. The goal here is indeed to move the slice state off the layer. The non-goal was meant to indicate that the remaining attributes of the layer will not be separated. So for example this work does not necessarily make it easier to share data between two layers with different colormaps.

docs/naps/9-multiple-canvases.md
aganders3: I think you're correct about this - I'm going to remove this and we can revisit if necessary. I agree it makes  more sense for slices to be held somewhere where they could potentially be grabbed by more than one canvas.

Actually I will hold off making specific changes around this until we resolve the other points around shared/independent layer lists. It's not entirely related but I will adjust the text and diagrams for this once that is settled.

aganders3:
> I have one main disagreement, and it seems it comes from a misunderstanding when we discussed before, because I thing this is quite important. You say:
> > * All canvases shall share a common layer list and (unsliced) layer data.
> 
> I think we need to ensure from the get-go that canvases can hold different layer lists, but can also share some layers. This is key for example for applications like highly multiplexed data (@melonora can add some more specifics here if needed). I fear that leaving this out will make it harder to support it afterwards.

My way of addressing this use case was specifying layers would gain a per-canvas visibility setting. I think this is also where misalignment on https://github.com/napari/docs/pull/249#discussion_r1365484821 comes from; using a common layer list would also inherently share space between canvases. If you want a truly independent space, I don't see that much benefit over opening a separate viewer window.

Changing to fully independent layer lists would definitely offer more flexibility, and may actually make this implementation simpler in some ways. My concern is that it will make for a more difficult UI design, and may make it more difficult to implement certain common use cases later (for example, where you do want a common space between canvases). That said, I am definitely still open to this approach. You (and @melonora) may have a better understanding of the use-cases that would benefit.

docs/naps/9-multiple-canvases.md
aganders3: Yes I think this was just to have a goal. I could see setting some limit to ensure a good experience, but that could probably be part of the UI/UX layer instead. I've removed this for now.

docs/naps/9-multiple-canvases.md
aganders3: I moved the definitions up and tried to clear this up. I agree it's important thought so let me know if you have more suggestions!

PR #253, opened 2023-10-26
Installation guide: Mention slow first launch time

### LINKED ISSUE #222: Warn that first run of napari after install is slow (~30 s)
## 🧰 Task

In the various install/getting started docs we should warn the user that the first-run time is slow (>30 s for me) but that subsequent starts are markedly faster.
A user could think something went wrong with the install and cancel or restart terminal.

(This was feedback from the LatAm workshop on Aug. 7, 2023)

psobolewskiPhD:
Closes https://github.com/napari/docs/issues/222

# Description
This PR adds a `Note` that first launch may be slow on some platforms (macOS, Windows) and that anti-malware and security software may also result in slower launches.
It's quite disconcerting when running `napari` doesn't visually do anything immediately!

### DIFF
--- a/docs/tutorials/fundamentals/installation.md
+++ b/docs/tutorials/fundamentals/installation.md
@@ -172,9 +172,9 @@ the current release {{ napari_version }}, using command: `napari --version` .
 
 ````{note}
 On some platforms, particularly macOS and Windows, there may be a ~30 second 
-delay before the viewer appears on first launch. This is normal and subsequent
+delay before the viewer appears on first launch. This is expected and subsequent
 launches should be quick. However, anti-malware and other security software 
-measures may delay launches even after the first launch.
+measures may further delay launches—even after the first launch.
 ````
 
 ## Choosing a different Qt backend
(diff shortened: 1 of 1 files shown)


docs/tutorials/fundamentals/installation.md
```diff
@@ -170,6 +170,13 @@ the current release {{ napari_version }}, using command: `napari --version` .
 ````
 ![macOS desktop with a napari viewer window without any image opened in the foreground, and a terminal in the background with the appropriate conda environment activated (if applicable) and the command to open napari entered.](../assets/tutorials/launch_cli_empty.png)
 
+````{note}
+On some platforms, particularly macOS and Windows, there may be a ~30 second 
+delay before the viewer appears on first launch. This is normal and subsequent
```
brisvag: I don't know if I woudl call it "normal", but I get your point xD

melissawm (approved):
Looks good to me!

docs/tutorials/fundamentals/installation.md
psobolewskiPhD: Would you prefer "common" ?

brisvag: I would rather point out that this is an *issue* with those platforms, cause as it is it sounds like there's not expectation/hope of this to get fixed.

psobolewskiPhD: I mean there isn't anything we can do, so it's not *a napari issue*, which was the point of this Note.

jni (approved):

PR #255, opened 2023-10-27
Fix pydantic import

melissawm:
# References and relevant issues
Closes #254 

# Description
Uses napari pydantic compatibility shim in `update_preference_docs.py`.

PR #256, opened 2023-10-27
Add option to fail docs build on gallery example execution error

### LINKED ISSUE #6392: Gallery examples failing to build should be an error
## 🧰 Task

In #6391, we noted that the vortex example doesn't build correctly in the gallery. This was actually true from the get-go, but wasn't noticed because gallery build failures don't fail the docs build check — they simply get printed as warnings. Should we turn them into errors?

@melissawm @lucyleeow thoughts on this? 🙏

melissawm:
Closes https://github.com/napari/napari/issues/6392
Depends on https://github.com/napari/docs/pull/255

# Description
Added a sphinx gallery option to fail the docs build on example execution failure.

Right now, this _should_ fail on CI due to one of our example failing, but the docs build will anyway depend on the #255.


psobolewskiPhD:
Build is failing due to `sphinx_tags` release today, which has:
https://github.com/melissawm/sphinx-tags/commit/6351df0baefab2cc871db838cfec4d03adadbb8a

This import requires sphinx 5.1...
I **think** we can lift the pin here to <6 and in https://github.com/napari/napari-sphinx-theme

Edit: I'm going to test it...

psobolewskiPhD:
I bumped the pin in napari-sphinx-theme and in this repo to `<6`, and the docs built locally with the new `sphinx_tags`.
I'll open PRs.
So this one either needs to pin pack `sphinx_tags` or we go with the `<6`

melissawm:
Thanks for checking that @psobolewskiPhD - so I guess this leaves us a choice, really. Do we cut a new napari-sphinx-theme release, or do we go with the new theme infrastructure, that already supports any sphinx<6? (We can also just pin sphinx-tags for now)

psobolewskiPhD:
Yes! This seems as good a reason as any to cut over to your theme!

melissawm:
ok - it looks like this is "failing correctly" 😅

psobolewskiPhD:
So the fail--expected--is vortex right? And that is fixed by https://github.com/napari/napari/pull/6475

psobolewskiPhD:
Going to merge main, because vortex fix has been merged. Then if everything is green this should be good to go!

psobolewskiPhD (approved):
Looks good!
❤️

PR #274, opened 2023-11-20
Improve flow of install page

dstansby:
# Description
- If I'm installing the bundled app, I don't care about pre-requisites for installing the Python package, so move that section under the "Installing as a Pytho package" section
- Modify the header levels, so the second level now only has "Install as Python package", "Install as bundled app", "Next steps". In particular, two sections which are only relevant to the Python package route have been moved down a level to simplify the top level contents.

### DIFF
--- a/docs/tutorials/fundamentals/installation.md
+++ b/docs/tutorials/fundamentals/installation.md
@@ -25,11 +25,8 @@ This guide will teach you how to do a clean install of **napari** and launch the
 If you want to contribute code back into napari, you should follow the [development installation instructions in the contributing guide](https://napari.org/developers/contributing.html) instead.
 ```
 
-## Prerequisites
+## Install as Python package (recommended)
 
-Prerequisites differ depending on how you want to install napari.
-
-### Prerequisites for installing napari as a Python package
 This installation method allows you to use napari from Python to programmatically
 interact with the app. It is the best way to install napari and make full use of
 all its features.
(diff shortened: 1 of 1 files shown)


melissawm:
It looks like the pr-milestone-check is causing the CircleCI redirector action to be skipped. In any case, this looks good to me, thanks @dstansby !

melissawm (approved):

docs/tutorials/fundamentals/installation.md
```diff
@@ -42,15 +39,6 @@ You may also want:
...
-### Prerequisites for installing napari as a bundled app
-This is the easiest way to install napari if you only wish to use it as a standalone GUI app.
-This installation method does not have any prerequisites.
-
-[Click here](#install-as-a-bundled-app) to see instructions
-for installing the bundled app.
-
-## Install as Python package (recommended)
-
 Python package distributions of napari can be installed via `pip`, `conda-forge`, or from source.
```
psobolewskiPhD, line 42:
```suggestion
````{note} 
While we recommend installing as a Python package, you can also install napari as [a conda bundled app](#install-as-a-bundled-app).
````

Python package distributions of napari can be installed via `pip`, `conda-forge`, or from source.
```


I agree with the logic, but I'd like to keep the early mention of the bundle so it's a bit more accessible. How's this for a compromise?

docs/tutorials/fundamentals/installation.md
```diff
@@ -42,15 +39,6 @@ You may also want:
 - an environment manager like [conda](https://docs.conda.io/en/latest/miniconda.html) or
 [venv](https://docs.python.org/3/library/venv.html) **(Highly recommended)**
```
psobolewskiPhD, line 41:
```suggestion

````{note} 
While we recommend installing as a Python package, you can also install napari as [a conda bundled app](#install-as-a-bundled-app).
````


I agree with the logic change, but maybe we could still mention the bundled app here? 
How's this for a compromise?

Edit: I hope the suggestion is behaving...

psobolewskiPhD:
The whole CircleCI job isn't proc'ing...

psobolewskiPhD (approved):
I can't make the suggestion, but what are y'alls thoughts on using tab sets instead of the admonitions? (new line 58)

Like here:
https://napari.org/dev/developers/contributing.html#setting-up-a-development-installation
Code:
https://github.com/napari/docs/blob/1a3192fd7fc549b4ef8730dccae8bcbc673096c9/docs/developers/contributing.md?plain=1#L31C5-L54

melissawm:
Love the tabs, maybe in a follow-up?

Issue #291, opened 2023-12-05
Remove dependency install from `make docs`


lucyleeow:
## 🧰 Task

I am not familiar with the background on this but this issue can at least start a discussion.

From my experience, having dependency installation in the Makefile/`made docs` command is unusual. Thus it may cause unintended consequences for the user if they did not know it included installation. It also confusing (e.g., if used CI workflows) and prevents more fine control of dependency installation if used in CI workflows.  

Potentially we could at least make `make docs` NOT install dependency and have another command (e.g., `make install docs` / `make docs dep` ??) that will install the dependencies and build docs.

Discussed: https://github.com/napari/docs/issues/283#issuecomment-1839396682

lucyleeow:
cc @melissawm

melissawm:
As I commented in the other thread, I am +1 on this. I think it's kind of a separation of concerns issue. While we're at it, I also wonder about the [explicit npe2 cloning](https://github.com/napari/docs/blob/0c7029e5c9674e9d56c5efbec1022fa302661280/docs/_scripts/prep_docs.py#L15): could we have npe2 as a [submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules) instead?

lucyleeow:
Let me open an issue just for visibility :shrug:

#295

psobolewskiPhD:
This should have been closed by https://github.com/napari/docs/pull/602

PR #294, opened 2023-12-05
Update headless-gui action in build (fix screenshots)

### LINKED ISSUE #283: Widget screenshot in `magicgui.md` error
## 🐛 Bug

Widget screenshot in `magicgui.md` (in dev docs: https://napari.org/dev/guides/magicgui.html) looks wrong:

image

Not familiar with `magicgui` so not sure if this is due to our example or the screenshot tool?

### LINKED ISSUE #285: Problem screenshots in "Using magicgui in napari" in dev
## 🐛 Bug

The screenshots in "Using magicgui in napari" in [dev](https://napari.org/dev/guides/magicgui.html#using-magicgui-in-napari) look squished:

image

on [stable](https://napari.org/stable/guides/magicgui.html#using-magicgui-in-napari) the same screenshot:

image

Note this is the worst example, the others on the page look better but still squished compared to stable.

Could it be related to https://github.com/napari/docs/pull/207 @aganders3  ?

aganders3:
~~Hopefully f i x e s~~ Pretty sure now this fixes both #283 and #285 

# Description
This configures the [headless-gui](https://github.com/aganders3/headless-gui) action to not run the tiling window manager it uses by default. My thinking is the tiling behavior in the window manager is causing issues with certain screenshots. See https://github.com/napari/docs/issues/283#issuecomment-1841621067 for more details.

This also updates the action to a new version with a more common default screen size. This change in bit-depth fixes the messed up magicgui screenshot.

Edit: I'm still not _totally_ clear why the stable docs would look okay but I think this is worth trying.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,7 @@ jobs:
           name: Install napari-dev
           command: |
             . venv/bin/activate
-            python -m pip install -e "napari/[all]"
+            python -m pip install -e "napari/[pyside,dev]"
           environment:
             PIP_CONSTRAINT: napari/resources/constraints/constraints_py3.10_docs.txt
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,7 @@ jobs:
           name: Install napari-dev
           command: |
             . venv/bin/activate
-            python -m pip install -e "napari/[pyside,dev]"
+            python -m pip install -e "napari/[all]"
           environment:
             PIP_CONSTRAINT: napari/resources/constraints/constraints_py3.10_docs.txt
       - run:
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -62,8 +62,9 @@ jobs:
           run:  make -C docs docs
           # setting these empty stops from running the default (tiling) window manager
           # the window manager is not necessary for docs builds at this time
-          linux-pkgs: ""
-          linux-setup: ""
+          linux-pkgs: " "
+          linux-setup: " "
+          linux-teardown: " "
 
 
       - name: Upload artifact
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Install Dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install "napari/[all]"
+          python -m pip install "napari/[pyside,dev]"
           python -m pip install -r docs/requirements.txt
         env:
           PIP_CONSTRAINT: ${{ github.workspace }}/napari/resources/constraints/constraints_py3.10_docs.txt
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -60,6 +60,10 @@ jobs:
           PIP_CONSTRAINT: ${{ github.workspace }}/napari/resources/constraints/constraints_py3.10_docs.txt
         with:
           run:  make -C docs docs
+          # setting these empty stops from running the default (tiling) window manager
+          # the window manager is not necessary for docs builds at this time
+          linux-pkgs: ""
+          linux-setup: ""
 
 
       - name: Upload artifact
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -53,19 +53,18 @@ jobs:
           python -c 'import napari.layers; print(napari.layers.__doc__)'
 
       - name: Build Docs
-        uses: aganders3/headless-gui@configurable-xvfb
+        uses: aganders3/headless-gui@v2
         env:
           GOOGLE_CALENDAR_ID: ${{ secrets.GOOGLE_CALENDAR_ID }}
           GOOGLE_CALENDAR_API_KEY: ${{ secrets.GOOGLE_CALENDAR_API_KEY }}
           PIP_CONSTRAINT: ${{ github.workspace }}/napari/resources/constraints/constraints_py3.10_docs.txt
         with:
           run:  make -C docs docs
-          # setting these empty stops from running the default (tiling) window manager
-          # the window manager is not necessary for docs builds at this time
+          # skipping setup stops the action from running the default (tiling) window manager
+          # the window manager is not necessary for docs builds at this time and it was causing
+          # problems with screenshots (https://github.com/napari/docs/issues/285)
           linux-setup: "echo 'skip setup'"
           linux-teardown: "echo 'skip teardown'"
-          xvfb-screen-size: "1280x1024x24"
-
 
       - name: Upload artifact
         uses: actions/upload-artifact@v3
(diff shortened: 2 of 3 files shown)


psobolewskiPhD:
Fingers crossed!
Stable docs are from July so maybe something changed?
Good news is that I'm pretty sure if the build artifact is fine then the deployed will also be fine, so we can test easily.

Edit2: Oh and @aganders3 could you also--optimistically--edit the deploy action? 
https://github.com/napari/docs/blob/0c7029e5c9674e9d56c5efbec1022fa302661280/.github/workflows/deploy_docs.yml#L56-L65

aganders3:
> Good news is that I'm pretty sure if the build artifact is fine then the deployed will also be fine, so we can test easily.

Nice - hopefully.

Bad news is I think I need to push a change and run it again because I think it needs a non-empty string to avoid the defaults 🤦.

edit: and we need to keep waiting because as I am learning about JavaScript and GH-Actions APIs in real-time 🤦🤦.

aganders3:
Just pointing out a weird traceback here in a recent build - possibly related but let's see how the docs look before I change anything else (for example trying another WM).
https://github.com/napari/docs/actions/runs/7106959550/job/19347465882#step:8:202

aganders3:
Hm, the first screenshot on that page is still messed up, but the rest look better.

I am looking at this build: https://github.com/napari/docs/actions/runs/7106959550

psobolewskiPhD:
Yeah, the size is fixed, but the first widget is messed up. So I guess this fixes #285 only at the moment.

aganders3:
Okay actually I think that worked!

@lucyleeow @psobolewskiPhD can you check the latest build?
https://github.com/napari/docs/actions/runs/7109505974

I converted to draft because this will still need a new release on the action.

psobolewskiPhD:
yes! Looks correct now in the artifact. I don't see issues in other pages!

.github/workflows/deploy_docs.yml
```diff
@@ -58,7 +58,6 @@ jobs:
         env:
           GOOGLE_CALENDAR_ID: ${{ secrets.GOOGLE_CALENDAR_ID }}
           GOOGLE_CALENDAR_API_KEY: ${{ secrets.GOOGLE_CALENDAR_API_KEY }}
-          PIP_CONSTRAINT: ${{ github.workspace }}/napari-repo/resources/constraints/constraints_py3.10_docs.txt
```
psobolewskiPhD: we need this env var because `make docs` does a pip install.

lucyleeow: :cry: Its so confusing #291

.github/workflows/build_docs.yml
```diff
@@ -53,13 +53,18 @@ jobs:
...
       - name: Build Docs
-        uses: aganders3/headless-gui@v1
+        uses: aganders3/headless-gui@configurable-xvfb
         env:
           GOOGLE_CALENDAR_ID: ${{ secrets.GOOGLE_CALENDAR_ID }}
           GOOGLE_CALENDAR_API_KEY: ${{ secrets.GOOGLE_CALENDAR_API_KEY }}
           PIP_CONSTRAINT: ${{ github.workspace }}/napari/resources/constraints/constraints_py3.10_docs.txt
         with:
           run:  make -C docs docs
+          # setting these empty stops from running the default (tiling) window manager
```
lucyleeow: 'setting these empty' refers to skipping 'setup' and 'teardown' ?

aganders3: Yeah I'll update this before marking it ready. Initially I set these values empty (`""` or `" "`) but it would still fall back to defaults because I think those are both falsely in JavaScript.

.github/workflows/deploy_docs.yml
aganders3: Ah sorry I think I deleted this on error due to rebasing my local `main`. I'll fix before marking ready to review.

psobolewskiPhD:
@czaki I saw the same problems in the artifact from the 0.4.19 docs PR, so I think this will need to be cherry picked over to that branch.

aganders3:
Okay I'm pretty sure this is good-to-go now using `headless-gui@v2`. Sorry everyone!

psobolewskiPhD (approved):
I just checked the latest artifact and everything looks good.

.github/workflows/build_docs.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       - name: Install Dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install "napari/[all]"
+          python -m pip install "napari/[pyside,dev]"
```
psobolewskiPhD: wonder if we should just stick with pyqt and even consider moving circleCI to pyqt.
pyside (and pyside6 especially) has been the more flaky backend...

psobolewskiPhD: Actually, for this PR, I think I'd rather not change the Qt backend, since it's not relevant to the fix and we can make a decision on Qt backends and harmonizing the workflows. Which reminds me, once this is merged, the change needs to also be passed on to:
https://github.com/napari/napari/blob/main/.github/workflows/build_docs.yml
I can take care of that, no problem.

psobolewskiPhD:
```suggestion
          python -m pip install "napari/[all]"
```

aganders3: I agree and am open to any option here. `all` seemed like a bit much but now that I look it's actually just `pyqt5` + plugin manager. I think it's better to try changing CircleCI to use that.

.circleci/config.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           name: Install napari-dev
           command: |
             . venv/bin/activate
-            python -m pip install -e "napari/[pyside,dev]"
+            python -m pip install -e "napari/[all]"
```
psobolewskiPhD:
```suggestion
            python -m pip install -e "napari/[pyside,dev]"
```
This is unrelated to this PR, so let's not change the CircleCI workflow which is currently working fine.

lucyleeow:
Should we get another review? Not sure who would be best...

lucyleeow:
Also I question, I see from here: https://github.com/napari/docs/issues/283#issuecomment-1841604776 it seems that the setup in GH and CircleCI is different.

Would trying to mimic Talleys QT setup action to get consistency between CI workflows be a good idea?

psobolewskiPhD:
> Also I question, I see from here: [#283 (comment)](https://github.com/napari/docs/issues/283#issuecomment-1841604776) it seems that the setup in GH and CircleCI is different.
> 
> Would trying to mimic Talleys QT setup action to get consistency between CI workflows be a good idea?

Not sure if the differences are a) required or b) significant. Either way I think changes to CircleCI workflows are out of scope of this PR, but maybe worth noting in an issue?

Czaki (approved):

lucyleeow:
No problem, it is more from a maintenance point of view - I guess we assume they both do the same thing, and if the docs look fine in CircleCI we assume they would look okay when we deploy but they use *slightly* different workflows? I am no good with Qt and window stuff though. (And I really don't know the ins and outs of our doc CIs)
And yes, further changes are definitely out of scope of this PR.

aganders3:
Thanks for merging @psobolewskiPhD 🚀

Note it did not close the two associated bugs automatically, probably because I edited in the "closing words". 🙁

psobolewskiPhD:
I just wanted to be sure after deployment. Looks good though:
https://napari.org/dev/guides/magicgui.html

PR #125, opened 2023-03-22
Layer documentation enhancements

melissawm:
# Description
This PR contains several improvements to the layers how-to guides content.

Includes new sections on GUI buttons and elements, some updates and fixes to existing text and links to other documentation pages.

Content updates authored by @Nadalyn-CZI 

EDIT: [Start here to see the changes](https://output.circle-artifacts.com/output/job/af326c08-de76-4e03-b0c3-b0719c751fce/artifacts/0/docs/_build/howtos/layers/index.html)

## Type of change

- [x] Fixes or improves existing content
- [ ] Adds new content page(s)
- [ ] Fixes or improves workflow, documentation build or deployment

# References

## Final checklist:
- [x] My PR is the minimum possible work for the desired functionality
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added [alt text](https://webaim.org/techniques/alttext/) to new images included in this PR

PR #298, opened 2023-12-06
Add mp4 and png fallback to webm videos

### LINKED ISSUE #60: Documentation videos invisible or hard to play after move to webm
## 🐛 Bug

napari/napari#4207 converted all of our gifs to webm, in an attempt to stop the bleeding on our LFS bandwidth.  Unfortunately, it seems to have had some bad side effects on usability of the site.

On Chrome 99.0.4844.83 (macOS), it's not very obvious that the movies are anything other than still images.  The mouse doesn't change icon when you hover, clicking does nothing, etc... I had to right click on the image and show controls in order to do anything:

https://user-images.githubusercontent.com/1609449/159525999-344a7fc9-cb35-4d26-b562-40c2fd6091a1.mov

In safari 15.3, I see nothing at all:

Untitled

based on caniuse link that @Carreau added in napari/napari#4207: https://caniuse.com/webm ... it appears that I fall in the "3.4%" of unsupported computers because my desktop isn't on big sur or later yet

on firefox, I also had to right click to get anything to play

Given that LFS has so many other issues, and this change wasn't without some unfortunate side effects, I wonder if perhaps we shouldn't consider reverting this one?

melissawm:

Closes #60 and https://github.com/napari/napari-sphinx-theme/issues/140

# Description

Adds mp4 and png fallback to all webm videos. Also moves them to the appropriate folder for html builds. The downside here is that, in addition to size (because we have more files now), the notebook view of the guides can't seem to be able to play the videos (or use the fallback png). I don't think this is something we use often (open the .md files as notebooks) but ideally I'd like to keep it working.

EDIT: 
~~Some of these videos are outdated. For the moment, I only converted them but will follow up with updates before 0.5~~ Videos have been updated as part of #125 

- Since the fallback videos are now being created on deployment, the only caveat left is the notebook observation above.


psobolewskiPhD:
Huh, why did CircleCI action not run?

docs/guides/layers.md
```diff
@@ -84,7 +84,18 @@ or 3 less than the total number of dimensions of the layer, allowing you to
 browse volumetric timeseries data and other high dimensional data. See for
 example these cells undergoing mitosis in this volumetric timeseries:
 
-![napari viewer with an image of a cell undergoing mitosis. The scroll bar below the canvas controls the timeseries, allowing different stages of mitosis to be visible.](./images/mitosis.webm)
+```{raw} html
```
psobolewskiPhD, line 94: Not sure whats up with this one
I get
image

psobolewskiPhD: Can't get it to play in Safari or on phone Safari.

psobolewskiPhD:
This fixes https://napari.org/stable/tutorials/annotation/annotate_points.html on my phone! (Safari)
🎉

docs/guides/layers.md
```diff
@@ -139,8 +150,18 @@ napari.view_image(retina, name='retina', scale=[1,10,1,1])
 viewer.layers['retina'].scale = [1,10,1,1]
 ```
 
-![napari viewer with an image where all layers are scaled equally; when rotated, the image appears flat. By using console below the canvas and applying a scale factor to one of the dimensions, the image's volume becomes apparent.](images/scaling.webm)
-
+```{raw} html
```
psobolewskiPhD, line 161: Same as above. This one doesn't play for me on macOS safari and phone Safari.

docs/guides/layers.md
```diff
@@ -84,7 +84,18 @@ or 3 less than the total number of dimensions of the layer, allowing you to
 browse volumetric timeseries data and other high dimensional data. See for
 example these cells undergoing mitosis in this volumetric timeseries:
 
-![napari viewer with an image of a cell undergoing mitosis. The scroll bar below the canvas controls the timeseries, allowing different stages of mitosis to be visible.](./images/mitosis.webm)
+```{raw} html
+<figure>
+  <video width="100%" controls autoplay loop muted playsinline>
+    <source src="../../_static/images/mitosis.webm" type="video/webm" />
+    <source src="../../_static/images/mitosis.mp4" type="video/mp4" />
+    <img src="../../_static/images/mitosis.png"
```
psobolewskiPhD:
```suggestion
    <source src="../_static/images/mitosis.webm" type="video/webm" />
    <source src="../_static/images/mitosis.mp4" type="video/mp4" />
    <img src="../_static/images/mitosis.png"
```
I think the path is just wrong. We're in `docs/guides` here so `_static` is just up one and over.

docs/guides/layers.md
```diff
@@ -139,8 +150,18 @@ napari.view_image(retina, name='retina', scale=[1,10,1,1])
...
 ```
 
-![napari viewer with an image where all layers are scaled equally; when rotated, the image appears flat. By using console below the canvas and applying a scale factor to one of the dimensions, the image's volume becomes apparent.](images/scaling.webm)
-
+```{raw} html
+<figure>
+  <video width="100%" controls autoplay loop muted playsinline>
+    <source src="../../_static/images/scaling.webm" type="video/webm" />
+    <source src="../../_static/images/scaling.mp4" type="video/mp4" />
+    <img src="../../_static/images/scaling.png"
```
psobolewskiPhD:
```suggestion
    <source src="../_static/images/scaling.webm" type="video/webm" />
    <source src="../_static/images/scaling.mp4" type="video/mp4" />
    <img src="../_static/images/scaling.png"
```
Same as above.

psobolewskiPhD (changes requested):
I checked everything on my mac and phone. Two videos don't work--I think it's a path thing.
I'm not sure how to check the static image?

docs/guides/layers.md
melissawm: I'm not sure what's happening... All videos work fine for me. Since I used the same converting command for all videos, maybe this is a transient issue?

docs/guides/layers.md
melissawm: I'll try running the converter again.

docs/guides/layers.md
melissawm: Wow that is very weird. How come they were playing for me even with the wrong path? Anyway, I fixed it. Please confirm if it works for you now 😄

Czaki:
How expensive will be built them during build steep instead of storing a few versions in a repository?

melissawm:
The pngs are manually generated for now. The mp4s come from a ffmpeg conversion step so potentially could be automated. This could also only be done on deployment instead of every commit.

Czaki:
> This could also only be done on deployment instead of every commit.

we may also run this when someone add/modify video content

psobolewskiPhD (approved):

psobolewskiPhD:
Gah, this has conflicts with the big layer PR, which also had videos.
I think those videos are all the correct ones, but the trick to get them to load from this PR is still needed?

melissawm:
It's expected, I can fix those, I've already created the alternate versions when updating the layer videos.

melissawm:
I believe I am going to take @Czaki's suggestion here and make this a deployment step. Marking as draft in the meantime

melissawm:
This should be good to go now!

jni (approved):
Nice! @Czaki just want to make sure you're ok with this now before merging. But ❤️!

@melissawm I guess we can make a follow-up issue for notebooks. It's kinda weird that it doesn't work, do you have an ideas about why?

Makefile
```diff
@@ -59,3 +59,9 @@ html-noplot: clean prep-docs
 
 linkcheck-files:
 	NAPARI_APPLICATION_IPY_INTERACTIVE=0 sphinx-build -b linkcheck -D plot_gallery=0 --color docs/ docs/_build ${FILES} -D sphinx_gallery_conf.examples_dirs=$(GALLERY_PATH) $(SPHINXOPTS)
+
+fallback-videos:
+	for video in $(basename $(wildcard docs/_static/images/*.webm)); do \
+		echo $$video; \
+		ffmpeg -i $$video.webm -vf "scale=trunc(iw/2)*2:trunc(ih/2)*2" -c:v libx264 -crf 18 -preset slow -c:a aac -b:a 192k -strict -2 -y $$video.mp4; \
+	done
```
Czaki: We should call `ffmpeg` only when `mp4` file does not exist or is older than the source video. 
Otherwise each time makefile is called the new binary file will be created (because creation date will be different). 

Here sample code that I found:

```Makefile
VIDEOS_DIR := ./videos
WEBM_FILES := $(wildcard $(VIDEOS_DIR)/*.webm)
MP4_FILES := $(WEBM_FILES:.webm=.mp4)

.PHONY: all
all: $(MP4_FILES)

%.mp4: %.webm
	ffmpeg -i $< -c:v libx264 -preset slow -crf 22 -c:a aac -b:a 128k $@

clean:
	rm -f $(MP4_FILES)
```

It will simplify local testing

Czaki (dismissed):
The PR description is confusing. 
It mentions adding mp4 files, but not mp4 files are added. 
The second part suggests that it is intentionally because mp4 files will be generatd during deployment. But there is no rule in `.gitignore` to ignore `*.mp4` files so we are fragile on accidental upload. 

I think that PR description need to be updated. `*.mp4` should be added to `.gitignore`. 

We also need to think how teduce size of commits to `napari/napari.github.io` as each commit will be huge now. But we could do this later.

jni:
> We also need to think how teduce size of commits to `napari/napari.github.io` as each commit will be huge now. But we could do this later.

Well, I think commits reuse blobs when binary files match, so I don't think the size of the snapshot will be multiplied by the number of commits. We also overwrite dev each time, so the only thing we need to do is reuse large files between releases. We could indeed come up with a solution for that. However I think our priority should rather be to move to autogenerated videos...

Czaki:
> Well, I think commits reuse blobs when binary files match

Yes. But part of the binary of video are metadata which contains the creation date. And dates will be different, so binary blobs will be different. 

> We also overwrite dev each time

No. We are adding a new version of binary files each time. This is why this repository is so huge.

melissawm:
Do we need to? Or is rebasing to keep the history lighter an option?

I'm not sure how to solve this issue of the fallbacks otherwise...

Czaki:
> Do we need to? Or is rebasing to keep the history lighter an option?
> 
> I'm not sure how to solve this issue of the fallbacks otherwise...

I am not convinced that we need to solve this in this PR, as it will require significant changes in deploy workflows. I only highlight potential problems. I think that we should start squashing gh-pages branch until we find a proper solution.

Makefile
melissawm: I did it slightly differently for readability, but the effect is the same. Thanks!

melissawm:
Ok -this should be what we needed. Let me know if there's any improvements I can do. Thanks!

psobolewskiPhD:
Awesome! I get what's going on here, but probably we need either some comments or better yet a note in the docs?

psobolewskiPhD:
Also closes https://github.com/napari/napari-sphinx-theme/issues/140

melissawm:
Let me know if this note is enough, cheers!

EDIT: Direct link to preview https://output.circle-artifacts.com/output/job/1cb56e41-b3f1-4ff2-94e8-0958fda80b08/artifacts/0/docs/docs/_build/developers/contributing/documentation/docs_template.html#include-pictures

docs/developers/contributing/documentation/docs_template.md
```diff
@@ -45,6 +45,25 @@ You should include pictures/videos, particularly when describing interactions wi
 accompanied by complete and descriptive alt-text. If you're using arrows/circles to highlight portions of the image, make sure
 they contrast well against the background of the image.
 
+`````{note} Adding videos
```
psobolewskiPhD: It's not letting me suggest, but the title isn't rendering.
image


 I think this is because of how notes are handled.
I think this should be:
```
{admonition} Adding videos
:class: note
```
or maybe better yet:
```
{important} Adding videos
```
Based on https://myst-parser.readthedocs.io/en/latest/syntax/roles-and-directives.html#directives-a-block-level-extension-point

psobolewskiPhD:
Tiny nitpick otherwise good to go!

docs/developers/contributing/documentation/docs_template.md
melissawm: you are right, nice catch!

PR #300, opened 2023-12-06
Update .gitignore

melissawm:
# Description
Simple maintenance PR to update .gitignore file.


.gitignore
```diff
@@ -139,24 +132,25 @@ windows/
...
+docs/jupyter_execute/
+docs/sg_execution_times.rst
+
 # come from npe2 docs
 docs/plugins/_npe2_*.md
-
-napari/settings/napari.schema.json
-
-docs/jupyter_execute/
-docs/.jupyter_cache/
```
lucyleeow, line 161: Was this not needed anymore?

lucyleeow (approved):
Naive review, but LGTM!

.gitignore
melissawm: You are right, the jupyter_execute still needs to be there. I just moved it to the "documentation" section. The .jupyter_cache folder really doesn't get created with our current workflow but I'm happy to add it back as it really doesn't hurt 😄

Czaki (approved):

melissawm:
Thanks for the reviews - merging after discussion with @Czaki. Cheers!

Issue #289, opened 2023-12-04
Explore automatic generation of videos with graphical/mouse interactions


melissawm:
## 🧰 Task

In order to keep the documentation up to date, the videos (webm) we have on our docs pages should be automated whenever possible. This doesn't mean they would be rebuilt at every docs build, but maybe once every release or when we have important gui changes.

The initial proposal is to have a script that is associated with each video and describes the sequence of actions that needs to be executed to capture that output. For example: for [this video on the points layer guide](https://napari.org/dev/howtos/layers/points.html#putting-it-all-together), a script would be:

```
- Click add points icon
- Add three new points
- Click select points icon
- Select two points individually
- Drag mouse to select group of points
- Change face color 
- Change edge color
- Select group of points with different colors
- Use slider to increase point size
- Select another group of points
- Use slider to decrease point size
- With no point selected, click symbol dropdown and select cross
- Use slider to decrease opacity
- Use slider to increase opacity
- Select group of points and click the "delete selected symbol" icon
- Click the add points icon
- Use the face color dropdown to select a different color
- Use the slider to increase point size and add new points
```

In the future, we can maybe explore tools such as https://pyautogui.readthedocs.io/en/latest/

psobolewskiPhD:
Oh wow, that's a great idea!

melissawm:
Another thought: can we add a "review videos and images for outdated content" to the release guide?

The release manager doesn't need to be responsible for this but maybe they can ping the documentation leads. This would make it a soft regular requirement.

kephale:
I just want to cross reference this issue with this open napari issue about video recording from @GenevieveBuckley: https://github.com/napari/napari-animation/issues/225

psobolewskiPhD:
I will add to the cross reference:
docs issue: https://github.com/napari/docs/issues/289
We may want to consider napari-animation in this context too https://github.com/napari/napari-animation

psobolewskiPhD:
Slightly related? but something like the guide tour idea here: https://github.com/napari/napari/issues/2361 could be repurposed to create some viewer states programmatically for docs.

PR #308, opened 2023-12-19
WIP, PoC: Add pyautogui workflow

melissawm:
Addresses #289 

# Description

This is a proof of concept for autogeneration of videos for the docs. There are probably several ways in which this could be improved/changed but I wanted to get some early feedback on the general workflow. Specifically:

- Is this worth doing? We won't have many such scripts right now, but long term they will probably pay off in terms of detecting GUI changes and maintenance.
- This script only performs the actions, but the video needs to be captured by some other tool, as far as I can tell (pyautogui only takes screenshots, not video). I experimented with obs but any other screen recording tool would work. I'd recommend writing docs on the appropriate standard settings so we have consistency over the recorded videos. Note that this workflow also helps with that as the napari window size, theme and settings would all be the same across videos.
- I've had to force the events processing twice for each action because otherwise I would only see the events when the next one was triggered - I did some research and this could maybe be related to some async processing? Unfortunately I don't that much about qt to debug it. Any help is appreciated!


jni:
Ah! This is failing for me, I *think* because my screen is scaled 2x and `locateCenterOnScreen` is not scale-invariant. 😬 After a quick browse of the PyAutoGUI docs I don't see any support for image scaling, so I suspect we might need to use `locate` directly after processing the images with screen scaling. We can potentially even be faster by getting a screenshot of the viewer, grabbing the coordinates within that screenshot, then transforming to screen space.

ANYWAY. This is SO. COOL. It's actually not at all what I was thinking (I thought all the code would be model code and we would just forgo "clicks", just see things happening in the viewer and on screen), but in a lot of ways it's better because of the clicks.

Other ideas:
- grab the button images from the napari repo instead of saved screenshots. That way if we change the icons the scripts still work! 🤯 
- It would be really good if the messages could be integrated with the screen capture method to add captions to the video. Amazing, really. 😃 
- set the image paths using `pathlib.Path('__file__').parent / '../../images/...` so that you can run the script from any directory.

melissawm:
Reviving this after a long time, but after talking to @villares he pointed out this approach: https://www.shedloadofcode.com/blog/record-mouse-and-keyboard-for-automation-scripts-with-python

The idea is that instead of writing every interaction manually in these scripts (which honestly would be a HUGE PAIN) we could record these interactions and create scripts automatically. I haven't tried this yet, but I will!

PR #309, opened 2023-12-19
Telemetry: Add sample code for collecting data and sample outputs

Czaki:
# Description

Based on previous discussion, this PR contains sample code how we could collect data and how a report for telemetry could look.


docs/naps/8-telemetry.md
```diff
@@ -170,6 +170,334 @@ In napari itself, the following changes should be implemented:
 4) code required to init `napari-telemetry` package
 
 
+### Code for collecting data
+
```
DragaDoncila: I would add little descriptive sentence here e.g.:

```suggestion
The following code is a sample Python implementation of the data collecting code that would be part of `napari-telemetry`. It includes functions for assigning a unique ID to the device running `napari`, inspecting the plugin manager and viewer for information and generating the three different levels of report. Sample reports that would be generated by this code are also provided.
```

docs/naps/8-telemetry.md
```diff
@@ -170,6 +170,334 @@ In napari itself, the following changes should be implemented:
...
+        self.save()
+
+
+def get_full_information():
+    middle_info = get_middle_information()
+    data_collector = DataSizeTypeCollector.get_collector(
+        Path(appdirs.user_config_dir("napari")) / "data_size_info.json"
+    )
+    # app_model = get_app() - there is a need to add a collection of data to app_model
```
DragaDoncila, line 356: I'm not sure what you mean here or if it needs to be removed?

DragaDoncila: Ahh I think based on your comment below you mean we need to add a log option to `app-model` so that we can access a history of commands? Maybe just expand a bit on that.

docs/naps/8-telemetry.md
```diff
@@ -170,6 +170,334 @@ In napari itself, the following changes should be implemented:
...
+        "command_usage": { # it should be collected from app_model
+            'napari:window:file:open_files_dialog': 50,
+            'napari:window:file:copy_canvas_screenshot': 10,
+            'napari:window:file:restart': 1,
+            'napari:window:file:close': 100,
+        },
+    }
+```
+
+Code for generate sample data (requires above code):
```
DragaDoncila: Again maybe expand a bit:

```suggestion
The script below generates sample reports for each level of information gathered. The information about layer data is mocked so that this script can be executed independently of `napari`.
```

DragaDoncila (commented):
Just left some suggestions for expanding the text a bit and a few nit comments. Thanks for the detailed sample @Czaki, I think having this info should make it much easier for folks to assess how they feel about the proposed level of telemetry.

jni (approved):

PR #310, opened 2023-12-27
Set `NAPARI_CONFIG` to clean settings

Czaki:
# Description
During work on 0.4.19 I had problem with build docs because of local settings modification. 

This PR set the `NAPARI_CONFIG` variable to an empty value to prevent usage of local settings during build docs.


melissawm (approved):
Thanks @Czaki !

psobolewskiPhD (approved):
This is a broadly useful trick versus the mode brutal `napari --reset`!
We should document it somewhere!

PR #312, opened 2024-01-03
NAP-6: Updates after discussion and review

DragaDoncila:
# Description
This PR updates NAP-6 after feedback on #77, discussion at the napari hackathon (notes [here](https://hackmd.io/fmKp0If5RkiwWIxYYRdKpg#Menu-contributions-notes) but also linked in doc), and discussions in community meetings in recent weeks.

## Still to be decided

- [x] `npe2` allows users to define new submenus as well as menu items. **Should plugin developers be allowed to add new submenus to contributable menus?** Decision: yes, but not to the top level `Layers` menu, and the context menu shouldn't actually be contributable as it's already very long.
- [x] It's not clear that the `Tracks` and `Classification` submenus are as well defined as the rest. Classification at least I'm not even sure how to describe. **Should we remove `Tracks` and `Classification` from the initial list of contributable menus?** Decision: No, we like both!


docs/naps/6-contributable-menus.md
```diff
@@ -318,13 +328,38 @@ Plugins
...
+native napari actions, it's important that users are able to distinguish
+the source of menu items.
+
+To that end, napari items should always be grouped separately to plugin items
+in all menus. Additionally, the plugin's name should also be listed with 
+each plugin contribution. Since plugin names can be quite long, future work should
+consider more concise ways to indicate menu item sources, including using icons.
+
+Outside of a plugin's own submenu, the `order` and `group` keys will be ignored
+for menu contributions. napari will make its own decisions about the grouping
```
DragaDoncila: what is napari strat

docs/naps/6-contributable-menus.md
```diff
@@ -153,22 +153,28 @@ actions.
...
-The five proposed submenus are `Projection`, `Segmentation`, `Classification`, `Registration` and `Tracks`.
+
+The remaining submenus under `Layers` deal with generating new layers based on processing of existing layers
+or other information. They define common *analysis* actions users may want to take on their layers.
+
+4. `Registration` - Commands that allow the user to perform image registration on one or more layers
+5. `Projection` - Commands that generate various projections based on one or more layers
+6. `Segmentation` - Commands that generate image segmentations
+7. `Tracks` - Commands that take in an image or segmentation and produce timelapse tracks
+8. `Classification` - ??
```
lucyleeow: Why the "??"?

docs/naps/6-contributable-menus.md
```diff
@@ -224,44 +237,42 @@ Plugin developers can organize all contributions under this submenu as they see
...
+├─ Tracks           # remove?
+├─ Classification   # remove?
 Acquisition
-Plugins/<my_plugin>    # only individual plugins submenu contributable
-
-Layers Context  # already open for contribution
 ```
 
-As a case study, we take four plugins offering between 9 and 14 widget contributions and arrange their widgets in these menus:
+As a case study, we take four plugins offering between 9 and 14 widget contributions and arrange their widgets in these menus:
```
lucyleeow, line 311: nitpick, is that a trailing white space?
```suggestion
As a case study, we take four plugins offering between 9 and 14 widget contributions and arrange their widgets in these menus: 
```

lucyleeow: I think there are a few added in this PR.

docs/naps/6-contributable-menus.md
```diff
@@ -441,40 +491,45 @@ in the plugin's submenu at the top level.
 
 ## Future Work
 
-As mentioned above, a key feature to support rapid browsing for actions is the
+**Command paltte:** As mentioned above, a key feature to support rapid browsing for actions is the
```
lucyleeow:
```suggestion
**Command palette:** As mentioned above, a key feature to support rapid browsing for actions is the
```

docs/naps/6-contributable-menus.md
```diff
@@ -372,6 +407,21 @@ a Pull Request will be raised opening up this menu for contribution.
...
+the plugin developer's perspective. In general, napari will continue
+to be opinionated about what plugin developers can and cannot change 
+about the napari interface through menu contributions. Our guiding principle
+here is that users should never be surprised by the napari interface as a result
+of standard plugin contributions - menus shouldn't move around, or be hidden, layer
+controls should behave in the same way regardless of what plugins are installed, etc.
+
+However, napari users **should** be able to freely customize their own napari installation
+to a much greater extent. This should include contribution-level enablement control
+for plugins, as well as building their own custom menus, whether this be a brand new top-level
```
lucyleeow: Naive question, how would users control plugin contribution enablement..?

docs/naps/6-contributable-menus.md
```diff
@@ -318,13 +328,38 @@ Plugins
 │  ├─ predict_from_model
 ```
 
+### Item Descriptions
+
+To allow users to browse menus more effectively without having to click each
+item to figure out what it does, menu contributions should be given a `description`
+field that is available when the user is inspecting the menu. The `description`
+should be required, so that
```
lucyleeow: To be finished.
Also just to clarify we would want description to be 'tooltip' field in `Action` which appears in the bottom status bar right?

docs/naps/6-contributable-menus.md
```diff
@@ -318,13 +328,38 @@ Plugins
...
+
+Previously, plugin contributions were limited to the `Plugins` menu and grouped
+under the plugin's name. Now that plugin contributions can be colocated with 
+native napari actions, it's important that users are able to distinguish
+the source of menu items.
+
+To that end, napari items should always be grouped separately to plugin items
+in all menus. Additionally, the plugin's name should also be listed with 
+each plugin contribution. Since plugin names can be quite long, future work should
+consider more concise ways to indicate menu item sources, including using icons.
```
lucyleeow, line 404: If we do have an icon to indicate plugin contribution, we could add the plugin name to 'description' (i.e. tooltip field, so it appears on hover)

lucyleeow (commented):
Flyby review. Will have a better look later!

docs/naps/6-contributable-menus.md
```diff
@@ -274,27 +285,26 @@ Layers
...
+│  ├─ SDetectorDog (napari-stracking)
+│  ├─ SDetectorDoh (napari-stracking)
+│  ├─ SDetectorLog (napari-stracking)
+│  ├─ SDetectorSeg (napari-stracking)
+├─ Tracks
+│  ├─ SLinkerShortestPath (napari-stracking)
+│  ├─ Tracks (napari-mm3)
+├─ Classification
+├─ make_point_cloud_sampling (napari-clemreg)
+├─ make_point_cloud_registration (napari-clemreg)
```
jni: why is this not in Registration?

docs/naps/6-contributable-menus.md
jni: I like this idea. Potential icon: 🔌

docs/naps/6-contributable-menus.md
jni: I imagine, a tree selection widget like this one:

flutter tree view

I think there should be some internal npe2 data structure for this, but the UI itself should itself be a plugin. (Maybe built-ins.)

lucyleeow: Would this be separate to our plugin manager (i.e. the install/uninstall/enable/disable dialog)?

(plugin manager has several meanings in napari: https://github.com/napari/docs/issues/321#issuecomment-2011580837 :see_no_evil: )

jni: I'm ambivalent about it, and would leave that decision to whoever ends up implementing it.

On the one hand, it can be an independent dialog — enablement of contributions only needs npe2 and stuff already installed, and doesn't need to depend on conda, pip, etc. So you eliminate a lot of complexity. Additionally, I personally would never use the plugin manager (I directly install things with micromamba as needed), but I would probably use the enablement UI because the alternative, I think, is editing some random yaml or something? 😂 Which I'm not super excited about.

On the other hand, I can imagine a user being in the enablement UI and thinking hmm, I really don't use this plugin, I should remove it. And they would probably be grateful if that can be done in the same widget rather than have to launch a totally different interface.

jni:
Ok so I finally read NAP-6. Better late than never???? 😅😅😅😅😅 Thank you @DragaDoncila for your patience with me. 😅 

And thanks for writing the original NAP and for this update! Here are my current comments:

- I think eventually we will have an Edit top level menu that would include copy, paste, delete, undo, redo, select all, deselect all. Is that confusing with Layer > Edit? (ie should we rename the layer > edit submenu? or should we remove it and put those items in Edit? or put certain items in both?)
- I found this bit super icky:

  ```
  menus:
    napari/layers/context:
      - submenu: context_submenu
      - command: napari-demo.menu_item
    context_submenu:
      - command: napari-demo.submenu_item

  submenus:
    - id: context_submenu
      label: A new submenu
  ```

  Do we really need a submenus section? I suggest that a menu is a menu is a menu, and it can be nested in another menu, and the only thing we don't allow is not nesting ancestors (ie no cycles). 

- For each menu or submenu proposed, there should be at least one concrete example. I had to think quite a bit about what would go in the "visualization" menu. Auto-contrast-limits? Toggle visibility?
- This is the grammar nazi in me but we need to be consistent with word type/conjugation. It was jarring to have "Segmentation" followed by "Tracks". I have a mild preference for present imperative verbs: Acquire; Filter; Segment; Track; Measure. Not least because they are significantly shorter than "-ation" words.
- File > I/O and Acquire (using new nomenclature 😂) are quite related. I wonder whether we should have an I/O top level menu under which Acquire, Open Dialog, and Export would be menus/submenus. But I see the tension with the File menu...
- In a similar vein, things like the midi controllers, stream decks, space mouses, etc, would all live in something like "Controllers", "Devices", or "Hardware", the last of which could naturally group with "Acquire". Just a thought, since moving/removing is hard compared to adding, so an Acquisition top-level menu is a Big Deal.
- (I super love that the NAP documents the process for updating the list of contributable menus though ❤️‍🔥)
- in the "user configuration" section, there's a bunch of proposals that I think don't need to happen immediately — I suggest making it clear that this is for future work.
- there's a typo "suers" instead of "users". Hopefully not an omen. 😜 
- Finally, the discussion section needs to be a bit more thorough in providing a synthesis of past discussion, including objections to the current proposal and the outcome to those objections. Even if the outcome is "brushed aside as too onerous" 😂 But reading "there is much discussion" is jarring. 😂 

That's it from me! It's not a tiny list but I think this is very close. ❤️ Thanks!

lucyleeow:
> I found this bit super icky:

I think this ties in with the wider discussion about what the manifest should look like: https://github.com/napari/napari/issues/6227

With submenus in particular, you either have to give the full path (i.e. parent menu) every time you use it, or you have a submenu section that defines submenu and its parent, and then use the shorter name every time you use the submenu. 

With an updated manifest, having a submenu section may not look as bad?

docs/naps/6-contributable-menus.md
DragaDoncila: Expanded.

docs/naps/6-contributable-menus.md
DragaDoncila: Done.

docs/naps/6-contributable-menus.md
DragaDoncila: Added a sentence.

docs/naps/6-contributable-menus.md
DragaDoncila: Iunno... it is now 🙈

DragaDoncila:
Thanks @lucyleeow and @jni I've made some updates would appreciate another pass. Also @jni some responses inline.

> * I think eventually we will have an Edit top level menu that would include copy, paste, delete, undo, redo, select all, deselect all. Is that confusing with Layer > Edit? (ie should we rename the layer > edit submenu? or should we remove it and put those items in Edit? or put certain items in both?)

Yep you're not wrong, @psobolewskiPhD has brought this up too. I don't know what exactly the distinction should be there. I would look to other tools - I know Photoshop also has `Edit` and `Layer` so that's probably a good place to start...? @isabela-pf you might have insights here specifically. 

>   Do we really need a submenus section? I suggest that a menu is a menu is a menu, and it can be nested in another menu, and the only thing we don't allow is not nesting ancestors (ie no cycles).

Yeah I don't love it.. in a world where `command`s have associated `menu` items, then the `submenus` thing would just have the label and there'd be no `menus` section, so maybe wouldn't be as bad, or you could nest that too. I have kept the NAP so far in its current state, but if you think it's worth pushing a manifest change before 0.5.0, we can update the NAP. Otherwise we can just follow up to the ugliness with an auto-convert for the updated manifest?
  
> * For each menu or submenu proposed, there should be at least one concrete example. I had to think quite a bit about what would go in the "visualization" menu. Auto-contrast-limits? Toggle visibility?

I've added a couple more examples. You're welcome to suggest your own 😛 Otherwise definitely let me know which ones you can't think of an example for. W.r.t the visualization one specifically, Lorenzo and I did have [this discussion](https://github.com/napari/docs/pull/77#discussion_r1052142797).

> * This is the grammar nazi in me but we need to be consistent with word type/conjugation. It was jarring to have "Segmentation" followed by "Tracks". I have a mild preference for present imperative verbs: Acquire; Filter; Segment; Track; Measure. Not least because they are significantly shorter than "-ation" words.

Fixed, I like the shorter imperatives as well. I think I had it like that at one stage but then changed it... 

> * File > I/O and Acquire (using new nomenclature 😂) are quite related. I wonder whether we should have an I/O top level menu under which Acquire, Open Dialog, and Export would be menus/submenus. But I see the tension with the File menu...
> * In a similar vein, things like the midi controllers, stream decks, space mouses, etc, would all live in something like "Controllers", "Devices", or "Hardware", the last of which could naturally group with "Acquire". Just a thought, since moving/removing is hard compared to adding, so an Acquisition top-level menu is a Big Deal.

I think I/O at top level is pretty strange compared to` File -> I/O...`. But yes I agree top-level is a big deal. We could do `File -> Acquire` which is honestly not bad...?

> * in the "user configuration" section, there's a bunch of proposals that I think don't need to happen immediately — I suggest making it clear that this is for future work.

Added a sentence!

> * there's a typo "suers" instead of "users". Hopefully not an omen. 😜

🙈 

> * Finally, the discussion section needs to be a bit more thorough in providing a synthesis of past discussion, including objections to the current proposal and the outcome to those objections. Even if the outcome is "brushed aside as too onerous" 😂 But reading "there is much discussion" is jarring. 😂

Boo 😜 Still need to do this and update the TODOs.

jni:
> I know Photoshop also has Edit and Layer so that's probably a good place to start...?

But Photoshop doesn't seem to have a Layer > Edit menu. ([screenshot](https://www.dummies.com/wp-content/uploads/229494.image0.jpg) from the web.)

Reading the example text, "The items in this submenu allow you to change the data of your layer", I would consider renaming this to Layer > Modify, or Layer > In-Place > ...

I would also say "to change the data *and metadata* of your layer".

Finally, I'm quite apprehensive about supporting/encouraging in-place filters at all. I might suggest that we should remove filters from the example and focus instead on metadata...

> if you think it's worth pushing a manifest change before 0.5.0, we can update the NAP. Otherwise we can just follow up to the ugliness with an auto-convert for the updated manifest?

My gut feeling is that it is worth updating the schema. I'm not 100% sure about Lucy's idea that this would imply always typing in the full menu path. I think it's possible to say that *if* the tail of the ID is unique, you can just use that ID; if there is ambiguity, you must use the full path. So, you'd have to write "layer/edit", but "segment" would automatically point to with "layer/segment". And also the parent path would be implied in the `submenu:` entry (maybe rename to "contains:"?), so you would have:

```
menus:
  layers:
    - contains: edit
  layers/edit:
    - command: copy_transformations
```

Honestly, I can easily imagine that this approach could devolve into a mess, but my gut feeling is that it's worth trying it out.

> You're welcome to suggest your own 😛

> An example would be a widget that allows you to select which subset of points is visible based on feature properties

I know we don't have auto-actions/purely functional inputs, but how far are those? Can we use those examples in the NAP? I would like something like "Another example would be a function that automatically sets the contrast limits based on properties of the data". And toggle visibility surely would be in this menu?

> I think I/O at top level is pretty strange compared to File -> I/O.... But yes I agree top-level is a big deal. We could do File -> Acquire which is honestly not bad...?

I'm trying to say that "Acquire" is a subset of "I/O". So maybe File > Acquire, File > Load?

Anyway, a side benefit of my menus proposal above is that people can just contribute to "acquire" and we can actually move acquire around and everything would Just Work!

-------------

Getting there!

lucyleeow:
> My gut feeling is that it is worth updating the schema.

Are you talking about over hauling the manifest template and addressing https://github.com/napari/napari/issues/6227 ?

>  I'm not 100% sure about Lucy's idea that this would imply always typing in the full menu path

My bad, I was already thinking in terms of a 'new' manifest schema where for each command you list all its 'contribution' types. At least this is format we think we want the manifest data to be represented as, internally in napari, from discussions with @DragaDoncila. Something like:

```
Dict[command-id, dict[ContributionType,  List[ContributionType]
```

lucyleeow:
```
menus:
  layers:
    - contains: edit
  layers/edit:
    - command: copy_transformations
```

I don't mind this but I think 'submenu' is clearer than 'contains'. We also need to include the label info for a new submenu. 

Edit: Thinking about it more, I think it is nice that the 'menu contributions' lives with the new submenus definitions.
Edit edit: I realise that if we update the manifest schema (to address problems raised in https://github.com/napari/napari/issues/6227) we may not have individual contribution sections (i.e., a `menu` section) so the example above would not work.

docs/naps/6-contributable-menus.md
```diff
@@ -153,34 +153,43 @@ actions.
...
+The remaining submenus under `Layers` deal with generating new layers based on processing of existing layers
+or other information. They define common *analysis* actions users may want to take on their layers.
+
+4. `Register` - Commands that allow the user to perform image registration on one or more layers
+5. `Project` - Commands that generate various projections based on one or more layers
+6. `Segment` - Commands that generate instance segmentations e.g. watershed, SLIC, threshold
+7. `Track` - Commands that take in an image or segmentation and produce timelapse tracks
+8. `Classify` - Commands that predict classes/categories for pixels, objects, points, etc.
 
 Many of the actions in this menu exist in the right click layer context menu. These items
```
lucyleeow, line 193: Question - would all items in the layer context menu be replicated in the `Layers` menubar menu?

docs/naps/6-contributable-menus.md
```diff
@@ -153,34 +153,43 @@ actions.
...
 should be replicated in the `Layers` menu as needed, both to aid discoverability and
 to ensure users are not met with empty menus on initially opening `napari`. It's
 also possible that exceedingly common operations e.g. thresholds, are provided in the future
 by `napari` itself.
 
-### The `Acquisition` Menu
+### The `Acquire` Menu
 
-In addition to the `Layers` menu, we add `Acquisition` as a top level menu.
+In addition to the `Layers` menu, we add `Acquire` as a top level menu.
```
lucyleeow:
```suggestion
In addition to the `Layers` menu, we would add `Acquire` as a top level menu.
```

docs/naps/6-contributable-menus.md
```diff
@@ -153,34 +153,43 @@ actions.
...
+the existing layer data.
+
+The remaining submenus under `Layers` deal with generating new layers based on processing of existing layers
+or other information. They define common *analysis* actions users may want to take on their layers.
+
+4. `Register` - Commands that allow the user to perform image registration on one or more layers
+5. `Project` - Commands that generate various projections based on one or more layers
+6. `Segment` - Commands that generate instance segmentations e.g. watershed, SLIC, threshold
+7. `Track` - Commands that take in an image or segmentation and produce timelapse tracks
+8. `Classify` - Commands that predict classes/categories for pixels, objects, points, etc.
```
lucyleeow: Question, would these be submenus in `Layers` or would they one layer deeper e.g., `Layers/Processing/...` ? (just thinking about how long the `Layers` menu would be)

docs/naps/6-contributable-menus.md
```diff
@@ -205,6 +214,13 @@ It is likely, therefore, that some plugins will always have bespoke interfaces f
 exporting various file formats. These interfaces will be exposed via the new `File->I/O Utilities`
 menu.
 
+### The `File -> New Layer` Menu
+
+This menu will contain built-in `napari` commands for creating new empty layers (similar to 
+the new layer buttons atop the layerlist), but also allow plugins to declare commands that
```
lucyleeow:
```suggestion
the 'new layer' buttons atop the layerlist), but also allow plugins to declare commands that 
```

docs/naps/6-contributable-menus.md
```diff
@@ -217,51 +233,84 @@ It may never make sense for such plugins to distribute their widgets across the
...
 
 We therefore give plugin developers full control over their own submenu under `Plugins->My Plugin`.
-Plugin developers can organize all contributions under this submenu as they see fit, including adding their own submenus of arbitrary depth.
+Plugin developers can organize all contributions under this submenu as they see fit, including adding 
+their own submenus of arbitrary depth.
+
+All commands offered by a plugin will always be available in the plugin's submenu, regardless of
+their presence in other menus. If the plugin defines its own submenus, these will always be first,
+followed by a separator, then any commands with no dedicated menu entry for the plugin's own 
+submenu. These will be alphabetically ordered.
```
lucyleeow, line 255: I think I forgot this detail. Are we sure we want to duplicate menu items (i.e., place contribution in the plugin's submenu *and* in another menu, defined by the plugin?)?

docs/naps/6-contributable-menus.md
```diff
@@ -441,40 +531,45 @@ in the plugin's submenu at the top level.
...
 the syntax for declaring these contexts is strictly defined, we should be
-able to surface information to suers about what is required for the action
+able to surface information to users about what is required for the action
 to be enabled and functional. For example, an action could declare itself
-enabled only when a points layer *and* an image layer are selected. If
+enabled only when a points layer *and* an image layer are selected. If 
 the user has only selected an image layer, we could indicate the missing
 context to the user e.g. "Action takes a points and image layer, but no
-points layer is selected".
+points layer is selected".
```
lucyleeow:
```suggestion
points layer is selected".
```

nit - trailing whitespace. Noted below too..

lucyleeow (approved):
LGTM! Just some questions and nits. Feel free to ignore!

docs/naps/6-contributable-menus.md
jni: No, we are not sure, but personally I think it is better to have a way to see all of a plugin's contributions in one place. But we are definitely not sure, this is 100% open to discussion.

docs/naps/6-contributable-menus.md
DragaDoncila: ok I think I've removed them all!

docs/naps/6-contributable-menus.md
DragaDoncila: These are in `Layers` itself, after a separator. I did initially have them under `Layers/Generate` (with the idea being that they "generate" new layers from existing layers, but folks preferred to just move everything out to the top level)

docs/naps/6-contributable-menus.md
DragaDoncila: Great question! I imagine they will all somehow be replicated either somewhere in the `Layers` menu or somewhere in the `Edit` menu? But again not super clear to me just now what the distinction between those should be, so I want to have a closer look at e.g. Photoshop

DragaDoncila:
> But Photoshop doesn't seem to have a Layer > Edit menu. ([screenshot](https://www.dummies.com/wp-content/uploads/229494.image0.jpg) from the web.)

Sure yeah, I guess we need to take a closer look at what exactly the separation is that they've made, and see where it makes sense for us, and maybe add the `Edit` menu at the top level now with all the stuff we think should be in it.

> Finally, I'm quite apprehensive about supporting/encouraging in-place filters at all. I might suggest that we should remove filters from the example and focus instead on metadata...

I think people will absolutely want to be editing their data **and** their metadata potentially so idk, unless you want to make some blanket rule like "every action returns a new layer" I'd rather have a dedicated spot for that stuff to go...

> My gut feeling is that it is worth updating the schema. 

I'm open to that with the caveat that it **will** delay things by potentially some weeks. I've uploaded a tentative new manifest [here](https://gist.github.com/DragaDoncila/1d8e3b7419e66b90f3215fcdfdae5808) for us to have a look at and make sure we're talking about the same thing. In this one, as @lucyleeow has mentioned, the menus are attached directly to their commands, but we keep a `submenu` section where we declare the labels for the submenus. We still determine the root of the submenu from its prefix. What do you think about this?
 
> I know we don't have auto-actions/purely functional inputs, but how far are those? 

You know nothing, Jon Snow. Functional inputs are totally possible, and if they return layer type things those get added to the viewer, and if they request layer type things those get provided afaik. The `File -> New Layer` menu should contain a few pure commands I imagine.

> Can we use those examples in the NAP? I would like something like "Another example would be a function that automatically sets the contrast limits based on properties of the data". 

I've added that example, and I've expanded the section about 'what do menu contributions do' to talk about them.

> And toggle visibility surely would be in this menu?

Unless it goes in top level edit?

> I'm trying to say that "Acquire" is a subset of "I/O". So maybe File > Acquire, File > Load?

Let's go with `File > Acquire`. Updated.

> Anyway, a side benefit of my menus proposal above is that people can just contribute to "acquire" and we can actually move acquire around and everything would Just Work!

This I **love**, and I've kept it in the proposed manifest above.

DragaDoncila:
> Anyway, a side benefit of my menus proposal above is that people can just contribute to "acquire" and we can actually move acquire around and everything would Just Work!

Just realized it's going to be messy having the keys be just the menu name with absolutely no prefix, because of potential clashes with other plugin submenus etc. So I think it should be `napari/acquire` (for example), but we just don't fill in the full path.

jni:
> Just realized it's going to be messy having the keys be just the menu name with absolutely no prefix, because of potential clashes with other plugin submenus etc. So I think it should be napari/acquire (for example), but we just don't fill in the full path.

We control the contributable menus, so we can make sure there are no name collisions there. We can also decide on a resolution order *or* raising if a name clashes and is not fully qualified. So, if napari-special has an edit menu, they should refer to it as napari-special/edit, *or* it will end up in our own menu.

DragaDoncila:
> We control the contributable menus, so we can make sure there are no name collisions there. 

Yeah, but I think explicit is better than implicit, and it's nice to have the menus either start with `napari`, or the plugin's own name

jni:
> Yeah, but I think explicit is better than implicit, and it's nice to have the menus either start with napari, or the plugin's own name

no, it is a nightmare to type "napari" or the plugin name umpteen times. That is one of my manifest gripes, you'll recall, and moving it from command-id to the menus is really poor consolation...

If we can come up with a shortcut for the plugin's own menu, that would be fine — so it's either nothing (napari contributable), or `self` — the plugin's menu. (or `own`?)

> > And toggle visibility surely would be in this menu?
> 
> Unless it goes in top level edit?

I would find that super weird. Maybe a View top level menu?

Also, we could/should rename Layer > Edit to Layer > Modify.

BUT ALSO,

> I think people will absolutely want to be editing their data and their metadata potentially so idk, unless you want to make some blanket rule like "every action returns a new layer" I'd rather have a dedicated spot for that stuff to go...

We don't have such a blanket rule *and* some widgets might want to have a toggle for in-place processing. So I actually don't think that they should all go in one place. For example, scikit-image morphology remove_small_objects could work both in place or return a copy, and users should be able to decide. I don't think these should be two separate commands.

jni:
> You know nothing, Jon Snow. Functional inputs are totally possible,

but are they possible from plugins???

DragaDoncila:
> but are they possible from plugins???

yes!!! 😜

DragaDoncila:
> We don't have such a blanket rule and some widgets might want to have a toggle for in-place processing.

Yeah I think for widgets it's a bit different because I feel the user will probably be making choices and clicking buttons of their own volition. With **commands** I'm worried, because if someone clicks on something and expects a new layer but it actually edits the layer, that's a problem - especially since we don't have undo. 

So for me that was kinda the guideline - put stuff in this edit menu and people should expect that it works on the layer data. But that's where I think we need to have more specific contribution types that guarantee things to the user about what's going to happen once they click this button. 

Happy to think of other ways to organize them, with the key goal being I don't want users to experience trauma at the hands of an overzealous command.

lucyleeow:
> I've uploaded a tentative new manifest [here](https://gist.github.com/DragaDoncila/1d8e3b7419e66b90f3215fcdfdae5808) for us to have a look at and make sure we're talking about the same thing.

That's exactly what I was thinking. Since there are other sections (sample URI, theme) as well as submenu, I don't mind there being a submenu section.

### Edit - in-place vs new layer

>  With commands I'm worried, because if someone clicks on something and expects a new layer but it actually edits the layer, that's a problem - especially since we don't have undo.

and

> For example, scikit-image morphology remove_small_objects could work both in place or return a copy, and users should be able to decide. I don't think these should be two separate commands.

@jni when you say these shouldn't be two separate commands, do you mean that this would be one (contributable) menu item but there would be a pop-up asking what the user wants to do (in-place to return new layer?). Or do you think there should be 2 menu items? How do you see this working?

### Menu paths - full vs partial

First, I have forgotten, but have we decided if a plugin should be able to contribute to another plugins menu?

> no, it is a nightmare to type "napari" or the plugin name umpteen times. That is one of my manifest gripes, you'll recall, and moving it from command-id to the menus is really poor consolation...

> If we can come up with a shortcut for the plugin's own menu, that would be fine — so it's either nothing (napari contributable), or self — the plugin's menu. (or own?)

@jni is it really that bad, and it would be for a few submenu items?

I think a plugin should be able to specify which menu an item goes into, which I don't think will be fixed by having a conflict resolution order. How do you propose a plugin be able to specify whether an item goes into napari menu or it's own menu, or even another plugins menu?

jni:
> @jni when you say these shouldn't be two separate commands, do you mean that this would be one (contributable) menu item but there would be a pop-up asking what the user wants to do (in-place to return new layer?). Or do you think there should be 2 menu items? How do you see this working?

Well, for this specific example there is an extra parameter (the minimum object size) needed, so you will need a widget one way or another — and that widget would have a boolean "in-place" that would be off by default. For cases where you *only* need a boolean you could instead present two buttons "run in-place" "run and create new layer".

> I think a plugin should be able to specify which menu an item goes into, which I don't think will be fixed by having a conflict resolution order. How do you propose a plugin be able to specify whether an item goes into napari menu or it's own menu, or even another plugins menu?

- foo -> goes in napari's foo menu
- self/foo -> goes in current plugin's foo menu
- other/foo -> goes in other plugin's foo menu

ie, we make napari a privileged set of menus, everything else must be specified, but with a shortcut for self.

lucyleeow:
> For cases where you only need a boolean you could instead present two buttons "run in-place" "run and create new layer".

Two buttons - are you talking about when the action is a widget? Is it possible that the action is just a menu item? In that case should there be two menu items?

jni:
I'm envisioning that in the future, if, say, a command requires an integer, selecting that command from the menu will bring up a short-lived pop-up to select the value, then you click "run" and the dialog disappears and the command runs, rather than a full-on docked widget, which is more suited to long-lasting interactions, rather than single runs of a thing. (The converse could also happen: don't use a dialog, dock the widget, but by default hitting "run" disappears the widget.)

In this world, having a single tick-box and a run button is not as good as having two buttons, "run ticked" and "run unticked".

lucyleeow:
> The converse could also happen: don't use a dialog, dock the widget, but by default hitting "run" disappears the widget.)

I think I prefer the dialog, which is what I was thinking above.

> In this world, having a single tick-box and a run button is not as good as having two buttons, "run ticked" and "run unticked".

I this world - in the world where the menu item opens a docked widget?

I may have missed it, but what did we decide about commands that can become a widget or be a menu item?

jni:
> In this world - in the world where the menu item opens a docked widget?

in the the world where a menu item will open a (probably magicgui-generated) dialog if we can't automatically determine the value of all the arguments to the function from the current context.

> I may have missed it, but what did we decide about commands that can become a widget or be a menu item?

I don't know that it's been decided, but the macOS convention is that you have an action normally and if you hold down Option the menu item changes from `Title` to `Other Title…` and you get a widget. (In general, any menu item that will require a dialog should have a "…". See for example Firefox > Bookmarks > Bookmark All Tabs…) See e.g. Pages:

https://github.com/napari/docs/assets/492549/c626c3e7-53d9-4d62-9913-9077013c4810

I think currently most plugins are widgets one way or another, so we can just leave everything as a widget and work with the community to have *most* actions be contextual, and holding option for specifying parameters. Dock Widgets would be their own thing.

lucyleeow:
> in the the world where a menu item will open a (probably magicgui-generated) dialog

Oh the buttons would be in the pop up dialog? That makes more sense

jni (approved):

jni:
Since this PR does not affect the NAP status, and we've reached something close to a stable state, I'm going to merge so the document can be in published form for (hopefully final) discussion and acceptance. 😊 *Massive* thanks to Draga for all the work so far and @lucyleeow for the discussion — which has helped make the document *and* our own brains clearer! 😊

PR #313, opened 2024-01-07
Use tabs for the install as Python package guide

psobolewskiPhD:
This came up at the docs workgroup and also:
https://github.com/napari/docs/pull/274#pullrequestreview-1740859412

# Description
Currently, the install docs have pip, conda and GitHub install instructions in series followed by the `checking that it worked`. No-one will use all three approaches.
In this PR instead Tabs are used for *each* of the approaches, to let the user choose an install strategy. 
This makes the install page simpler and cleaner, leading the user to the the next logical step.

What it looks like:
image

### DIFF
--- a/docs/tutorials/fundamentals/installation.md
+++ b/docs/tutorials/fundamentals/installation.md
@@ -78,7 +78,6 @@ If you want to install napari with PySide2 as the backend you need to install it
 ```sh
 conda install -c conda-forge napari pyside2
 ```
-````
 
 ````{note}
 In some cases, `conda`'s default solver can struggle to find out which packages need to be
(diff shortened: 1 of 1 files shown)


psobolewskiPhD:
Here's the new section in the CircleCI docs preview:
https://output.circle-artifacts.com/output/job/5bd089ca-8cda-48fa-a4af-d9cf370c1e0a/artifacts/0/docs/docs/_build/tutorials/fundamentals/installation.html#install-as-python-package-recommended

jni (approved):
Love this! 😍

btw we can/should probably add "conda nightlies" as an option. Might get me using them more. 😂 But that can come in a subsequent PR.

psobolewskiPhD:
I was thinking about that. 
Not even sure i like the `from main` on the same level as the regular install options, but if we are to keep it there, then adding conda nightlies to that tab makes sense.

melissawm (approved):
Love it!

jni:
I've raised #315 to discuss the nightlies issue in a dedicated space. 😊

PR #314, opened 2024-01-08
Fix execution of magicgui guide

melissawm:
# Description
Fixes a typo that generates an error while building the docs.

### DIFF
--- a/docs/guides/magicgui.md
+++ b/docs/guides/magicgui.md
@@ -409,7 +409,7 @@ function is called:
 ```{code-cell} python
 :tags: [remove-output]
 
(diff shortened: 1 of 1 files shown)


psobolewskiPhD (approved):
Nice catch! Took me a while to figure out how you found that: lots of scrolling and then finally I located the traceback in the Artifacts tab here, for future reference:
https://app.circleci.com/pipelines/github/napari/docs/473/workflows/9c1f028d-4f26-4f66-98aa-2eb3122db758/jobs/463/steps?invite=true#step-106-78540_93

melissawm:
I'm working on reducing the number of errors so we don't miss those anymore!

PR #335, opened 2024-01-23
Move note after screeenshot in install guide

dstansby:
I think it makes sense to put the image next to the text that references it, instead of the note getting between them. Currently:

Screenshot 2024-01-23 at 16 08 31

### DIFF
--- a/docs/tutorials/fundamentals/installation.md
+++ b/docs/tutorials/fundamentals/installation.md
@@ -152,13 +152,14 @@ simply running
 napari
 ```
 
-An empty napari viewer should appear as follows.
+An empty napari viewer should appear as follows:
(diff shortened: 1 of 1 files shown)


melissawm (approved):
That's reasonable, thanks @dstansby !

psobolewskiPhD:
That makes two notes back to back which look gather garish:
image

Should we maybe make the 2nd note just regular text?

dstansby:
Maybe move the (currently) 2nd note up to below the screenshot, as it relates to opening the napari window (that the screenshot shows), and move the 1st note down and make it into normal text?

psobolewskiPhD (approved):

PR #337, opened 2024-01-28
Close calendar event popover when clicking outside it

### LINKED ISSUE #138: Calendar details view improvement suggestions
- [ ] Ideally, it should be dismissable by clicking anywhere outside of the details box
- [ ] It should also be dismissable with Esc
- [ ] It should be centered on the page or near the click, so that one can quickly look at many events in sequence.

CC @melissawm @psobolewskiPhD 

Just implementing the easiest one of the above would be enough for a first pass. Right now, there's a lot of mousing back and forth between the calendar and the X to close the event!

melissawm:
Closes napari/napari-sphinx-theme#138, together with https://github.com/napari/napari-sphinx-theme/pull/147

# Description
Clicking outside of the calendar event modal or pressing the ESC key closes the modal.

### DIFF
--- a/docs/community/meeting_schedule.md
+++ b/docs/community/meeting_schedule.md
@@ -52,10 +52,21 @@ If you are using napari or interested in how napari could be used in your work, 
         modal.style.display = "block";
         var eventTitle = eventObj.title.charAt(0).toUpperCase() + eventObj.title.slice(1);
         document.getElementById("details").innerHTML = '<b>' + eventTitle + '</b>' + '<br>' + eventObj.extendedProps.description;
-        //When the user clicks on <span> (x), close the modal
+        // When the user clicks on <span> (x), close the modal
         span.onclick = function() {
           modal.style.display = "none";
         }
+        // When the user clicks anywhere outside of the modal, close it
+        modal.onclick = function(event) {
+          if (event.target.id == "myModal") {
+            modal.style.display = "none";
+          }
+        }
+        window.addEventListener('keydown', function (event) {
+          if (event.key === 'Escape') {
+            modal.style.display = 'none'
+          }
+        })
       },
       eventDisplay: 'block',
     });


docs/community/meeting_schedule.md
```diff
@@ -52,10 +52,21 @@ If you are using napari or interested in how napari could be used in your work,
...
         var eventTitle = eventObj.title.charAt(0).toUpperCase() + eventObj.title.slice(1);
         document.getElementById("details").innerHTML = '<b>' + eventTitle + '</b>' + '<br>' + eventObj.extendedProps.description;
-        //When the user clicks on <span> (x), close the modal
+        // When the user clicks on <span> (x), close the modal
         span.onclick = function() {
           modal.style.display = "none";
         }
+        // When the user clicks anywhere outside of the modal, close it
+        modal.onclick = function(event) {
+          if (event.target.id == "myModal") {
```
psobolewskiPhD, line 61: Maybe a stupid question (definitely a naive one): Should there be a `!` or a `not` somewhere here to trigger when the click is outside the modal target area?
(I've not tested this because on CI the calendar events don't show)

jni: yeah my intuition is the same as you @psobolewskiPhD, very good catch!

melissawm: Actually no, `MyModal` is the shadow that is overlayed in front of the whole page while the popover itself is the `event`, so this is correct.

jni: Ah, amazing! Maybe change the name to "mySneakyBackgroundHidingModalNotTheModalYouReThinking"??? 😜

jni: @melissawm so to clarify, this will only work once the new theme is released *and* the constraints are upgraded, right?

jni (approved):

docs/community/meeting_schedule.md
Czaki: The logic here is correct. But I prefer to change the name of `myModal` to something more readable. For example,  `eventDetailBackground`

jni: @Czaki that would require a theme change, right? I suggest we merge this and then make the parallel changes after releasing 0.4.19...

Czaki: No. It works with current theme:
obraz

jni: I'm confused: isn't the ID of the background set by the theme? Clearly this is not an arbitrary string, this is a string matching an element of the DOM? Where is the ID of the modal coming from if not the theme?

jni: In the expression `if (event.target.id == "myModal")`

PR #338, opened 2024-01-29
Fix broken links and references

### LINKED ISSUE #331: [dev, latest] Links on Community > Team page go back to stable.
## 🐛 Bug
The first two links here:
https://napari.org/dev/community/team.html
to the Governance model switch the version of the site to Stable.

I think this should be considered a bug, because out-dated information could be shown.

melissawm:
Closes #331

# Description
Fixes a few references that were creating "duplicate target" warnings when building the docs. Also fixes links in team page and a few other pages.

A few references are renamed, because they trigger duplicate reference warnings. For example, if a file is called "benchmarks.md" and its reference targe is `(benchmarks)=`, the link would have to be marked as either the `ref` role or the `doc` role. To make it easier to link, I've changed the reference name to `napari-benchmarks`. There's a couple other instances of the same issue.

This is meant to guarantee the links will still be valid even after reorganizations such as the recent contributing docs reorg.


psobolewskiPhD:
Oh! that makes sense!
Should we also fix `usage` in `usage.md`? It doesn't look like it's used as a link, but it's the same issue if we ever do link to it?

Edit: And am I understanding correctly that `tutorials` works (on that usage.md) because it will point to `tutorials/index.md`?

lucyleeow (approved):
LGTM, thanks!

melissawm:
> Oh! that makes sense! Should we also fix `usage` in `usage.md`? It doesn't look like it's used as a link, but it's the same issue if we ever do link to it?
> 
> Edit: And am I understanding correctly that `tutorials` works (on that usage.md) because it will point to `tutorials/index.md`?

I'd actually wait for that problem to happen, honestly 🤷🏻‍♀️ I'm not sure we'll ever need to link to that page.

And for the tutorials, I'm actually not sure, but your hypothesis does make sense 😄

psobolewskiPhD (approved):

PR #380, opened 2024-03-27
[Maint] fix the download and deploy part of workflow

psobolewskiPhD:
The second part, deployment, of the new unified workflow is failing:
`Error: Unable to download artifact(s): Artifact not found for name: docs`
https://github.com/napari/docs/actions/runs/8457634044/job/23170039408
If you check the timestamps it's running *concurrently* with the `build&upload` job, so the artifact probably isn't there to download.
Additionally, I think the `publish_dir` is wrong, because it's the dir the docs were built into, but now the docs are being downloaded as an artifact, so the artifact should be used.

# Description
In this PR:
- I make the 2nd job depend on the completion of the first using `needs`
- I change the name to `Download & Deploy` to reduce confusion with the previous job.
- I think the extra `path` in for the `actions/download-artifact@v4` step  is spurious, so I delete it.
- fix the `publish_dir` of the deploy action to use the downloaded artifact

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -80,18 +80,18 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: docs
-          path: docs/docs/_build/html
+          path: docs/docs/_build/html/
 
   deploy:
     if: github.event_name == 'push' && (startsWith(github.ref, 'refs/tags/v') || startsWith(github.ref, 'refs/heads/main'))
-    name: Build & Upload Artifact
+    name: Download & Deploy Artifact
+    needs: build-and-upload
     runs-on: ubuntu-latest
     steps:
       - name: Download artifact
         uses: actions/download-artifact@v4
         with:
           name: docs
-          path: docs
 
       - name: get directory name
         # if this is a tag, use the tag name as the directory name else dev
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -107,7 +107,7 @@ jobs:
         with:
           deploy_key: ${{ secrets.ACTIONS_DEPLOY_KEY }}
           external_repository: napari/napari.github.io
-          publish_dir: docs/docs/_build/html
+          publish_dir: ./docs
           publish_branch: gh-pages
           destination_dir: ${{ env.branch_name }}
           cname: napari.org


melissawm (commented):
I think all of that is reasonable, but testing would take a fake deployment in a branch. I can try doing that tomorrow.

melissawm:
Ok - looking at https://github.com/napari/napari-sphinx-theme/pull/160 I figured it out:

OK! That actually explains the failure: https://github.com/napari/docs/actions/runs/8462011401/job/23182687077

We don't use `_build/html` as an output for the docs build process in the Makefile: https://github.com/napari/docs/blob/54ec184b4fdff1267409ffbaa05b3fbaaf8f694c/Makefile#L29

Meaning the artifact we are trying to upload here does not exist: https://github.com/napari/docs/blob/54ec184b4fdff1267409ffbaa05b3fbaaf8f694c/.github/workflows/build_and_deploy.yml#L83

So we need to update the workflow to upload `_build` (removing html), or we need to update the Makefile to build the docs to html. In that case, the CircleCI action will also need to be updated, as well as the contributing documentation guide.

psobolewskiPhD:
> We don't use _build/html as an output for the docs build process in the Makefile:

But I think it should still be the correct path to upload, because the `docs-build` in make uses the -M flag:
https://github.com/napari/docs/blob/54ec184b4fdff1267409ffbaa05b3fbaaf8f694c/Makefile#L29-L30
This was changed in the unify-docs PR:
https://github.com/napari/docs/pull/348
So with -M, from the docs:
> The default output directory locations when using make-mode differ from the defaults when using [-b](https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-b).
> - doctrees are saved to <outputdir>/doctrees
> - output files are saved to <outputdir>/<builder name>

the outputdir is `docs/_build` so the html should end up in `docs/_build/html` (because html is the builder). Then we need `.../html/` I think for the artifact upload.

In that Unify PR I also fixed the CircleCI workflow:
https://github.com/napari/docs/blob/54ec184b4fdff1267409ffbaa05b3fbaaf8f694c/.circleci/config.yml#L46-L51
So that works correctly -- you can see in the action that ran for this PR.
That gives me some confidence that this should work as well.

So the only question is setting the paths correct for the download here.
My understanding is that the download_artifact action will just unzip it to the workspace so the path structure won't be there as it would be when building in the same workflow step. That's why I changed the path in the deploy step.
Testing this is pretty hard though!

Czaki (approved):

lucyleeow (approved):
LGTM from my limited knowledge. It doesn't seem like the trailing slash isn't strictly needed in:

```
path: docs/docs/_build/html/
```
https://github.com/actions/upload-artifact?tab=readme-ov-file#upload-an-entire-directory

but works fine either way. Thanks for fixing this!

PR #382, opened 2024-04-04
Use intersphinx for magicgui objects in `magicgui.md`

lucyleeow:
Some links were broken so I have amended all to use magicgui intersphinx instead. I think the links were there in the original document, not sure why intersphinx wasn't used.

### DIFF
--- a/docs/howtos/extending/magicgui.md
+++ b/docs/howtos/extending/magicgui.md
@@ -110,9 +110,9 @@ associated with those function parameters.
 
 Third party packages (like napari in this case) may provide support for their types
 using
-[`magicgui.register_type`](https://pyapp-kit.github.io/magicgui/usage/types_widgets.html#register-type).
+[`magicgui.register_type`](https://pyapp-kit.github.io/magicgui/api/type_map/#magicgui.type_map.register_type).
 Indeed napari uses
-[`magicgui.register_type`](https://pyapp-kit.github.io/magicgui/usage/types_widgets.html#register-type)
+[`magicgui.register_type`](https://pyapp-kit.github.io/magicgui/api/type_map/#magicgui.type_map.register_type)
 to provide support for napari-specific type annotations. This makes it easy to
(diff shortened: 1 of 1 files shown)


psobolewskiPhD (approved):

docs/howtos/extending/magicgui.md
```diff
@@ -102,17 +101,13 @@ widget_demo.show()
...
 parameters. In the example above, `call_button` specifies the button text and the
 `param_options` `slider_float` and `dropdown` let you customize the widget
 associated with those function parameters.
 
 Third party packages (like napari in this case) may provide support for their types
-using
-[`magicgui.register_type`](https://pyapp-kit.github.io/magicgui/usage/types_widgets.html#register-type).
-Indeed napari uses
-[`magicgui.register_type`](https://pyapp-kit.github.io/magicgui/usage/types_widgets.html#register-type)
+using {func}`magicgui.register_type`. Indeed napari uses {func}`magicgui.register_type`
```
melissawm: These links seem to not be resolving, I think we need
```suggestion
using {func}`magicgui.type_map.register_type`. Indeed napari uses {func}`magicgui.type_map.register_type`
```

melissawm (approved):
Small note, other than that looks good, thanks @lucyleeow !

docs/howtos/extending/magicgui.md
lucyleeow: Thanks, sorry I forgot to check the links. I was trying to use the same object path as I saw in magicgui docs (see [here](https://github.com/pyapp-kit/magicgui/blob/1e78e6f0bee37aeb9ce7d300a52f5f120141f834/docs/type_map.md?plain=1#L273)) but Talley must use different API auto doc thing for magicgui.

lucyleeow: That fixed it, I wonder what Talley uses internally.

PR #386, opened 2024-04-05
Fix napari.org rendering

Czaki:

# Description


brisvag (approved):
Quick-and-dirty merge on request of @Czaki because it's the only way to see if it worked.

PR #389, opened 2024-04-06
Fix doc preview in PR CI

lucyleeow:
* Update names downloading artifacts: we changed workflow name and build dir from 'docs' to 'html' -> this updates this and updates the images to match new names
* Adds more info on difference between local vs PR preview and circle CI vs artifact download preview.


psobolewskiPhD:
Wait, that's the contributing documentation guide, so for this repo. 
In this repo, on PRs, we still have the action that make the artifact. I think it's useful to check it -- it's more of the ground truth than the CircleCI because in principle it's what gets deployed to gh-pages, same runners, same workflow now (since it was unified with all the consequences).
So I'm not sure I like totally deleting the info on downloading the artifact...

(That said on napari/napari it's just circleCI now, which is fine I think because most docs things will go through this repo.)

lucyleeow:
No problem, we can document both. Still need to update the docs as we get to the artifact a different way now.

lucyleeow:
I've updated everything (since we updated the CIs things have changed names) - hopefully its okay, I'll come back and double check later!

lucyleeow:
🤦 oh man I just realised that you could still download the artifact and the old docs described how to do both and were fine really.
Oh well updated names and added more info

docs/developers/contributing/documentation/index.md
```diff
@@ -350,48 +357,60 @@ This will prevent all but the first napari window from being shown during the do
...
    [http://localhost:8000](http://localhost:8000):
 
 ```shell
-$ cd ~/Downloads/docs  # or the path where you extracted the 'docs' artifact
+$ cd ~/Downloads/html  # `cd` to the path where you extracted the 'html' artifact
 $ python3 -m http.server
 ```
 
+(docs_submit_pull_request)=
+
```
psobolewskiPhD:
```suggestion
```
I think this blank line is spurious?

psobolewskiPhD (approved):
Thanks for updating things!

docs/developers/contributing/documentation/index.md
melissawm: I don't think it matters for the output, but it's unnecessary indeed.

lucyleeow: Done!

melissawm (approved):
Looks great to me, thanks @lucyleeow !

jni (approved):

PR #388, opened 2024-04-06
[maint] fix app-model intersphinx link

psobolewskiPhD:
# References and relevant issues
Closes: https://github.com/napari/docs/issues/387

# Description
This PR fixes the link to the app-model API docs in the intersphinx_mapping in conf.py

PR #393, opened 2024-04-06
Fix ALL Sphinx warnings

### LINKED ISSUE #111: Remaining Sphinx warnings
## 📚 Documentation

Currently, we count 310 Sphinx warnings when building the docs. This is a bit cumbersome, and also stops us from seeing new Warnings that are created when something goes wrong in the docs build, because these new Warnings are lost in this big output.

There are currently three main kinds of warnings left:
- [x] ImportErrors when building the docs: Because of how autosummary and autodoc are applied together here, some attributes are not fully qualified when the autosummary template is parsed. [A proposal to fix this is here](https://github.com/melissawm/napari/commit/ca1d0dbbf66f830ca318cb0977f4d3b08f9ac0b3); however, I'm not sure if this is the best solution. I am open to any suggestions!
- [ ] `WARNING: duplicate object description`: again, this is because of how autosummary works, if an attribute is inherited and we expect it to have multiple docstrings (for example, as a parameter on a class docstring, or as a property declared later in the same class or a child class) this generates a duplicate description warning. I don't have an ideal solution for this at the moment.
- [ ] Warnings about docstrings from imported members: there are a few warnings about functools.partial or tqdm, which are created when those docstrings are injected into the napari sphinx docs. We should probably add these functions to a blocklist and not actually document them with autosummary, but rather link to their own documentation. Does this sound reasonable?

If there are any other ideas or opinions on how to fix these autosummary warnings, please share.

### LINKED ISSUE #227: Build should fail if a new page is added but it's not in the TOC
## 🧰 Task

In #144, a new page was added but it wasn't being built in the docs because it wasn't in the toc.yml. This is a very easy step to miss (I have done so many times). Ideally the build would fail if that's the case, or, if that's too hard (because for some reason sometimes we want pages not in the toc?), we could make a bot that warns when that happens? 🤷 Anyway, it would be good to not have to rely on understanding the build system to miss these things...

melissawm:
Closes #111 
Addresses #227 (will need a follow-up turning on `sphinx -W` option)

# Description
This PR includes fixes to broken references and markdown syntax fixes. It also includes a new filter for napari.qt.threading module docstrings that were creating a few syntax warnings. These are not easily ignored by Sphinx and can't be fixed, as they are Markdown being injected into rst docstrings through the napari.qt module.

After this is merged, we will be able to make the build docs CI check to fail on new warnings, and new errors will be more visible.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -21,11 +21,13 @@ from importlib import import_module
 from importlib.metadata import distribution
 from pathlib import Path
 from urllib.parse import urlparse, urlunparse
+import logging
 
 from jinja2.filters import FILTERS
 from sphinx_gallery import scrapers
 from sphinx_gallery.sorting import ExampleTitleSortKey
 from sphinx.highlighting import lexers
+from sphinx.util import logging as sphinx_logging
 from packaging.version import parse as parse_version
 from pygments.lexers import TOMLLexer
 
@@ -226,6 +228,7 @@ exclude_patterns = [
     '.jupyter_cache',
     'jupyter_execute',
     'plugins/_*.md',
+    'plugins/building_a_plugin/_layer_data_guide.md',
     'gallery/index.rst',
 ]
 
@@ -305,6 +308,38 @@ def add_google_calendar_secrets(app, docname, source):
     if docname == 'community/meeting_schedule':
         source[0] = source[0].replace('{API_KEY}', GOOGLE_CALENDAR_API_KEY)
 
+class FilterSphinxWarnings(logging.Filter):
+    """Filter 'duplicate object description' warnings.
+
+    These warnings are a result of autosummary limitations and are not useful -
+    they don't result in any missing documentation or rendering issues, so we
+    can safely ignore them.
+
+    """
+    def __init__(self, app):
+        self.app = app
+        super().__init__()
+
+    def filter(self, record: logging.LogRecord) -> bool:
+        msg = record.getMessage()
+
+        filter_out = (
+            "duplicate object description",
+        )
+
+        if msg.strip().startswith(filter_out):
+            return False
+        return True
+
+def qt_docstrings(app, what, name, obj, options, lines):
+    """Only show first line of Qt threading docstrings.
+
+    Avoids syntax errors since the Qt threading docstrings are written in
+    Markdown, and injected into rst docstring automatically.
+    """
+    if "WorkerBase" in name:
+        if len(lines) > 0:
+            del lines[1:]
 
 def setup(app):
     """Set up docs build.
@@ -313,11 +348,21 @@ def setup(app):
       files found for document when generating the gallery
     * Rewrites github anchors to be comparable
     * Adds google calendar api key to meetings schedule page
+    * Filters out 'duplicate object description' Sphinx warnings
+    * Cleans out Qt threading docstrings
 
     """
     app.registry.source_suffix.pop(".ipynb", None)
     app.connect('source-read', add_google_calendar_secrets)
     app.connect('linkcheck-process-uri', rewrite_github_anchor)
+    app.connect('autodoc-process-docstring', qt_docstrings)
+    logger = logging.getLogger("sphinx")
+
+    warning_handler, *_ = [
+        h for h in logger.handlers
+        if isinstance(h, sphinx_logging.WarningStreamHandler)
+    ]
+    warning_handler.filters.insert(0, FilterSphinxWarnings(app))
 
 
 def get_attributes(item, obj, modulename):
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -311,9 +311,11 @@ def add_google_calendar_secrets(app, docname, source):
 class FilterSphinxWarnings(logging.Filter):
     """Filter 'duplicate object description' warnings.
 
-    These warnings are a result of autosummary limitations and are not useful -
-    they don't result in any missing documentation or rendering issues, so we
-    can safely ignore them.
+    These warnings are a result of autosummary limitations when we have
+    Attributes and Properties in a class sharing the same name.
+
+    The warnings are not useful - they don't result in any missing documentation
+    or rendering issues, so we can safely ignore them.
 
     """
     def __init__(self, app):
--- a/docs/developers/architecture/app_model.md
+++ b/docs/developers/architecture/app_model.md
@@ -37,7 +37,7 @@ Currently, the primary purpose of the `app` is to compose the following
   [commands](app-model-commands) (the actual callable objects) that have been
   registered with the application. Accessible via `app.commands`.
 * {class}`~app_model.registries.MenusRegistry`: maintains all of the
-  [menus and submenus](appp-model-menus) that have been registered with the application.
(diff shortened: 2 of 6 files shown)


docs/conf.py
```diff
@@ -305,6 +308,38 @@ def add_google_calendar_secrets(app, docname, source):
     if docname == 'community/meeting_schedule':
         source[0] = source[0].replace('{API_KEY}', GOOGLE_CALENDAR_API_KEY)
 
+class FilterSphinxWarnings(logging.Filter):
```
lucyleeow, line 311: I thought you could filter warnings like this: https://github.com/sphinx-gallery/sphinx-gallery/pull/521/files

But I can't yet find documentation on this.

melissawm: Usually yes - but this warning in particular you can't (or I couldn't!). See https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-suppress_warnings for a list of warnings that can be explicitly silenced. I think it falls into a similar category as https://github.com/sphinx-doc/sphinx/issues/9750 - if you find a way to filter it, that would be great!

psobolewskiPhD:
Looking at CircleCI output:
``https://app.circleci.com/pipelines/github/melissawm/napari-docs/340/workflows/20e668a6-ad6f-4d21-bca7-31327ec1ed0a/jobs/328?invite=true#step-106-33254_29
``Just 16 warnings! Down from >140!

Looks like a few are fixed by:
https://github.com/napari/docs/pull/388
and then a few more by:
https://github.com/napari/napari/pull/6812

docs/conf.py
lucyleeow: Oh man I didn't know any of this. Using `warnings.filterwarnings` in the `conf.py` file - how does this even work? I can't find documentation of this. Is this a Sphinx thing?

I guess in https://github.com/sphinx-gallery/sphinx-gallery/pull/521/ , the warning is passed on by the extension Sphinx-Gallery? In the Sphinx docs it says:

> Then extensions can also define their own warning types.

Does that mean that extension warning types are also able to be suppressed?

Anyway, I had no idea, this seems reasonable way to fix it AFAICT! Thanks!

lucyleeow (approved):
Thank you!

melissawm:
Look at this https://github.com/napari/docs/actions/runs/8600978497/job/23567119812?pr=393#step:9:1522 😍 Finally!

Czaki:
This duplicated warning is signalig problem with cross references?

psobolewskiPhD:
Interesting this barks:
https://github.com/napari/docs/actions/runs/8600978497/job/23567119812?pr=393#step:9:663
```
WARNING: autodoc: failed to import function '_testsupport.make_napari_viewer' from module 'napari.utils'; the following exception was raised:
No module named 'pytest'
```
The relevant code, as far as I can tell:
https://github.com/napari/docs/blob/1c6155fd6e2f9dd053559bef73bb5f04bda5878c/docs/developers/contributing/testing.md?plain=1#L243-L255
And on napari.org/dev/ it's broken:
https://napari.org/dev/developers/contributing/testing.html#make-napari-viewer
 But with this PR it's working!
https://output.circle-artifacts.com/output/job/54961a00-f23c-4027-bc52-acec2f1e3e84/artifacts/0/docs/docs/_build/html/developers/contributing/testing.html#make-napari-viewer

melissawm:
> This duplicated warning is signalig problem with cross references?

Not really. As far as I can tell, this happens because some Attributes and Properties have the same name, which confuses sphinx (which doesn't deal with Attributes very well). In addition, in some cases when you have subclasses that inherit the parent's docstrings this can also happen (since both objects will be documented with the same parameters/attributes). But I will confess that as much as I've investigated this, it seems like nobody really knows what's going on exactly, including me 😅 (see, for example, https://github.com/sphinx-doc/sphinx/issues/7817 which is the closest I could find) 

We could potentially fix this without ignoring the errors, but that would mean having multiple templates for different classes which would make the docs generation less automatic and more manual. Since these warnings do not impact the rendering of the docs, I think they are safe to ignore.

melissawm:
> Interesting this barks: [napari/docs/actions/runs/8600978497/job/23567119812?pr=393#step:9:663](https://github.com/napari/docs/actions/runs/8600978497/job/23567119812?pr=393#step:9:663)
> 
> ```
> WARNING: autodoc: failed to import function '_testsupport.make_napari_viewer' from module 'napari.utils'; the following exception was raised:
> No module named 'pytest'
> ```
> 
> The relevant code, as far as I can tell:
> 
> https://github.com/napari/docs/blob/1c6155fd6e2f9dd053559bef73bb5f04bda5878c/docs/developers/contributing/testing.md?plain=1#L243-L255
> 
> And on napari.org/dev/ it's broken:
> [napari.org/dev/developers/contributing/testing.html#make-napari-viewer](https://napari.org/dev/developers/contributing/testing.html#make-napari-viewer)
> But with this PR it's working!
> [output.circle-artifacts.com/output/job/54961a00-f23c-4027-bc52-acec2f1e3e84/artifacts/0/docs/docs/_build/html/developers/contributing/testing.html#make-napari-viewer](https://output.circle-artifacts.com/output/job/54961a00-f23c-4027-bc52-acec2f1e3e84/artifacts/0/docs/docs/_build/html/developers/contributing/testing.html#make-napari-viewer)

I think we'd have to include pytest as a dependency for building the docs for this to work without this warning? Not sure why I didn't see this locally...

psobolewskiPhD:
> I think we'd have to include pytest as a dependency for building the docs for this to work without this warning? Not sure why I didn't see this locally...

Local build instructions use dev install of napari, so you probably have it.
Edit: derp, so does circleCI workflow:
https://github.com/napari/docs/blob/1c6155fd6e2f9dd053559bef73bb5f04bda5878c/.circleci/config.yml#L35

🤷

lucyleeow:
> is happens because some Attributes and Properties have the same name, which confuses sphinx (which doesn't deal with Attributes very well). In addition, in some cases when you have subclasses that inherit the parent's docstrings this can also happen

Maybe we could add this info in the `FilterSphinxWarnings` docstring?

melissawm:
@lucyleeow let me know if this is enough or if I should elaborate!

lucyleeow (approved):
LGTM!

psobolewskiPhD (approved):
everything looks good to me -- and it's working!
Thanks!!

psobolewskiPhD:
SO I don't get the lxml warning:
https://github.com/napari/docs/actions/runs/8647550683/job/23709258930?pr=393#step:9:120

I tested locally and I also get it, but yet both lxml and lxml_html_clean are installed.

Why is it referring to `html-clean`? it's `[html_clean]` in the requirements.txt which matches their setup.py
It is a wierd setup.py, so maybe we should just use `lxml_html_clean` directly?

melissawm:
I don't get this warning locally - but it looks like changing that requirement did it!

psobolewskiPhD (approved):
Everything looks good, so this is great, will make real warnings be so much more obvious now.

lucyleeow:
ping @psobolewskiPhD could this go in? :grimacing:

psobolewskiPhD:
Yeah sorry; i'm terrible at circling back to merge things.

PR #403, opened 2024-04-12
Update `app-model` doc with new information about `_mock_app` fixture

DragaDoncila:
Depends on [napari/#6823](https://github.com/napari/napari/pull/6823)

# Description
Updates the information about `_mock_app` to state that we use it explicitly in `_make_napari_viewer` and all
other tests that require a `NapariApplication`.

Also make note that plugins that use `make_napari_viewer` now get the benefit of the `_mock_app`.

### DIFF
--- a/docs/developers/architecture/app_model.md
+++ b/docs/developers/architecture/app_model.md
@@ -485,7 +485,7 @@ singleton `app` may keep a reference to an object, e.g., a
 {class}`~napari._qt.qt_main_window.Window`, that has
 since been cleaned up at the end of a previous test.
 Thus, we mock the `app` in a `_mock_app` fixture, and
-explicitly use it in `make_napari_viewer` as well as in all tests that
+explicitly use it in [`make_napari_viewer`](make_napari_viewer) as well as in all tests that
 use the `get_app` function. This way, a new instance of `app` is returned 
 every time {func}`~napari._app_model.get_app`
 is used inside a test. This 'test' `app` is available for use throughout the test's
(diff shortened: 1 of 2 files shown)


lucyleeow (approved):
LGTM!

docs/developers/architecture/app_model.md
```diff
@@ -482,16 +482,17 @@ Segmentation faults can also occur as a
...
 is used inside a test. This 'test' `app` is available for use throughout the test's
 duration and will get cleaned up at the end.
 
 ```{note}
-Since the `_mock_app` fixture is autouse, a
-:class:`~napari._app_model._app.NapariApplication` is instantiated during setup
-of every test.
+Since the `_mock_app` fixture is used in `make_napari_viewer`, plugins and other
+external users that write tests using `make_napari_viewer` will also have the benefit
+of a new app for each test.
```
lucyleeow, line 497: Should we mention that it is exported just like `make_napari_viewer` so envs with `napari` installed will be able to use these fixtures? We do mention this when we talk about `make_napari_viewer`: https://napari.org/dev/developers/contributing/testing.html#make-napari-viewer

(ignore the failed include, I think thats fixed in #393)

DragaDoncila: I thought about it but since we've kept it private (kinda with the underscore), I'd rather not mention it explicitly as I don't necessarily want to encourage direct usage of it by external packages. If we find use for it I'd be happy to amend then.

lucyleeow: Gotcha, good point.

Slightly related. could we add a link to the `make_napari_viewer`, maybe further up, when we first mention it?

psobolewskiPhD (approved):

lucyleeow (approved):
LGTM! Thank you!

PR #316, opened 2024-01-09
Exclude inherited tqdm members from documentation

melissawm:
# References and relevant issues
Addresses #111 - reduces current sphinx errors from 208 to 135

# Description
This change excludes inherited tqdm methods from being documented in the API docs for `napari.utils.progress`.

Before this change, when building the docs we can see the following error messages:

```sh
/home/melissa/micromamba/envs/napari-dev/lib/python3.11/site-packages/napari/utils/progress.py:docstring of tqdm.std.tqdm.format_meter:36: ERROR: Unexpected indentation.
/home/melissa/micromamba/envs/napari-dev/lib/python3.11/site-packages/napari/utils/progress.py:docstring of tqdm.std.tqdm.format_meter:37: WARNING: Block quote ends without a blank line; unexpected unindent.
/home/melissa/micromamba/envs/napari-dev/lib/python3.11/site-packages/napari/utils/progress.py:docstring of tqdm.std.tqdm.format_meter:42: WARNING: Definition list ends without a blank line; unexpected unindent.
/home/melissa/micromamba/envs/napari-dev/lib/python3.11/site-packages/napari/utils/progress.py:docstring of tqdm.std.tqdm.format_meter:50: WARNING: Inline emphasis start-string without end-string.
/home/melissa/micromamba/envs/napari-dev/lib/python3.11/site-packages/napari/utils/progress.py:docstring of tqdm.std.tqdm.wrapattr:3: ERROR: Unexpected indentation.
<snip>
/home/melissa/micromamba/envs/napari-dev/lib/python3.11/site-packages/napari/utils/progress.py:docstring of tqdm.std.tqdm.format_meter:36: ERROR: Unexpected indentation.
/home/melissa/micromamba/envs/napari-dev/lib/python3.11/site-packages/napari/utils/progress.py:docstring of tqdm.std.tqdm.format_meter:37: WARNING: Block quote ends without a blank line; unexpected unindent.
/home/melissa/micromamba/envs/napari-dev/lib/python3.11/site-packages/napari/utils/progress.py:docstring of tqdm.std.tqdm.format_meter:42: WARNING: Definition list ends without a blank line; unexpected unindent.
/home/melissa/micromamba/envs/napari-dev/lib/python3.11/site-packages/napari/utils/progress.py:docstring of tqdm.std.tqdm.format_meter:50: WARNING: Inline emphasis start-string without end-string.
[...]

PR #404, opened 2024-04-12
Fixes display of inherited members of classes

melissawm:
Follow-up to #316

# Description
I inadvertently added a Jinja syntax error in #316. Also adds a comment to explain why this is necessary.

### DIFF
--- a/docs/_templates/autosummary/class.rst
+++ b/docs/_templates/autosummary/class.rst
@@ -5,8 +5,7 @@
 .. autoclass:: {{ objname }}
    :members:
    :show-inheritance:
-   {# These classes inherit docstrings from the raw qt source, which
-   generates rst syntax errors when building the docs #}
+   {#- These classes inherit docstrings from the raw qt source, which generates rst syntax errors when building the docs #}
    {% if objname not in ["progress", "cancelable_progress"] -%}
    :inherited-members:
    {%- endif %}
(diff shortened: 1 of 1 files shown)


melissawm:
Before: https://napari.org/dev/api/napari.components.Camera.html (for example)

Screenshot_20240412_153727

After:

Screenshot_20240412_160930

psobolewskiPhD (approved):
looks good to me!
thanks!

docs/_templates/autosummary/class.rst
```diff
@@ -5,9 +5,10 @@
 .. autoclass:: {{ objname }}
    :members:
    :show-inheritance:
-   {% if objname != "progress" %}
-      :inherited-members:
-   {% endif %}
+   {#- These classes inherit docstrings from the raw qt source, which generates rst syntax errors when building the docs #}
+   {% if objname not in ["progress", "cancelable_progress"] -%}
+   :inherited-members:
```
lucyleeow, line 10: Question only, does this need to be indented, or is that just style?

lucyleeow: Seems like style only!

melissawm: It can't be indented - that's what I had gotten wrong before. `:inherited-members:`  needs to have the same indentation level as `:show-inheritance:` and the other options, and we can't have any whitespace between these directive options (which is why the jinja template lines start and end with an extra `-` (this prevents extra whitespace from appearing in the final document)

lucyleeow: Ah thank you!

PR #409, opened 2024-04-18
Turn on warnings as error option for sphinx build

melissawm:
--

# Description
Turns on `--fail-on-warning` option for Sphinx, so that CI fails the build when we have new warnings.


DragaDoncila (approved):
Looks good to me @melissawm! Can't believe we've actually got to this point finally 🎉 Tiny nitpick - the description still said "this PR is in draft mode" so I will update that.

DragaDoncila:
@jni and I decided to merge now because then we can merge this into other PRs

PR #410, opened 2024-04-18
Fix broken magicgui and app-model docs links

melissawm:
Towards #409 

# Description
Fixes one broken link in the magicgui how-to, and one warning about an ambiguous reference in the app model docs.


psobolewskiPhD (approved):
Thanks!

PR #430, opened 2024-06-24
Draft release notes for 0.5.0

jni:

PR #434, opened 2024-06-26
Add warnings around draft message

jni:
I thought after merging #430 that the fact that the release hasn't officially happened yet should be more prominent. 😅


Czaki (approved):

Czaki:
You should add #434 to the release notes :P

jni:
They will be added when the script is re-run. For now it is not that important. 😊

PR #438, opened 2024-06-29
Update labels.md for polygon tool stopping behavior

psobolewskiPhD:
Updates docs for changes in:
https://github.com/napari/napari/pull/7019
https://github.com/napari/napari/pull/7023

# Description
In the two PR above, the polygon tool icon was changed from a triangle to a polygon. Additionally, the double-click condition of the Labels version of the tool was switched to match the Shapes one, with the proximity feature being moved to Settings > Experimental.
I also updated the copy for the Shapes version of the tool, because I noticed it doesn't mention double-click.


psobolewskiPhD:
One last thing: in working on this PR i noticed that for Shapes, `Esc` completes, while for Labels `Esc` cancels and `Enter` completes. I dunno if we should fix that? Escape to complete is a poor choice IMO.

jni:
> One last thing: in working on this PR i noticed that for Shapes, Esc completes, while for Labels Esc cancels and Enter completes. I dunno if we should fix that?

They should certainly be consistent.

> Escape to complete is a poor choice IMO.

Although I mostly agree, (a) Enter is awkward to reach if you are using a mouse right-handed, and (b) I just tried out Keynote and Esc completes. So whatever we choose should aim to be consistent with broader software out there. I don't have Photoshop & co to try out. (A lot of early Shapes work was done by @sofroniewn who is very familiar with Illustrator & co, so I *suspect* Esc to complete is actually reasonably standard.)

The other thing is that cancel is not that important an operation *as long as there is undo*, or easy deletion. I think this is true in some form or another for both labels and shapes.

jni:
> as long as there is undo, or easy deletion. I think this is true in some form or another for both labels and shapes.

(But of course we need to improve discoverability of these things. 😅)

jni (approved):

PR #457, opened 2024-07-16
Add docs about the new `napari-base` structure

jaimergp:

# Description

Describes the actions taken at https://github.com/conda-forge/napari-feedstock/pull/63

### DIFF
--- a/docs/developers/coredev/packaging.md
+++ b/docs/developers/coredev/packaging.md
@@ -41,9 +41,10 @@ Check {doc}`release` for more details about the conda-forge release process and 
 We provide three different outputs in the conda package recipe:
 
 - `napari-base`: This is the package that ships the actual source and data. The runtime
-  requirements only include the basic functionality. This package is only recommended as a
-  dependency for plugins and other projects. End users should not need this, since it mostly
-  correspond to the "library layer" of `napari`
+  requirements only include the basic functionality. This package is recommended as a
+  dependency for plugins and other projects. Most end users will prefer the `napari`
+  package, below, but this one may be useful for those wanting a more minimal
+  environment.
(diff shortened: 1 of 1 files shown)


jni (approved):

docs/developers/coredev/packaging.md
```diff
@@ -36,6 +36,21 @@ Due to the staging steps and CDN synchronization delays, the conda packages can
...
 ```
 
+#### conda packaging split
+
+We provide three different outputs in the conda package recipe:
+
+- `napari-base`: This is the package that ships the actual source and data. The runtime
+  requirements only include the basic functionality. This package is only recommended as a
+  dependency for plugins and other projects. End users should not need this, since it mostly
+  correspond to the "library layer" of `napari`
```
jni: I think this was a little too "scary". I've suggested slightly softer language, below.

```suggestion
- `napari-base`: This is the package that ships the actual source and data. The runtime
  requirements only include the basic functionality. This package is recommended as a
  dependency for plugins and other projects. Most end users will prefer the `napari`
  package, below, but this one may be useful for those wanting a more minimal
  environment.
```

jaimergp: Applied! Thanks!

jaimergp:
I can't set the milestone here but I guess it should be 0.5.0? Or 0.5.1?

melissawm:
We can backport to 0.5.0 if needed 👍🏻

DragaDoncila (approved):
This is great, thanks @jaimergp! Would be good as `napari-base` becomes even lighter to update the installation guide to mention it explicitly, but for now I think this is sufficient.

psobolewskiPhD (approved):

PR #460, opened 2024-07-19
Add short guide on updating constraints files

melissawm:
N/A

# Description
Adds a short guide on updating the constraints files for core developers. I wonder if we should automate this with a bot or action (expanding the update test constraints action perhaps?)

### DIFF
--- a/docs/developers/coredev/release.md
+++ b/docs/developers/coredev/release.md
@@ -117,10 +117,11 @@ on the `stdout` if any test fails.
 
 ## Update constraints files
 
-`napari` uses a set of constraints files to make sure all dependencies are compatible with each other.
-These files need to be updated regularly, especially before a release. You can find these files at
+`napari` uses a set of constraints files to prevent test failures due to dependency updates. This also allows for reproducible builds (see [](dev-installation)).
+These constraints files need to be updated at least weekly on Monday, and may also be triggered manually by a maintainer. You can find these files at
 [resources/constraints](https://github.com/napari/napari/tree/main/resources/constraints).
 
+To get updated constraints for a PR, use `@napari-bot update constraints` in a PR comment, then follow the instruction added by the bot to the conversation.
 
 ````{admonition} Example
 To update the docs constraints file, assuming you have the `napari/napari` repo and `napari/docs` repo cloned next to each other, you can install [uv](https://astral.sh/blog/uv) and run the following command from the root of the `napari` repo:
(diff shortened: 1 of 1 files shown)


Czaki:
There is already such an action https://github.com/napari/napari/blob/main/.github/workflows/upgrade_test_constraints.yml

melissawm:
But that is for tests only, right?

Czaki:
> But that is for tests only, right?

No, it generates PRs like this https://github.com/napari/napari/pull/7045
If you see the list of modified files, you will see that docs constraints are also updated.

jni:
Should we also add the bot usage here, and @Czaki does the bot work in the docs repo?

melissawm:
Hi folks, not sure what the next step is here. Maybe we should close since the action runs automatically?

docs/developers/coredev/release.md
```diff
@@ -115,6 +115,23 @@ on the `stdout` if any test fails.
   pytest tools/ --tb=short
   ```
 
+## Update constraints files
+
+`napari` uses a set of constraints files to make sure all dependencies are compatible with each other.
```
Czaki:
```suggestion
`napari` uses a set of constraints files to prevent random test crash because of some depndecy update. It allow also for future reprducible build (see linkt to installation instruction here). 
```

docs/developers/coredev/release.md
```diff
@@ -115,6 +115,23 @@ on the `stdout` if any test fails.
...
+These files need to be updated regularly, especially before a release. You can find these files at
```
Czaki:
```suggestion
These files need to be updated weekly at monday. Such update may be triggere manually by core dev also. You can find these files at
```

docs/developers/coredev/release.md
```diff
@@ -115,6 +115,23 @@ on the `stdout` if any test fails.
...
+[resources/constraints](https://github.com/napari/napari/tree/main/resources/constraints).
```
Czaki, line 122:
```suggestion
[resources/constraints](https://github.com/napari/napari/tree/main/resources/constraints).

To get updated constraints for your PR use `@napari-bot update constraints` in PR comment, then folow instruction in comment added by a bot.
```

Czaki (dismissed):
I have added a few suggestions with information that should be added, but I think that not using my wording. If you agree with me, feel free to write it the own way.

melissawm:
Updated, thanks @Czaki !

psobolewskiPhD (approved):

psobolewskiPhD:
Thanks @melissawm ❤️

melissawm:
Hi friends, any last comments?

PR #463, opened 2024-07-22
Add actionlint to prevent GHA workflow errors

jni:
This adds actionlint from napari/napari#7049 to help prevent errors in GHA workflows.

**Update:** @Czaki also fixed the errors found by actionlint and ensured that new docs builds from tags (a) end up in their own folder (rather than dev/), and (b) don't include alpha, beta and rc tags.

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -106,11 +106,18 @@ jobs:
 
       - name: get directory name
         # if this is a tag, use the tag name as the directory name else dev
+        env:
+          REF: ${{ github.ref }}
         run: |
-          if [[ "${{ github.ref }}" == "refs/tags/"* ]]; then
-            echo "branch_name=${{ github.ref_name }}" >> $GITHUB_ENV
+          TAG="${GITHUB_REF/refs\/tags\/v/}"
+          VER="${TAG/a*/}"  # remove alpha identifier
+          VER="${VER/b*/}"  # remove beta identifier
+          VER="${VER/rc*/}"  # remove rc identifier
+
+          if [[ "$REF" == "refs/tags/v"* ]]; then
+            echo "branch_name=$VER" >> "$GITHUB_ENV"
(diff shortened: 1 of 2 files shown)


jni:
@Czaki, actionlint [fails with errors in these lines](https://github.com/napari/docs/actions/runs/10037254013/job/27736651600?pr=462#step:3:27):

https://github.com/napari/docs/blob/ffc22d0f2ecaa7c7e0a3a4cf7b0a616cdeddf4c4/.github/workflows/build_and_deploy.yml#L109-L114

> .github/workflows/build_and_deploy.yml:109:9: shellcheck reported issue in this script: SC2193:warning:1:27: The arguments to this comparison can never be equal. Make sure your syntax is correct [shellcheck]

which explains why the deployment with 0.5.0 only happened in dev and didn't go into a folder. Do you have ideas on how to fix it?

.github/workflows/build_and_deploy.yml
```diff
@@ -106,11 +106,18 @@ jobs:
...
+          TAG="${GITHUB_REF/refs\/tags\/v/}"
+          VER="${TAG/a*/}"  # remove alpha identifier
+          VER="${VER/b*/}"  # remove beta identifier
+          VER="${VER/rc*/}"  # remove rc identifier
+
+          if [[ "$REF" == "refs/tags/v"* ]]; then
+            echo "branch_name=$VER" >> "$GITHUB_ENV"
           else
-            echo "branch_name=dev" >> $GITHUB_ENV
+            echo "branch_name=dev" >> "$GITHUB_ENV"
```
Czaki, line 120: ealier code have 2 problems. 

- it will put `v0.5.1` as version instead of `0.5.0`
- it will create folders for rc, alpha and beta. 

I think that this code should work. We may test it by creating 0.5.1b1 and check if `0.5.0` is available (still may be not selectable from dropdown).

Czaki (approved):

PR #475, opened 2024-08-11
Add desctiption about `numpy<2` macOS arm problems

Czaki:
https://github.com/napari/napari/pull/7178

# Description

Add information about `numpy<2` macOS ARM problems. 
required by https://github.com/napari/napari/pull/7178

### DIFF
--- a/docs/tutorials/fundamentals/installation.md
+++ b/docs/tutorials/fundamentals/installation.md
@@ -128,6 +128,7 @@ notation.)*
 
 *(See [Using constraints file](#using-constraints-files) for help installing older versions of napari)*
 
+*(See [Problem with `numpy<2` on macOS arm](#problem-with-numpy<2-on-macos-arm) for a known issue with `numpy` on macOS arm64 machines)*
 :::::
 
 :::::{tab-item} From the main branch on Github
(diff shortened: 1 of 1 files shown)


docs/tutorials/fundamentals/installation.md
```diff
@@ -223,6 +224,23 @@ For example, if you would like to install napari on python 3.10:
 pip install napari[all, pyqt] -c constraints_py3.10.txt
 ```
 
+### Problem with `numpy<2` on macOS arm
```
Czaki, line 227: @melissawm How to get a nice anchor for this, for easier link to this part of documentation?

jni: @Czaki 

```
(anchor-text)=
# Actual text
```

Ref: https://jupyterbook.org/en/stable/reference/cheatsheet.html#target-headers

jni:
@Czaki I thought the workaround on napari/napari#7178 would mean that this problem wouldn't come up? (And so these docs wouldn't be needed?)

jni:
(Basically — I don't think we should release this if this documentation is accurate.)

jni:
(But if it's *performance* issue, not a crash issue, then it's ok.)

Czaki:
> @Czaki I thought the workaround on [napari/napari#7178](https://github.com/napari/napari/pull/7178) would mean that this problem wouldn't come up? (And so these docs wouldn't be needed?)

Any plugging that will use threads may trigger this. And bug is specyfic for macos arm, so if deweloper is using Linux or Windows. An example you may se here https://github.com/scverse/napari-spatialdata/pull/283 (initial implementation of loading data in thread triggered this, now macos arm is without threads).

docs/tutorials/fundamentals/installation.md
melissawm: We also autogenerate anchors for sections until level 4: https://github.com/napari/docs/blob/b979d31312a57575c5d0ee98f7463d2de80bac41/docs/conf.py#L197 and you can see the autogenerated anchors like so:

```
$ myst-anchors -l 3 docs/tutorials/fundamentals/installation.md
```

Because there are (to me) non intuitive rules around special characters since you have quotes and a `<` in the section title I would create a manual anchor like Juan said.

Czaki:
If the current state of https://github.com/napari/napari/pull/7178 (changing BLAS tread num) will be accepted then this PR is obsolete.

Czaki:
Closes because of napari/napari#7178 is merged with BLAS solution

PR #477, opened 2024-08-12
Improve deploy workflow by use deploy directory from workflow dispatch input 

Czaki:

# Description

As `convictional/trigger-workflow-and-wait` do not pass proper reference to workflow, this PR changed to use workflow dispatch input. This will allow triggering deploy docs manually to a given directory without the need to create a tag.

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -17,6 +17,11 @@ on:
     branches:
       - main
   workflow_dispatch:
+    inputs:
+      target_directory:
+        description: 'The directory to deploy the docs to'
+        required: true
+        default: 'dev'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_deploy.yml
```diff
@@ -17,6 +17,11 @@ on:
     branches:
       - main
   workflow_dispatch:
+    inputs:
+      target_directory:
+        description: 'The directory to deploy the docs to'
+        required: true
+        default: 'dev'
```
jni, line 24: @Czaki one question here — what happens when you trigger manually? You get a small text field to fill in?

Czaki: Yes. You may play with the benchmark workflow on the main repo.

jni: Neat! ❤️

melissawm (approved):
Interesting! Thanks @Czaki !

melissawm:
I think this is ready to go, starting the clock if there are no objections.

PR #483, opened 2024-08-23
Move images to _static folder

melissawm:
Related to https://github.com/napari/napari.github.io/pull/412 and other considerations around artifact size

# Description
This PR moves all images to be under _static/images with the rationale that it will potentially be easier to compress/redirect them and keep the artifact size smaller.

If this is not a good idea, I'm happy to close. It just seems better to gather all binaries in the same folder and not have them spread over the repo.


jni (approved):
thanks @melissawm! I think we probably will want to start using some kind of folder structure in there, but we can think about that later. 😅

Issue #490, opened 2024-09-02
Add napari docs build action to release process, including creating folder for new release in napari.github.io and updating symlink


melissawm:
Related to #486

PR #497, opened 2024-09-26
Add a release action

### LINKED ISSUE #486: Deploy action should deploy to both the release folder *and* dev on release
I just had a minor heart attack thinking that the docs deployment had failed for 0.5.3 because the release notes didn't appear in napari.org/dev. What had actually happened though, is that the release docs had been pushed to napari.org/0.5.3 but *not* dev. (The previous build to dev had been cancelled by the release build.) Therefore, dev was actually *behind* stable, which would be fixed soon enough but is a weird state to be in.

I think this would be most easily solved by making two copies.

melissawm:
Addresses https://github.com/napari/docs/issues/490
Closes https://github.com/napari/docs/issues/486
Depends on https://github.com/napari/napari.github.io/pull/417

# Description
This action will build and deploy docs for a new release, copying the built docs artifacts to both the dev and the {version} folders in napari.github.io. It will also update the stable symlink to point to the new version folder.

This action will be triggered by a new release tag pushed to the repo or by manually trigerring the action with an input specifying the new version (any input different from the default dev will trigger the action.)


melissawm:
I am creating this as a draft, since there will need to be at least some docs written around this strategy, and I was hoping to get some feedback before we commit to it.

.github/workflows/build_and_deploy.yml
```diff
@@ -109,13 +111,55 @@ jobs:
...
+      # Because the actions-gh-pages triggers a pages deployment, it can be some
+      # time before the repo is updated. We need to wait for the deployment to
+      # complete before we can create a second deployment, or we will have a git
+      # conflict. The 5m sleep is completely arbitrary.
+      - name: Wait for first deployment
+        if: (github.event_name == 'release' || (github.event_name == 'workflow_dispatch' && github.event.inputs.target_directory != 'dev'))
+        run: sleep 5m
+
+      - name: Deploy Release Docs
+        if: (github.event_name == 'release' || (github.event_name == 'workflow_dispatch' && github.event.inputs.target_directory != 'dev'))
```
melissawm: This is not correct, since it will update the symlink to release candidates as well.

.github/workflows/build_and_deploy.yml
```diff
@@ -109,13 +111,55 @@ jobs:
...
+
+      - name: Set up stable symlink
+        if: (github.event_name == 'release' || (github.event_name == 'workflow_dispatch' && github.event.inputs.target_directory != 'dev'))
+        uses: convictional/trigger-workflow-and-wait@v1.6.5
+        with:
+          owner: napari
+          repo: napari.github.io
+          github_token: ${{ secrets.WORKFLOW_PAT }}
+          workflow_file_name: symlink-stable.yml
+          client_payload: '{"target_directory": "${{ github.event.inputs.target_directory || github.event.release.tag_name }}"}'
```
melissawm, line 175: Next, we also need to add a step to update the version_switcher.json file.

.github/workflows/build_and_deploy.yml
```diff
@@ -109,13 +111,55 @@ jobs:
...
+
+      - name: Deploy Release Docs
+        if: (github.event_name == 'release' || (github.event_name == 'workflow_dispatch' && github.event.inputs.target_directory != 'dev'))
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          deploy_key: ${{ secrets.ACTIONS_DEPLOY_KEY }}
+          external_repository: napari/napari.github.io
+          publish_dir: ./html
+          publish_branch: gh-pages
+          destination_dir: ${{ github.event.inputs.target_directory || github.event.release.tag_name }}
```
melissawm, line 163: This could potentially create a wrong folder if the wrong version string is passed as input to the action.

psobolewskiPhD:
I think this makes sense to me and would make releases easier--they're happening more frequently now, so it's worth it!

.github/workflows/build_and_deploy.yml
melissawm: I think I addressed this now. Will test on my fork.

.github/workflows/build_and_deploy.yml
melissawm: I'm not sure how to validate this entry :(

melissawm:
This is ready for another look, thanks all!

melissawm:
Converting to draft until I check for https://github.com/napari/napari.github.io/pull/418

TimMonko:
Closing because the issues are still open and it will probably be easier to make new PRs.
Thanks for your work @melissawm :) Re-open if you like

Issue #377, opened 2024-03-24
[Maint] napari now uses pyproject.toml and ruff


psobolewskiPhD:
## 🧰 Task

Check the developer guides (and plugin guides?) for references to setup.cfg and flake8, isort, etc. and update as needed.

PR #502, opened 2024-10-22
Update docs to remove flake8, isort mentions and add ruff

melissawm:
Partially addresses #377. 

Mentions to setup.cfg remain in the plugin docs which need to be updated separately.

# Description
Updates mentions of flake8 and isort by ruff.

Also removes some explicit mentions of ignore rules since these get out of date quickly and can be checked directly in the pyproject.toml file.


DragaDoncila (approved):
LGTM! Thanks @melissawm

docs/developers/contributing/dev_install.md
```diff
@@ -88,13 +88,9 @@ In order to make changes to `napari`, you will need to [fork](https://docs.githu
...
 
    Code will also be linted to enforce the stylistic and logistical rules specified
-   in our [`flake8` configuration](https://github.com/napari/napari/blob/main/pyproject.toml), which currently ignores
-   [E203](https://lintlyci.github.io/Flake8Rules/rules/E203.html),
-   [E501](https://lintlyci.github.io/Flake8Rules/rules/E501.html),
-   [W503](https://lintlyci.github.io/Flake8Rules/rules/W503.html) and
-   [C901](https://lintlyci.github.io/Flake8Rules/rules/C901.html).  For information
-   on any specific flake8 error code, see the [Flake8
-   Rules](https://lintlyci.github.io/Flake8Rules/).  You may also wish to refer to
+   in our [`ruff` configuration](https://github.com/napari/napari/blob/b9dc8dca1e1afa7567586e7ffa7a36cc7cd9fb6f/pyproject.toml#L219).  For information
```
Czaki: I am not convinced that we should pin a specific commit, as configuration may change in time. We should rather write something like "`[tool.ruff]` section in [pyproject.toml](https://github.com/napari/napari/blob/main/pyproject.toml)"

melissawm: Sounds good, done.

docs/plugins/virtual_environment_docs/4-developer-tools.md
```diff
@@ -25,12 +25,6 @@ These _check_ your code.
 These _auto-modify_ your code.
 * [black](https://github.com/psf/black)
   Forces code to follow specific style, indentations, etc...
```
Czaki: We are using `ruff-format`, not `black` now. 
It is slimier to black, but not equivalent.

melissawm: Ah right -I only see one rule though

https://github.com/napari/napari/blob/b9dc8dca1e1afa7567586e7ffa7a36cc7cd9fb6f/pyproject.toml#L245

so I'll again just point people to the pyproject.toml in case there are future updates.

Czaki: yes. The list of used linters is `.pre-commit-config.yaml`  https://github.com/napari/napari/blob/b9dc8dca1e1afa7567586e7ffa7a36cc7cd9fb6f/.pre-commit-config.yaml#L6

willingc: @melissawm If you scroll down from your link, ruff is performing all of the task that you had in the original text since it is a drop-in replacement for these tools. Perhaps link out to ruff docs: https://docs.astral.sh/ruff/linter/#the-ruff-linter

willingc (approved):

docs/plugins/virtual_environment_docs/4-developer-tools.md
```diff
@@ -25,12 +25,6 @@ These _check_ your code.
 These _auto-modify_ your code.
 * [black](https://github.com/psf/black)
```
psobolewskiPhD:
```suggestion
* [ruff-formater](https://docs.astral.sh/ruff/formatter/)
```

psobolewskiPhD:
I popped the `ready to merge` because there is one more place where black needs to be replaced with ruff-formatter.
Otherwise, looks great.

melissawm:
Updated 🙏🏻

PR #505, opened 2024-10-25
Add information about `qt_viewer` fixture

Czaki:

Try to address https://github.com/napari/napari/pull/7251#issuecomment-2363381615

# Description

Add information about `qt_viewer` fixture and try to clarify when to use it.

### DIFF
--- a/docs/developers/contributing/testing.md
+++ b/docs/developers/contributing/testing.md
@@ -259,13 +259,65 @@ def test_something_else(qtbot):
     ...
 ```
 
+(qt_viewer)=
+#### `qt_viewer` and `viewer_model`
+
+For the test that are using only `ViewerModel` api or are only checking 
+rendering of the viewer, since `napari==0.5.4` we have implemented the `qt_viewer` [pytest fixture](https://docs.pytest.org/en/stable/explanation/fixtures.html).
+For the current moment, it is only for internal use and is not exported to the global scope, 
+as it is defined in `conftest.py` file. 
+
+The `qt_viewer` fixture return the instance of {class}`~napari.qt.QtViewer` class.
+This class do not provide the same api as the {class}`~napari.ViewerModel` class, 
+but have associated {class}`~napari.ViewerModel` instance, which can be accessed by `viewer` attribute.
+Alternatively you could use `viewer_model` fixture, which return this instance of {class}`~napari.ViewerModel` class.
+
+```python
+def test_something(qt_viewer):
+    qt_viewer.viewer.add_image(np.random.random((10, 10)))
+    assert len(viewer.layers) == 1
+    assert viewer.layers[0].name == 'Image'
+```
+
+or 
+
+```python
+def test_something(qt_viewer, viewer_model):
+    viewer_model.add_image(np.random.random((10, 10)))
+    assert len(viewer.layers) == 1
+    assert viewer.layers[0].name == 'Image'
+```
+
+The `qt_viewer` fixture take care about proper teardown of all qt widgets related to the viewer.
+If you need to adjust the QtViewer for a given [test file](https://docs.pytest.org/en/stable/how-to/fixtures.html#override-a-fixture-on-a-test-module-level) you can use `qt_viewer_` fixture.
+
+```python
+@pytest.fixture
+def qt_viewer(qt_viewer_):
+    # in this file we need to have added data and 3d view for all tests in file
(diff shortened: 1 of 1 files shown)


docs/developers/contributing/testing.md
```diff
@@ -259,13 +259,65 @@ def test_something_else(qtbot):
...
+
+For the test that are using only `ViewerModel` api or are only checking 
+rendering of the viewer, since `napari==0.5.4` we have implemented the `qt_viewer` [pytest fixture](https://docs.pytest.org/en/stable/explanation/fixtures.html).
+For the current moment, it is only for internal use and is not exported to the global scope, 
+as it is defined in `conftest.py` file. 
+
+The `qt_viewer` fixture return the instance of {class}`~napari.qt.QtViewer` class.
+This class do not provide the same api as the {class}`~napari.ViewerModel` class, 
+but have associated {class}`~napari.ViewerModel` instance, which can be accessed by `viewer` attribute.
+Alternatively you could use `viewer_model` fixture, which return this instance of {class}`~napari.ViewerModel` class.
```
melissawm:
```suggestion
Since `napari==0.5.4` we have implemented the `qt_viewer` [pytest fixture](https://docs.pytest.org/en/stable/explanation/fixtures.html) which can be used for tests that are only using the `ViewerModel` api or are only checking rendering of the viewer. 
For the current moment, it is only for internal use and is not exported to the global scope, 
as it is defined in `conftest.py` file. 

The `qt_viewer` fixture returns the instance of the {class}`~napari.qt.QtViewer` class.
This class does not provide the same api as the {class}`~napari.ViewerModel` class, 
but has an associated {class}`~napari.ViewerModel` instance, which can be accessed by the `viewer` attribute.
Alternatively, you could use the `viewer_model` fixture, which returns this instance of {class}`~napari.ViewerModel` class.
```

docs/developers/contributing/testing.md
```diff
@@ -259,13 +259,65 @@ def test_something_else(qtbot):
...
+
+```python
+def test_something(qt_viewer, viewer_model):
+    viewer_model.add_image(np.random.random((10, 10)))
+    assert len(viewer.layers) == 1
+    assert viewer.layers[0].name == 'Image'
+```
+
+The `qt_viewer` fixture take care about proper teardown of all qt widgets related to the viewer.
+If you need to adjust the QtViewer for a given [test file](https://docs.pytest.org/en/stable/how-to/fixtures.html#override-a-fixture-on-a-test-module-level) you can use `qt_viewer_` fixture.
```
melissawm:
```suggestion
The `qt_viewer` fixture takes care of proper teardown of all qt widgets related to the viewer.
If you need to adjust the QtViewer for a given [test file](https://docs.pytest.org/en/stable/how-to/fixtures.html#override-a-fixture-on-a-test-module-level) you can use the `qt_viewer_` fixture.
```

docs/developers/contributing/testing.md
```diff
@@ -259,13 +259,65 @@ def test_something_else(qtbot):
...
 #### `make_napari_viewer`
 
-We provide a
-[pytest fixture](https://docs.pytest.org/en/stable/explanation/fixtures.html) called
-`make_napari_viewer` for tests that require the {class}`~napari.Viewer`. This
-fixture is available globally and to all tests in the same environment that `napari`
+For more complex test cases where we need to test fully application behaviour 
+(for example using `viewer.window` API) we can use `make_napari_viewer` [pytest fixture](https://docs.pytest.org/en/stable/explanation/fixtures.html).
+However, the creating and teardown of the whole viewer is more fragile and slower than using just `qt_viewer` fixture.
+This fixture is available globally and to all tests in the same environment that `napari`
```
melissawm:
```suggestion
For more complex test cases where we need to fully test application behaviour 
(for example, using the `viewer.window` API) we can use `make_napari_viewer` [pytest fixture](https://docs.pytest.org/en/stable/explanation/fixtures.html).
However, the creating and teardown of the whole viewer is more fragile and slower than using just the `qt_viewer` fixture.
This fixture is available globally and to all tests in the same environment that `napari`
```

melissawm (approved):
Left a couple of wording suggestions, take them or leave them! Thanks @Czaki

psobolewskiPhD (approved):
This is really nice, thanks to both of you!
❤️

PR #512, opened 2024-11-11
Add notes about versionadded/versionchanged directives to dev guide

### LINKED ISSUE #7192: Add .. versionadded:: directives when modifying APIs
## 📚 Documentation

The versionadded directive is a really nice way to tell users when particular arguments or functions were added to an API (example [here](https://github.com/scikit-image/scikit-image/blob/8ba3bfffa79f24810e3be8e5f833ea7e0f954750/skimage/filters/_median.py#L31)). We haven't been using them in napari, which leads to some confusion about when particular APIs were added (who has time to read through a bunch of release notes?) and therefore what the minimum version of napari is to be able to use those APIs (ex: https://github.com/BiAPoL/napari-clusters-plotter/pull/308#issuecomment-2285460086).

melissawm:
Closes napari/napari#7192
Complements napari/napari#7359

# Description
Adds note about using versionadded/versionchanged directives to the core dev and contributing guides.

### DIFF
--- a/docs/developers/contributing/index.md
+++ b/docs/developers/contributing/index.md
@@ -120,7 +120,7 @@ identify the cause and where to optimize.
 
 ### API Changes
 
-If you are changing and existing API or adding a new one, make sure you update
+If you are changing an existing API or adding a new one, make sure you update
 the corresponding docstring to contain the `.. versionadded::` or
 `.. versionchanged::` directive. For more information see the
 [Sphinx documentation](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#describing-changes-between-versions).
--- a/docs/developers/contributing/index.md
+++ b/docs/developers/contributing/index.md
@@ -123,7 +123,9 @@ identify the cause and where to optimize.
 If you are changing an existing API or adding a new one, make sure you update
 the corresponding docstring to contain the `.. versionadded::` or
 `.. versionchanged::` directive. For more information see the
-[Sphinx documentation](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#describing-changes-between-versions).
+[Sphinx documentation](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#describing-changes-between-versions). 
+Please also consider documenting any major features/changes in our
+[tutorials](tutorials) and other [usage documentation](usage).
 
 ### Contributing documentation
 
--- a/docs/developers/contributing/index.md
+++ b/docs/developers/contributing/index.md
@@ -118,6 +118,13 @@ also include benchmarks to show the new feature is not too slow but this is less
 poor performance, [profiling](profiling) and [performance monitoring](napari-perfmon) can help
 identify the cause and where to optimize.
 
+### API Changes
+
+If you are changing and existing API or adding a new one, make sure you update
+the corresponding docstring to contain the `.. versionadded::` or
+`.. versionchanged::` directive. For more information see the
+[Sphinx documentation](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#describing-changes-between-versions).
+
 ### Contributing documentation
 
 See [](contributing-docs) for information on how to contribute documentation.
--- a/docs/developers/coredev/core_dev_guide.md
+++ b/docs/developers/coredev/core_dev_guide.md
@@ -95,9 +95,13 @@ version numbers `0.x` and do not have a deprecation policy, but we will work to 
 4. **Documentation and tutorials:** All new methods should have appropriate doc
 strings following [PEP257](https://peps.python.org/pep-0257/) and the
 [NumPy documentation guide](https://numpy.org/devdocs/dev/howto-docs.html#documentation-style).
-For any major new features, accompanying changes should be made to our
-[tutorials](tutorials). These should not only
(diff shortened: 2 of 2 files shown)


docs/developers/contributing/index.md
```diff
@@ -118,6 +118,13 @@ also include benchmarks to show the new feature is not too slow but this is less
 poor performance, [profiling](profiling) and [performance monitoring](napari-perfmon) can help
 identify the cause and where to optimize.
 
+### API Changes
+
+If you are changing and existing API or adding a new one, make sure you update
```
lucyleeow:
```suggestion
If you are changing an existing API or adding a new one, make sure you update
```

docs/developers/coredev/core_dev_guide.md
```diff
@@ -95,9 +95,13 @@ version numbers `0.x` and do not have a deprecation policy, but we will work to
...
-For any major new features, accompanying changes should be made to our
-[tutorials](tutorials). These should not only
-illustrates the new feature, but explains it. Titles for all documents in napari should follow [sentence case capitalization](https://apastyle.apa.org/style-grammar-guidelines/capitalization/sentence-case), but the name `napari` should always be written in lowercase.
+For API changes, make sure to check that a `.. versionadded::` or
+`.. versionchanged::` directive has been added to the appropriate docstring. For
+any major new features, accompanying changes should be made to our
+[tutorials](tutorials). These should not only illustrate the new feature, but
+explain it. Titles for all documents in napari should follow
+[sentence case capitalization](https://apastyle.apa.org/style-grammar-guidelines/capitalization/sentence-case),
+but the name `napari` should always be written in lowercase.
```
lucyleeow, line 104: Glad to have this documented ! The other question, should napari always be in backticks?

lucyleeow (approved):
Thank you!

Do you think it's worth adding an example? E.g.,

```python
class Layer(KeymapProvider, MousemapProvider, ABC, metaclass=PostInit):
    """Base layer class.

    Parameters
    ----------
    data : array or list of array
        Data that the layer is visualizing. Can be N-dimensional.

        ... versionchanged::
            Data can now be a list of arrays.
```

this was probably a bad example as this wasn't changed, I don't know what API has been changed recently.

docs/developers/coredev/core_dev_guide.md
lucyleeow: I wonder if this part should be added somewhere more generally for all users (not just core devs) e.g., in our guide to contributing documentation?

melissawm: No I think this was meant to show how you should write napari in the sources.

I thought about it and decided against adding this to the general "contributing docs" pages for two reasons. One is that if you are contributing to the docs in general you are not going to be changing APIs so you will not face this case. The other is that I see adding this directive in general as the job of the core team, in that you should probably suggest this on review same as you would do for adding a release note, for example. So as long as the core team is informed, I think we are good.

lucyleeow: Thats a good point. I can see both sides. Pros of adding it to our general contributing guide (agree, docs is not the right place, should be in code contributing docs e.g., if you are changing the API ...):

* it highlights the importance of documentation (this is the culture we want!)
* I think I'd prefer not to add 'extra work' to maintainers and not put the onus on reviewer to check this. We could suggest to contributors that this should be added, but maintainers can handle it if they prefer

BUT I get that we may want to make it easier for contributors, and default let maintainers handle this.

Happy either way.

melissawm: So do you have a suggestion on the wording to use instead of the note I added above to the contributing/index doc?

docs/developers/contributing/index.md
```diff
@@ -118,6 +118,13 @@ also include benchmarks to show the new feature is not too slow but this is less
 poor performance, [profiling](profiling) and [performance monitoring](napari-perfmon) can help
 identify the cause and where to optimize.
 
+### API Changes
+
+If you are changing an existing API or adding a new one, make sure you update
+the corresponding docstring to contain the `.. versionadded::` or
+`.. versionchanged::` directive. For more information see the
+[Sphinx documentation](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#describing-changes-between-versions).
```
psobolewskiPhD:
```suggestion
[Sphinx documentation](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#describing-changes-between-versions). 
Please also consider documenting any major features/changes in our
[tutorials](tutorials) and other [usage documentation](usage).
```
How about this as a middle ground? encourage but not require?
At the very least would be nice to have issues created that a new feature requires docs -- this is something we should do in review.

psobolewskiPhD (approved):
Small suggestion, based on the discussion.

docs/developers/contributing/index.md
melissawm: Oh! I think I misunderstood @lucyleeow 's suggestion. I thought she meant adding a note about the versionadded/versionchanged directives in the "contributing docs" section, not the other way around. Sure, this is reasonable!

psobolewskiPhD: I'll let @lucyleeow merge (feel free) or mark ready-to-merge, in case I also misunderstood.

lucyleeow: Yup that's exactly what I was thinking, thanks everyone, I'll merge!

PR #518, opened 2024-11-13
Update benchmarks.md to use conda napari-env

willingc:
If the user followed instructions in our docs, they would use `napari-env` as their conda environment name. This makes it consistent.

The `venv` instructions do not provide a specific name so leaving that as is.


melissawm (approved):

melissawm:
I am merging this as a typo - thanks @willingc !

PR #519, opened 2024-11-15
Add contributing text and link about CI

willingc:
This is a follow up from https://github.com/napari/napari/issues/7364#issuecomment-2470394033

# Description
Adds text and link to CI CONTRIBUTING.md doc in napari/napari.

### DIFF
--- a/docs/developers/contributing/index.md
+++ b/docs/developers/contributing/index.md
@@ -84,6 +84,12 @@ keeping napari maintainable as it grows.
 We have dedicated documentation on [testing](napari-testing) that we recommend you
 read as you're working on your first contribution.
 
+### Automation and CI
+
+We use GitHub Actions to automate our continuous integration and project workflows.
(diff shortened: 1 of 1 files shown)


psobolewskiPhD (approved):
at some point we'll want more robust docs for our CI, especially the testing stuff, but this is certainly a start!

melissawm (approved):
Do we also want to point to the docs CI documentation just in case or is that too much information here? https://napari.org/stable/developers/contributing/documentation/docs_deployment.html

melissawm:
Sorry for the off-topic but it looks like the labeler is running successfully, but not adding labels?

psobolewskiPhD:
I think it's this issue:
https://github.com/actions/labeler/issues/814
They closed it with just a workaround--lets give it a shot.

(why do we not have these issues in napari/napari? it's also on v5 of the action)

Edit: i think it's the syntax, it's not recursing docs. PR inc.

Edi2: Yup the PR got auto labeled correctly, so the action is fine, but the syntax was too specific.

PR #514, opened 2024-11-11
Update tutorials

melissawm:
# References and relevant issues
Related to #280

# Description
Fixes a few typos, typesetting details.

Updates content for the viewer tutorial, including new gui elements.

Images will also need to be updated.

PR #520, opened 2024-11-18
Update napari version used in tutorial

willingc:
It would be nice to update this to the more recent 0.5 version.

I noticed this when looking through the tutorial. It's a nit but it would be nice to cite version 0.5


psobolewskiPhD:
oof, nice catch, that tag is likely stale. But I'm also not sure that everything in the guide is up to date for 0.5.x? I will try to check.

melissawm:
I'm not against this but I am updating the whole tutorials subfolder here: https://github.com/napari/docs/pull/514

In that PR I actually chose to remove that line, since now we are using versioned documentation and I don't think we need to mention that again in the text.

willingc:
Cool @melissawm. I wasn't aware of that. I prefer no version. Closing :smile:

PR #526, opened 2024-11-19
Update labeler.yml to check subfolders

psobolewskiPhD:
The labeler action still hasn't auto-labeled stuff:
https://github.com/napari/docs/pull/519#issuecomment-2485959231

# Description
This PR makes the labeler recurse through `docs`, which should fix the lack of labeling noted above.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,7 +1,7 @@
 # See: .github/workflows/labeler.yml and https://github.com/marketplace/actions/labeler
 documentation:
 - changed-files:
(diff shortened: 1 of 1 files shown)


psobolewskiPhD:
We could consider using `docs/**/*.md` and then breaking out the other types into a different tag?

psobolewskiPhD:
This PR got labeled fine, so I think that confirms the issue with missing labels was the lack of recursing, which hopefully this fixes.

Czaki (approved):

PR #528, opened 2024-11-30
Fix linkchecker

melissawm:
--

# Description
The linkchecker is failing due to the autogenerated docs that are not being generated - added this step to the linkcheck make target.

See for example https://github.com/napari/docs/actions/runs/12091824973/job/33720711245


DragaDoncila (approved):
Thanks @melissawm let's get this in and see if it fixes things!

psobolewskiPhD (approved):

PR #531, opened 2024-12-05
Add version warning banner for old versions of the docs

### LINKED ISSUE #82: Add warning banner to older versions of documentation
## 📚 Documentation

From https://github.com/napari/napari/pull/5096#issuecomment-1250673847:

> it would also be great if we had a nice colorful warning on the website that tells people that that's not the latest version (see for example the [numpy docs](https://numpy.org/doc/1.19/index.html))

melissawm:
Closes #82

# Description
We now get the version warning banner for free so we can turn it on :) 

It looks like this currently:

Captura de imagem_20241205_171236

I'm happy to do some CSS styling if we are not happy with how it looks!

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -29,3 +29,22 @@ div.sphx-glr-download a:hover {
   background-color: var(--napari-gray);
   color: #222832;
 }
+
+/* Version warning banner color */
+#bd-header-version-warning {
+  background-color: color-mix(in srgb, var(--pst-color-secondary-bg), transparent 30%);
+}
+
+#bd-header-version-warning .pst-button-link-to-stable-version {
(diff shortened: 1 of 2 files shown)


DragaDoncila:
I think the version warning banner looks good! I kinda wanna remove the "chat with us" banner though at this point. I don't think it should be a permanent fixture on the website, and seeing the stacked banners now is making my eye twitch lol. I remember some conversation about how we couldn't just 'turn it off'? Is that still true?

psobolewskiPhD:
this will also warn when the page is old right? I still hit old tutorials and things via google and image.sc links

melissawm:
How do we feel about this?

Captura de imagem_20241223_154848

psobolewskiPhD (approved):
Looks good to me!

DragaDoncila (approved):
Looks good to me!

PR #546, opened 2024-12-30
Fix build-on-windows link in README.md

TimMonko:
# Description

* Fixes windows instruction link being to a 404, by adding `contributing` location
* Updates windows instructions link to use `dev` instead of `stable` to be consistent with others in readme


psobolewskiPhD (approved):
Thanks for catching that!

PR #560, opened 2025-02-03
Rename Gallery to Examples

TimMonko:
Introducing the world's smallest PR.

~~Rename the Gallery to Example Gallery.~~

**Edit:** After much discussion, rename Gallery to Examples

Why? Well, it's referred to as such everywhere else in the docs. 
Why else? Because honestly its a bit too hidden and makes me think I'm just going to browse some cute vacation photos, and not the most awesome example gallery of any python library.


jni:
I actually think our skimage gallery is a bit more awesome, but thank you for the compliment! 😂 

In the home page it's "Examples gallery". I'm not sure which is better. Certainly if it was dogs it would be "dog gallery", so I guess singular, but for some reason I can't articulate, Examples sounds better to me. Let me know what you prefer, or if I'm flat-out wrong 😂, and we'll let others weigh in too, but either way, maybe your tiniest PR record can wait for another time and here we should harmonise the home page to use the same title. 😂

psobolewskiPhD:
I'm pretty sure the general rule is that nouns used as adjectives should be singular, much like how adjectives don't have plurals.
So `Example Gallery` would be proper for `A Gallery of Examples`
I'll check with my local english teacher tho!

And yeah may as well also fix:
image
Few others: https://github.com/search?q=repo%3Anapari%2Fdocs%20%22examples%20gallery%22&type=code

Edit: friendly local english teacher suggests that `Example Gallery` would tend to be `an example of a gallery`, so she leans towards `Examples Gallery` after all.

Edit: After further deliberation and consulting 3 AIs, I'm leaning towards `Gallery of Examples` 🤣 
I think there are arguments made towards both forms!

TimMonko:
Well we could just go the scikitimage route since it's so great and just rename napari's to `Examples` 🤣 And it's not like there are other Examples or Galleries to compete with in the docs, so it makes sense, right?

(you see, my problem is that I dislike directly using matplotlib so much that I automatically dock points off scikitimage examples, but that is 100% a ME problem) 

@psobolewskiPhD I'm laughing way to hard about your grammar adventure, but I appreciate the points you are making!

TimMonko:
So, I made the very minimal change where now it's 'Examples' instead of 'Gallery'.

I did not update any of the main body texts, because I think 'Examples Gallery' still works when referring to the 'Gallery of Examples' aka the Examples. 

But, willing also to do the more detailed switching of names if that's preferred.

DragaDoncila:
I think `Examples` is good as the general term! I don't think the `Gallery` bit makes it more discoverable or meaningful. 

However, I did come here to say that I'm 100% on Peter's side, it should :100: be `Example Gallery` not `Examples Gallery` :rofl:

jni:
I'm also happy with Examples but I will point out that @psobolewskiPhD's English teacher was spot-on about why I didn't gel with "Example Gallery" though I couldn't put my finger on it — it definitely can be parsed (and would be most commonly parsed) as "an example of a gallery". I'd be curious to know whether they can think of other exceptions to the singular rule.

DragaDoncila:
> and would be most commonly parsed

Big if true 😂 an Art Gallery is a gallery of art, and surely that's the most common closely related term

melissawm:
Just a reminder that if we do this it will be worth grepping for "gallery" in the docs since there's quite a lot of references to a Gallery floating around.

jni (approved):

PR #564, opened 2025-02-12
Update BlueSky link to our actual account (not masto bridge)

jni:
We now have a proper BlueSky account linked to our domain. This updates the
link to avoid bridgy, which can have a middling experience. Also BlueSky's
domain branding is pretty awesome, I must admit.

I've also slightly tweaked the wording and formatting.


psobolewskiPhD (approved):
Thanks!

PR #588, opened 2025-02-21
Reorder sections and add section separators

willingc:

# Description


willingc:
Replaced by #632

PR #598, opened 2025-02-27
Re-add empty cli image for installation tutorial

melissawm:
N/A

# Description
Re-adds the image that is currently missing from the installation tutorial (follow-up to #514)


Czaki (approved):

psobolewskiPhD:
I posted in the issue instead of here, but my suggestion is to just use `nbscreenshot` and take a screenshot of the viewer. The command line, which will look different for everyone, doesn't add anything, the important part is people know what they should expect to see.
https://github.com/napari/napari/issues/7658#issuecomment-2691664276

That way, we don't need to worry about re-doing this one manually for every release.

TimMonko:
Image is missing both from
1. installation: https://napari.org/stable/tutorials/fundamentals/getting_started.html
2. how to launch: https://napari.org/stable/tutorials/fundamentals/getting_started.html

So if we do `nbscreenshot` (I agree this is a good idea), we should make sure its in both tutorials.

willingc (approved):

jni:
Closed in favour of #606

Issue #600, opened 2025-02-27
Plugins docs reorg


melissawm:
## 🐛 Bug

We should reevaluate the organization of the plugin docs, especially the content that was added as part of workshop material (e.g. https://napari.org/dev/plugins/virtual_environment_docs/index.html and https://napari.org/dev/plugins/testing_workshop_docs/index.html)

## Additional context

We talked about this in the docs working group meeting on Feb 27, 2025.

PR #603, opened 2025-02-28
Add cards to the Advanced Topics landing page for plugins

willingc:
Partially addresses #600 by improving the Advanced Topics landing page.

# Description
This PR adds cards to guide the reader to the npe2 specification. It also more clearly indicates that napari plugin engine v1 is deprecated.

@DragaDoncila @melissawm @psobolewskiPhD FYI

### DIFF
--- a/docs/plugins/advanced_topics/index.md
+++ b/docs/plugins/advanced_topics/index.md
@@ -26,7 +26,7 @@ A technical specification for how plugins can contribute additional functionalit
 
 ## Deprecated: napari plugin engine v1
 
-:::{important} Use npe2.
+:::{important}
 New plugins should use the `npe2` standard.
 We encourage first generation plugin authors to migrate their plugin to the `npe2` standard.
 :::
--- a/docs/plugins/advanced_topics/index.md
+++ b/docs/plugins/advanced_topics/index.md
@@ -8,13 +8,6 @@ Some plugin developers may find more **in-depth, technical information** helpful
 ## npe2 advanced topics and reference
 
 ::::{grid}
-:::{grid-item-card} Migration guide to `npe2`
-:link: npe2-migration-guide
-:link-type: ref
-Have a plugin written for the first generation plugin system? This guide will help you migrate to the `npe2` plugin system.
-
-:::
-
 :::{grid-item-card} npe2 manifest reference
 :link: plugin-manifest
 :link-type: ref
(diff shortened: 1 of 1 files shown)


willingc:
Rendered page

image

TimMonko:
Looks great! Would it make more sense to move the migration card to the npe1 section? 

Basically, I'm thinking that card encourages migration, so that migration card should be to it's right

Thinking it can move too because npe2 has been out for a while now

willingc:
> Looks great! Would it make more sense to move the migration card to the npe1 section?

Great suggestion @TimMonko. Updating PR.

psobolewskiPhD:
I agree with Tim that the migration should be "under the fold" so to speak, but the side-by-side looks wierd:
image

I think either we should try a different layout or maybe break the 2nd paragraph out of the tile and have it under the header?

willingc:
Let me noodle on this a bit more since I'm not fully happy with the deprecated section.

psobolewskiPhD:
What about an admonition above the two npe1 tiles? with `New plugins should use ...`

psobolewskiPhD:
I dig it:
image

Also: whoops I disabled actions when trying to show Tim on Zulip!

docs/plugins/advanced_topics/index.md
```diff
@@ -1,20 +1,50 @@
...
-Have a plugin written for the first generation plugin system? This guide will help you migrate to the new plugin system.
 
+A technical specification for how plugins can contribute additional functionality and features to napari.
+
+:::
+::::
+
+## Deprecated: napari plugin engine v1
+
+:::{important} Use npe2.
```
psobolewskiPhD:
```suggestion
:::{important}
```
If it's `important` it can't have a custom title:
https://sphinx-book-theme.readthedocs.io/en/stable/reference/kitchen-sink/admonitions.html
See also my screenshot here which makes the copy a little pushy:
https://github.com/napari/docs/pull/603#issuecomment-2692960416

psobolewskiPhD (approved):
small nitpick, but otherwise good to go I think

docs/plugins/advanced_topics/index.md
willingc: Fair enough. Thanks! I think that admonition improved things.

psobolewskiPhD: Does it look different for you? I see:
image

TimMonko:
Love where this got to! The admonition is a great solution

DragaDoncila (approved):
Looks awesome, thanks @willingc!

TimMonko (approved):

PR #605, opened 2025-03-01
Add NAPARI_DOCS_VERSION env var to docs

TimMonko:

Depends on [napari/#7659](https://github.com/napari/napari/pull/7659)

# Description

Adds NAPARI_DOCS_VERSION environment variable to `build_and_deploy.yml` when received from tagged napari/napari version releases.

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -75,8 +75,8 @@ jobs:
 
       - name: Set napari docs version
         run: |
-          if [[ "${{ github.event.client_payload.version }}" != "" ]]; then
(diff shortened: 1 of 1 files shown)


psobolewskiPhD:
@TimMonko According to Grzegorz here https://github.com/napari/napari/pull/7659#issuecomment-2692391394
we shouldn't need this. Looking at `conf.py` we already do parsing of versions between release and dev, so this shouldn't be needed.

TimMonko:
Thanks @psobolewskiPhD. I'll close. We can re-open if the docs don't build visually with the tag on the release candidates

Issue #568, opened 2025-02-15
check the viewer tutorial to ensure GUI changes in 0.6.0 are covered


psobolewskiPhD:
list of PRs: https://github.com/napari/napari/pulls?q=is%3Apr+is%3Aclosed+milestone%3A0.6.0+label%3A%22ui+change%22
- [x] copy: mention corner tick marks on buttons indicate right-click https://github.com/napari/docs/pull/576
- [x] update button icons graphics https://github.com/napari/docs/issues/563
- [x] document grid mode spacing change https://github.com/napari/docs/pull/573
- [x] Shapes path button https://github.com/napari/docs/pull/574
- [x] update screenshots of viewer
- [x] admonition of the 0.6.0 gui changes https://github.com/napari/docs/issues/599

psobolewskiPhD:
I think the new 2d-3d "button" (https://github.com/napari/napari/pull/7608) and the new viewer buttons bar as a result needs to be updated. I think the other ones changing just with the corner chevron are less crucial.

psobolewskiPhD:
We discussed this in the docs meeting and agreed that it's important to fully update the docs for 0.6.0.

PR #606, opened 2025-03-01
Replace two manual screenshots of the viewer (launch_cli_empty and launch_cli_image) with nbscreenshots

psobolewskiPhD:
Alternative to: https://github.com/napari/docs/pull/598
Part of: https://github.com/napari/docs/issues/568
Needed by: https://github.com/napari/napari/pull/7659

# Description
In two places we want to show the outcome of opening the viewer, so the user knows what to expect.
Currently, these are manual screenshots with a CLI in the background and the viewer in the foreground.
Everyones CLI is different and we include the CLI command as a code cell, so I think it's best to just show the viewer.
This makes it one less thing that needs to be manually updated when the viewer GUI changes.


jni (approved):
❤️

PR #609, opened 2025-03-02
Reorganize plugin landing page and remove redundant index file

### LINKED ISSUE #595: Plugins index is a bit circular / has redundant links
## 🐛 Bug

https://napari.org/stable/plugins/index.html 

Current Links:
- Start using Plugins --> Using Published plugins + Napari Hub
    - Using published plugins --> Finding and Installing Plugins 
- Finding and installing plugins --> Finding and Installing Plugins

To fix:
Have the initial two links be
1. Finding and Installing Plugins
2. napari Hub

TimMonko:

Closes #595

# Description
1. Changes plugin landing page tabs to link immediatelly to 'finding_and_installing_plugins' and the napari hub, rather than to the index of start_using_plugins, which contained nothing more than the same information now on the landing page.
2. Removes index from start_using_plugins because it is not needed
3. Updates TOC to remove the index and using finding and installing plugins as top level header


TimMonko:
Uhh, did CircleCI just go poop, or is there some clue to something in my changes that is wrong?
With all the switching to new way to build docs, I think I haven't been able to rebuild my local docs environment correctly, so was hoping for CI to test for me. Can't tell if I messed up, or CI did.

willingc:
That looks like a CI timeout to me.

psobolewskiPhD:
Re-ran CI. Not sure what's up -- that point in the process takes a while for me locally too.

psobolewskiPhD (approved):
Looks great to me! I never realized how odd the setup was requiring extra wasted clicks.

willingc (approved):
Looks good. Thanks @TimMonko for the usability improvement.

PR #612, opened 2025-03-03
Suppress detached head warning in prep_docs script

willingc:
# Description
Add a -c option to suppress the detached head warning when building docs, specifically when checking out a tag.


psobolewskiPhD (approved):
nice!

jni (approved):

PR #631, opened 2025-03-15
Update conf.py to use packaging to detect dev vs release versions

psobolewskiPhD:
If you go to dev docs right now:
https://napari.org/dev/index.html
The sidebar link to the release notes is broken, I think because the version is being treated as a release and not dev.
It points to:
https://napari.org/dev/release/release_0_6_0a0.html
which doesn't exist.
instead of the index.html which would be used for a dev release:
https://napari.org/dev/release/index.html

So in this PR I use packaging to check if a version is a dev version and also uses packaging to get just the base version number of pre-releases, so we will get release_0_6_0.html


willingc (approved):
Thank you! 🚢

psobolewskiPhD:
Checking:
https://output.circle-artifacts.com/output/job/c902cc9c-bfe0-46f3-90de-6c69c2cecbdf/artifacts/0/docs/docs/_build/html/index.html
Looks like it worked

willingc:
I'm not sure if it's caching or the deploy workflow but stable says (0.5.6) but directs to 0.6.0 🤷🏼‍♀️

psobolewskiPhD:
Can you explain where? when I go to napari.org stable is 0.5.6 and has for example the missing screenshot in installation guide. When I use the circle build, it's just building dev, so it should also have stable as napari.org stable.

docs/conf.py
```diff
@@ -35,7 +35,7 @@
 from napari._version import __version_tuple__
```
Czaki, line 35:
```suggestion
from napari._version import __version_tuple__
from packaging.version import parse as parse_version
```

docs/conf.py
```diff
@@ -35,7 +35,7 @@
...
 
 release = napari.__version__
```
Czaki:
```suggestion
release = parse_version(napari.__version__)
```

docs/conf.py
```diff
@@ -35,7 +35,7 @@
...
-if "dev" in release:
+if "dev" in release or parse_version(release).is_prerelease:
```
Czaki:
```suggestion
if release.is_devrelease:
```

docs/conf.py
```diff
@@ -35,7 +35,7 @@
...
     version = "dev"
 else:
     version = release
```
Czaki:
```suggestion
    version = release.base_version
```

Czaki (dismissed):
We should deploy prerelease to separate dirctory. (for 0.6.0a1 it should be 0.6.0) but we should not parse it manually when `packaging` is available.

Czaki:
Please update the title to reflect the current state

jni (approved):

PR #636, opened 2025-03-18
update left sidebar to persist resources

willingc:
# Description

During the past few docs meetings and community meetings, we have been talking about ways to provide users information about using napari and engaging them in the napari community. One idea that kept coming up was how to provide users easy access to troubleshooting information and release notes.

After a recent docs meeting, @TimMonko came up with the idea to put a few frequently used links in the left sidebar. He made a PR that would add the links to the sidebar on the homepage. Someone else commented that it would be nice to have the links on other pages.

I had played around with the napari sphinx theme and its templates a couple of months ago. I had the idea to update the left sidebar's jinja template to include a persistent section of frequently used links. Here's a before and after view:

Screenshot 2025-03-20 at 8 37 42 AM

Screenshot 2025-03-20 at 8 46 27 AM

### DIFF
--- a/docs/_templates/sidebar-nav-bs.html
+++ b/docs/_templates/sidebar-nav-bs.html
@@ -1,5 +1,8 @@
 {# Displays the TOC-subtree for pages nested under the currently active top-level TOCtree element. #}
 <nav class="bd-docs-nav bd-links" aria-label="{{ _('Section Navigation') }}">
+  <p class="caption" role="heading" aria-level="1">
+    <span class="caption-text">>Section Navigation</span>
+  </p>
   <div class="bd-toc-item navbar-nav">
     {{- generate_toctree_html(
     "sidebar",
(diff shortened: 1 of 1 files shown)


melissawm:
I wonder if we should have a caption for those resources, to separate them from the main left-sidebar toc? It's not a blocker though, this looks good!

melissawm:
Also just noted that in Usage, the release notes item is repeated: https://output.circle-artifacts.com/output/job/5f685cda-57da-4a7d-8fd8-6cd2c1a949af/artifacts/0/docs/docs/_build/html/usage.html

Maybe there's a way to avoid that, or rename the items so the difference is clear (the Usage item is a list of all release notes; the persistent navbar only shows the latest Release notes)

melissawm:
Sorry one more thing: Instead of having the Meeting schedule page in the community rubric, maybe we should just have the "Community calendar" item link to that page and be persistent in the sidebar?

Same goes for Troubleshooting - I'd say remove it from Usage and keep it only in the persistent sidebar.

TimMonko:
I'm not seeing it? I remember you shared an image the other day with something on the lower left of the sidebar that would stay on page changes. 
image

psobolewskiPhD:
From your screenshot @TimMonko

release

I like the idea of some thing breaking up the flow between the normal toc items for a page and then "pinned" sidebar.
Even some white space would work I think.

Frankly, to me, Tims screenshot looks pretty overwhelming right now, so I think giving this more thought is definitely warranted.

willingc:
Yeah, I need to free up a bit of bandwidth to apply the changes from my fork here.

TimMonko:
LOL thanks @psobolewskiPhD, my bad, but I guess that's a bit of the issue

And no worries @willingc , I just figured based on @melissawm feedback it was already in good shape

willingc:
Hello humans @psobolewskiPhD @melissawm @TimMonko, Here's what the latest looks like:

**Home page with just links no headings** (we sould need to rework a different template to add a heading)

Screenshot 2025-03-19 at 3 39 30 PM

**Usage page** The release notes is in both sections since it is in the page contents as well as a static link to the current release notes.

Screenshot 2025-03-19 at 3 40 02 PM

**Just a random page**

Screenshot 2025-03-19 at 3 40 22 PM

TimMonko:
Overall, this is a nice improvement to the website functionality and the presentation of these key links that are not sections and are now on the homepage (i.e. we've wisely removed them from the navbar). I think this actually goes more to my goal of making these links key highlights, by bringing them outside of just the homepage 😄 

I was wishy-washy about the headers 'Section Navigation' and 'Resources', but stewing on it for a while, it really grew on me. 

It even works nicely on mobile, because it naturally goes down the screen and is scroll-able. 

Thanks @willingc !

TimMonko (approved):

jni:
@willingc do you mind updating the top description with a description of motivation, where this has landed after discussion, and maybe some before/after screenshots? 🙏

psobolewskiPhD (approved):
Love it!
I was iffy at first, but the latest incarnation is great!
Thanks for putting it out there!

melissawm (approved):
Looks great!

Coming back to my earlier points:
- I feel like we can remove the "Meeting schedule" toctree entry from the Community page, since it will already be added to the persistent navbar. 
- Same think for Troubleshooting: can we remove it as a toctree entry from the Usage page?

Those pages will always be on screen so I feel like repeating them in the toctree might be confusing (even more for the calendar since it has two different titles in the toctree)

willingc:
@melissawm Good suggestions. I'm happy to that in a follow up PR.

willingc:
>@willingc do you mind updating the top description with a description of motivation, where this has landed after discussion, and maybe some before/after screenshots? 🙏

@jni how's it look now?

PR #644, opened 2025-03-26
Withdraw NAP-5 new logo proposal

jni:
There wasn't much enthusiasm for this logo. Additionally, after a bit more
time and consideration, I think "fully flat" is actually a few years behind
the 8-ball, and folks are (correctly) going back to more textured logos. I
also read a lot more from e.g. the [macOS app logo design guidelines][1],
which suggest adding a 3D tool on top of the logo to depict what the app
does. There's some ideas we could explore in this vein for napari; for
example, I often introduce napari as a tool to "explore, annotate, and
analyse" image data, so we could have a tool bundle with a magnifying glass,
a pen or pencil, and a ruler or calipers.

Either way, these directions are incompatible with the code presented in the
NAP, so I'm withdrawing it.

[1]: https://developer.apple.com/design/human-interface-guidelines/app-icons/#macOS

### DIFF
--- a/docs/naps/5-new-logo.md
+++ b/docs/naps/5-new-logo.md
@@ -5,7 +5,7 @@
 ```{eval-rst}
 :Author: Juan Nunez-Iglesias <jni@fastmail.com>
 :Created: 2022-09-08
-:Resolution: <url> (required for Accepted | Rejected | Withdrawn)
+:Resolution: https://github.com/napari/docs/pull/644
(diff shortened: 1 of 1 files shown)


willingc (approved):

DragaDoncila (approved):
Agreed we probably missed the island ferry on this one :ferry:

Czaki (approved):

PR #655, opened 2025-04-05
Add missing magicgui imports

### LINKED ISSUE #652: magicgui examples that use the @magicgui decorator don't import it
## 🧰 Task

Most of the code about the @magicgui decorator here:
https://napari.org/dev/howtos/extending/magicgui.html#magicgui-decorated-functions
doesn't have the import in the code cell.

For example, even the "complete example" here doesn't import it:
https://napari.org/dev/howtos/extending/magicgui.html#annotating-as-a-layer-subclass

The issue is that it's a potentially confusing import.
You need to do
`from magicgui import magicgui` to get the decorator.

cnstt:
Fix #652

# Description
Add missing magicgui imports in code snippets.

### DIFF
--- a/docs/howtos/extending/magicgui.md
+++ b/docs/howtos/extending/magicgui.md
@@ -410,9 +410,9 @@ from the function should be added to the viewer.  The object returned from the
 function must be an actual {class}`~napari.layers.Layer` instance.
 
 ```python
+import numpy as np
 from magicgui import magicgui
 from napari.layers import Image
-import numpy as np
 
 @magicgui
 def my_widget(ny: int=64, nx: int=64) -> Image:
--- a/docs/howtos/extending/magicgui.md
+++ b/docs/howtos/extending/magicgui.md
@@ -347,9 +347,9 @@ from {mod}`napari.types` to indicate that you only want the data attribute from
 the layer (where `<LayerType>` is one of the available layer types).
 
 ```python
+import numpy as np
 from magicgui import magicgui
 from napari.types import ImageData
-import numpy as np
 
 @magicgui
 def my_widget(array: ImageData):
--- a/docs/howtos/extending/magicgui.md
+++ b/docs/howtos/extending/magicgui.md
@@ -563,8 +563,9 @@ a `magicgui` function:
 
 ```{code-cell} python
 :tags: [remove-output]
+import napari
+import numpy as np
 from magicgui import magicgui
-import napari.types
 
 @magicgui(call_button='Make Points')
 def make_points(n_points=40) -> napari.types.LayerDataTuple:
--- a/docs/howtos/extending/magicgui.md
+++ b/docs/howtos/extending/magicgui.md
@@ -189,6 +189,7 @@ For example, the threshold widget [shown above](returning-napari-types-data)
 could be provided as a napari plugin as follows:
 
 ```python
+import napari
 from magicgui import magic_factory
 
 @magic_factory(auto_call=True, threshold={'max': 2 ** 16})
@@ -274,6 +275,7 @@ def my_widget(image: Image):
 Using `Image`  annotation in `create_widget`:
 
 ```python
+import napari
 from magicgui.widgets import Container, create_widget
 
 class ImageWidget(Container):
@@ -495,8 +497,9 @@ annotation [described above](annotating-as-napari-types-data):
 ```{code-cell} python
 :tags: [remove-output]
 import numpy as np
-from magicgui import magicgui
+import napari
 from napari.types import LabelsData, ImageData
+from magicgui import magicgui
 
 @magicgui(call_button='Run Threshold')
 def threshold(image: ImageData, threshold: int = 75) -> LabelsData:
@@ -543,6 +546,8 @@ following three forms:
 The following are all valid {attr}`napari.types.LayerDataTuple` examples:
 
 ```python
+import numpy as np
+
 # an image array
 (np.random.rand(64, 64),)
 
@@ -616,6 +621,9 @@ time the function is called:
 
 ```{code-cell} python
 :tags: [remove-output]
+import napari
+import numpy as np
+from magicgui import magicgui
 
 @magicgui(call_button='Make Points', n_points={'max': 200})
 def make_points(n_points=40) -> napari.types.LayerDataTuple:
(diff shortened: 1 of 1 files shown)


docs/howtos/extending/magicgui.md
```diff
@@ -417,6 +423,8 @@ Here's a complete example
 
 ```{code-cell} python
 :tags: [remove-output]
+from magicgui import magicgui
```
psobolewskiPhD, line 430:
```suggestion
import napari
import numpy as np
from magicgui import magicgui
from napari.layers import Image
```
lets make this one runnable since it claims to be a "complete example"

docs/howtos/extending/magicgui.md
```diff
@@ -450,6 +458,7 @@ You can create multiple layers by returning a list of
 {class}`~napari.layers.Layer`.
 
 ```python
+from magicgui import magicgui
 from typing import List
```
psobolewskiPhD:
```suggestion
```
napari in 0.6.0 is dropping python 3.9 support, so while we're here we should delete this import of List and the modify line 456/465 to use `list[napari.layers.Layer]`

docs/howtos/extending/magicgui.md
```diff
@@ -289,6 +291,7 @@ Here's a complete example:
 
 ```{code-cell} python
 :tags: [remove-output]
+from magicgui import magicgui
 import napari
 import numpy as np
 from napari.layers import Image
```
psobolewskiPhD:
```suggestion
import napari
import numpy as np
from magicgui import magicgui
from napari.layers import Image
```
This should be the proper import order

docs/howtos/extending/magicgui.md
```diff
@@ -343,6 +347,7 @@ from {mod}`napari.types` to indicate that you only want the data attribute from
 the layer (where `<LayerType>` is one of the available layer types).
 
 ```python
+from magicgui import magicgui
 from napari.types import ImageData
 import numpy as np
```
psobolewskiPhD:
```suggestion
import numpy as np
from magicgui import magicgui
from napari.types import ImageData
```
While we're here, may as well fix the import order.

docs/howtos/extending/magicgui.md
```diff
@@ -405,6 +410,7 @@ from the function should be added to the viewer.  The object returned from the
 function must be an actual {class}`~napari.layers.Layer` instance.
 
 ```python
+from magicgui import magicgui
 from napari.layers import Image
 import numpy as np
```
psobolewskiPhD:
```suggestion
import numpy as np
from magicgui import magicgui
from napari.layers import Image
```
While we're here, may as well fix the import order.

docs/howtos/extending/magicgui.md
```diff
@@ -483,6 +492,7 @@ annotation [described above](annotating-as-napari-types-data):
 
 ```{code-cell} python
 :tags: [remove-output]
+from magicgui import magicgui
```
psobolewskiPhD:
```suggestion
import napari
import numpy as np
from magicgui import magicgui
```
Lets add the missing napari and numpy import so this runs.

docs/howtos/extending/magicgui.md
```diff
@@ -548,6 +558,7 @@ a `magicgui` function:
...
 import napari.types
```
psobolewskiPhD:
```suggestion
import napari
import numpy as np
from magicgui import magicgui
```
Fix import order and add missing imports.

docs/howtos/extending/magicgui.md
```diff
@@ -577,6 +588,7 @@ You can also create multiple layers by returning a list of
 {attr}`~napari.types.LayerDataTuple`.
 
 ```python
+from magicgui import magicgui
 from typing import List
```
psobolewskiPhD:
```suggestion
```
Let's remove this, un-needed in python 3.10 and on (napari 0.6.0).
Also need to fix line 583/595 to use `list[napari.types.LayerDataTuple]`

docs/howtos/extending/magicgui.md
```diff
@@ -129,6 +129,7 @@ as soon as we add the 'camera' image.
 
 ```{code-cell} python
 :tags: [remove-output]
+from magicgui import magicgui
```
psobolewskiPhD, line 133: While we're here, could you move the `import napari` line 135 (new 136) to the top? may as well use proper import sorting.

psobolewskiPhD (changes requested):
I can't suggest, but there is one more missing `@magicgui` import, in the:
Annotating as `napari.Viewer`
section, old line 368, new line 373
Could you please get that one too?
Also there are some other minor fixes, that, since you're here, maybe you can also take care of?
Thanks!! 😊

docs/howtos/extending/magicgui.md
cnstt: `import napari.types` doesn't seem to be needed, or am I wrong?

psobolewskiPhD: Um, i'm never sure! Shows how much I rely on IDE for this stuff. 
I *think* it's needed because the return type `napari.types.LayerDataTuple` in like 554/570
Other option would be to import `from __future__ import annotations` and then quote the type hint.
I could be wrong though!

cnstt: As `napari` is imported just above, `napari.types.LayerDataTuple` should just work fine I think :)

psobolewskiPhD: Oh derp, yes you're right! Let me fix the suggestion.

cnstt:
> Also there are some other minor fixes, that, since you're here, maybe you can also take care of?

I checked the whole file again and did some minor fixes, should be better now :)

psobolewskiPhD (approved):
Thanks!

TimMonko (approved):
This code cell is the only one missing the napari import now. Thanks for doing this! A lovely change that definitely would have helped me when I was starting 😄 

https://github.com/cnstt/docs/blob/ad71fd91c525922c7f0a6285e8a0c237a436f4bf/docs/howtos/extending/magicgui.md?plain=1#L567-L583

TimMonko:
This looks good to merge now

Czaki (approved):

PR #649, opened 2025-04-01
[maint, bugfix] Ensure stub targets generated by prep_stubs generate targets

psobolewskiPhD:
# References and relevant issues
Followup to #646 

# Description
In a clean repo, `prep_stubs` was generating warnings due to missing targets. The stub files were written, but lacked targets.
I fix this in this PR and add `clean-prep` and `clean-full` to be able to ensure all autogenerated content can be cleaned if needed.

PR #671, opened 2025-04-13
[bugfix] fix prep_docs logic to ensure npe2 files are written, but not over-written

psobolewskiPhD:
Fixes logic bug in https://github.com/napari/docs/pull/649, see https://github.com/napari/docs/pull/649#discussion_r2041137415 that is causing npe2 stubs to not be generated on CI runs, see: https://github.com/napari/docs/actions/runs/14430288050/job/40464266128?pr=669#step:8:250

# Description
When testing #649 this I removed the `not` so that it would overwrite, so I could iterate more easily. 
I forgot to switch it back. make culpa.
But without the not, on a clean run, where the files don't exist, they won't be generated -- this is the case of CI.
This PR puts the `not` back.


TimMonko:
> make culpa

is this a new `make` logic you're introducing 😜

TimMonko (approved):

PR #669, opened 2025-04-12
Default CI to `slimfast` and add comment control to run other builds.

psobolewskiPhD:
# References and relevant issues
Depends on https://github.com/napari/docs/pull/671

# Description
In https://github.com/napari/docs/pull/646 new make targets were added for faster builds that skip certain parts of the process.
In this PR I make them available on CI using a comment:
`napari-bot make <target>`
(add the `@` in front of napari-bot)

Here's the breakdown:
1. circleci/config.yml : Here I add a parameter `make_target` to make the make target configurable on-run. Default is `slimfast` (no gallery, no outside stuff, no notebook execution)
2. build_docs.yml : I added this back, as a reusable workflow, so that there is a build-only workflow that can be run on PR push and triggered with different make_target. (The build-deploy job needs to use the full build for obvious reasons). I made it so that this build-only job also be manually triggered by dispatch. Whether triggered or by manual dispatch, this workflow accepts a make target, one of: html html-noplot docs slimfast slimgallery. The default is `slimfast`, so on (every) push/open of PR it will use `slimfast`, but can be triggered with the other options. Note: at present the triggered state is not saved, so push always used the default.
3. build_trigger.yml : This is the workflow that detects PR comment, checks make target vs a list, and triggers the CircleCI and build_docs builds with the make target from the comment. It also has a step to report back that the  triggered build_docs workflow finished and provide a link to the run.
4. build_and_deploy.yml : I modified this to not run on PR push anymore, because it's a full build -- we do need full build for deployment and it's triggered by napari/napari as well.

I also modified the PR template to give info on the CI builds, as well as the contributor guide where we talk about docs building on CI.

Note: I don't think the new features will trigger in this PR, because it's from a branch.
You should be able to see this in action and test it here in my fork, where this PR branch is the default branch:
https://github.com/psobolewskiPhD/napari-docs/pull/74
You can see builds triggered by comments from Juan and Carol: https://github.com/psobolewskiPhD/napari-docs/actions/workflows/build_trigger.yml

Followups: 
- have the trigger add a reaction to the comment, so the user knows it triggered
- have the trigger add a label to the PR and then use the label as default for subsequent runs (e.g. push ones)

PR #675, opened 2025-04-20
Update Makefile to use -WT --keep-going for CI make targets

psobolewskiPhD:
Depends on napari/docs#669

# Description

This adds `-WT --keep-going` to the new CI make targets, rather than full builds.
Currently, all jobs are running with no warnings.
This will cause CI builds to return a `fail` if there are warnings. By adding this to the faster builds it will allow us to catch issues before running the slow full builds.
Typical examples include missing elements from toc, problems in fenced elements, and issues with references and anchors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,14 +71,14 @@ html-noplot-live: prep-stubs
 # full build, no gallery
 # will not remove existing gallery files
 html-noplot: clean prep-docs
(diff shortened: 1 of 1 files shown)


psobolewskiPhD:
I personally find this useful because the docs build generates a lot of output, so it's easy to miss a warning which is typically important. If the exit code is a fail, then you know to look more carefully!

TimMonko (approved):
I think this is a great change. We want this new fast CI to encourage new contributors to not be afraid of the formatting, and fast build + clear warnings seems like it could accomplish that. If the warnings instead scare away people because they feel like CI is hard to overcome, then we can change the strategy, but this seems the way to go instead of getting this only on the full build.

Issue #563, opened 2025-02-05
Update the Viewer tutorial with new icons


psobolewskiPhD:
## 🧰 Task

There have been small changes to Viewer buttons, to emphasize that they can be right-clicked.
https://github.com/napari/napari/pull/7556
https://github.com/napari/napari/pull/7580
I think there is also a plan to swap the 2D/3D icons to make what happens when you click clearer.

Once this overhaul is finished, the [Viewer tutorial](https://napari.org/stable/tutorials/fundamentals/viewer.html) should be updated with the new icons and a line of copy that the tick mark means you can right-click.

psobolewskiPhD:
cc: @TimMonko 
(not saying you have to do this, but wanted to flag you for your thoughts!)

TimMonko:
I was sitting in lab meeting yesterday in lab meeting distracted by this very idea. So, I could take a swing at it as long as it can wait until after Feb 12. I'll try to have it ready a few days before 0.6.0 (btw, ETA?, milestone is Feb 11 which seems rather soon)

Also planning to figure out how to style an indicator for dim sliders / playback, coming soon. Would be nice if I could get all that in before 0.6.0 and any other UI icon changes since those seem to be going over pretty well. 

Should `viewer_pan_zoom.webm` also be updated? looks to be a fairly old napari. I'm not sure how to make webms, but I am able to make mp4s with ShareX -- anyone have advice?

I love how `nbscreenshot` works so smoothly in docs to show changes through the version switcher. :smile:

TimMonko:
See: https://github.com/TheJacksonLaboratory/activelearning/blob/main/docs/tutorials/cellpose_tutorial.qmd 
For updating viewer screenshots with drawings on them automagically

PR #691, opened 2025-04-30
Add image highlighting right-click indicators in viewer tutorial

jni:
Part of #563

While working on #690, I noticed that the right click admonition had
been added, but without an image. This PR adds an image for that
admonition, which I think makes it nicer.


jni:
New look:

CleanShot 2025-04-30 at 19 24 31@2x

Czaki (approved):

PR #696, opened 2025-05-01
Final release notes for 0.6.0

jni:
I've also updated the command palette videos in the tutorials so that
they work on firefox. (I'll update the relnotes links later since they
are unversioned and there's a lot more of them to do.)


jni:
anyone?

TimMonko (approved):

TimMonko:
If you release in the next 90 minutes then its still April 30th in Silicon Valley so you'll have to update that ;)

TimMonko:
How did you add your button presses as an overlay to the command palette video?

jni:
> its still April 30th in Silicon Valley so you'll have to update that ;)

We (naturally) use Line Islands time (where Napari is), which [also happens to be the most futuristic time zone available (UTC+14)](https://www.timeanddate.com/time/zone/@4030941) 😃 .

jni:
> How did you add your button presses as an overlay to the command palette video?

I use CleanShotX to record, but QuickTime screen recording also allows that iirc. I presume there are similar utilities available for Windows.

PR #697, opened 2025-05-01
Update the version switcher for 0.6.0

jni:
This is safe to merge because the 0.6.0 folder is already there. I'm
about to update the symlink on napari.github.io and the tag has just
been pushed. 🥳

### DIFF
--- a/docs/_static/version_switcher.json
+++ b/docs/_static/version_switcher.json
@@ -5,10 +5,15 @@
 		"url": "https://napari.org/dev/"
 	},
 	{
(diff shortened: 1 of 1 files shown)


jni:
Actually, maybe wait to merge until the tagged docs build finishes 😅 

https://github.com/napari/docs/actions/runs/14770173874

jni:
napari.org/stable is now 0.6.0! 🥳 Now just self-merging this to get the version switcher working again.

Issue #698, opened 2025-05-01
Fix opengraph preview image location


TimMonko:
## 🧰 Task

Currently, the opengraph preview is only correctly finding the image (which can be found at `./{ver}/_static/opengraph_image.png`) on some random pages... by pointing the metadata to `../dev/...`. It makes no sense to me

The image does live at both [stable](https://napari.org/stable/_static/opengraph_image.png) and [dev](https://napari.org/dev/_static/opengraph_image.png)

If we could just shove the image into `/_static/opengraph_image.png` it should work, but I'm probably doing something wrong in the setup, not sure

1. [stable/plugins points to _static](https://www.opengraph.xyz/url/https%3A%2F%2Fnapari.org%2Fstable%2Fplugins%2Findex.html)
2. [stable/community points to ../dev/_static](https://www.opengraph.xyz/url/https%3A%2F%2Fnapari.org%2Fstable%2Fcommunity%2Findex.html)
3. [dev/community points to _static though....](https://www.opengraph.xyz/url/https%3A%2F%2Fnapari.org%2Fdev%2Fcommunity%2Findex.html)
4. [napari.org/stable points to ../dev/_static ........](https://www.opengraph.xyz/url/https%3A%2F%2Fnapari.org%2Fstable)

psobolewskiPhD:
So the images is actually located here:
https://github.com/napari/napari.github.io/tree/gh-pages/dev/_static
in dev/_static
because we always deploy to dev (or a version folder):
https://github.com/napari/docs/blob/24d59394332575e45da4588b39663e25d491fde4/.github/workflows/build_and_deploy.yml#L116

All the stuff int he root of the pages repo is from before versioning and is stale (we could remove it IMO since it's redirected anyways). So to fix the stable we should just put the image in root/_static on GitHub pages. This is not "correct" but it will fix the deployed pages with minimal fuss.
And then in this repo, I think we should change the link to use /dev/_static and everything will work going forward.

I'm not sure about where the .. came from, I assume it's related to the copy of the docs.

psobolewskiPhD:
OK here is the ..
it's being used to fix unversioned pages exactly for this sort of thing:
https://github.com/napari/napari.github.io/blob/0c0ec08d48a280fd66bc6d10227a1b4a40df3124/.github/workflows/unversioned_pages.yml#L37-L42

TimMonko:
After https://github.com/napari/napari.github.io/pull/424 the preview is looking good for all instances. 

[...]

PR #699, opened 2025-05-01
Point opengraph preview location to 'stable'

TimMonko:
# References and relevant issues

Closes #698 

# Description

Currently, opengraph preview is broken as https://napari.org/_static/opengraph_image.png does not exist, but https://napari.org/stable/_static/opengraph_image.png does now after 0.6.0 release. I mixed up versioning. 
I thought it would base it off ogp_canonical_url, but I was wrong! It bases it off of ogp_site_url

Maybe this fixes all metadata? we can check previews now that the images actually live on the live website.

PR #700, opened 2025-05-01
Update conf.py to try to fix opengraph image for dev and future deployments

psobolewskiPhD:
Part of: #698 

# Description
Based on my reasoning in the issue, I think this should fix things. We only deploy to /dev or a versioned folder.
So I think the image should be in /dev and then it should be available properly.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -211,7 +211,7 @@ suppress_warnings = ["myst.header", "etoc.toctree", "config.cache"]
 # OpenGraph configuration for link previews
 
 ogp_site_url = "https://napari.org/"
-ogp_image = "_static/opengraph_image.png" 
+ogp_image = "dev/_static/opengraph_image.png" 
 ogp_use_first_image = False
 ogp_description_length = 300
 ogp_type = "website"


TimMonko:
I think this is good, but my only worry about using dev would be that if we update it while changing the UI before cutting a new release, then the preview image will look different than stable since it will be deployed right away. But realistically, that will be rare I think, since we shouldn't update this preview image very often.

psobolewskiPhD:
/stable is just a symlink, so that won't help.
So the question is what happens when a versioned deployment occurs.
because /0.6.0/_static exists:
https://github.com/napari/napari.github.io/tree/gh-pages/0.6.0/_static

I think for any of the unversioned pages copied into the versioned folder, it should work with that search/replace in the action. So yeah I think we would need a similar search and replace to have the opengraph links be versioned, because I don't think at the docs build level (when sphinx is writing the metadata header) any version numbers exist.

psobolewskiPhD:
So looking at https://github.com/napari/docs/pull/699 as the alternative, it's the same thing sort of.
/stable is a symlink, so right now it points to 0.6.0 and but in 0.6.1 it will point to 0.6.1 which will mean that any docs deployed with /stable/_static will have the same graphic. So right now the two options we have on the table are :
- this PR which will make things share the `dev` graphic
- #699 which will make things share the last released graphic

TimMonko (approved):

TimMonko:
My final verdict on this is that it will best serve us for maintenance to use /dev/_static as you've proposed here. Thanks!

PR #702, opened 2025-05-02
Update sidebar-nav-bs.html to try to fix links

### LINKED ISSUE #682: Resources links (side bar) is broken when accessing from pages other than in the root
## 🐛 Bug

For example, the link to the Release Notes in the side bar on the left is broken when accessed from the Tutorials, e.g. this state:
Image

It points to `https://napari.org/dev/tutorials/fundamentals/release/index.html` which is wrong. This might just be due to the dev vs stable docs?

This feature was introduced in https://github.com/napari/docs/pull/636

It appears to break when accessed from any page that is not in the root
So if you click `Usage` in the nav bar, then the Resources links will work. But if you click on the `Quick start` tile, the links will be broken. Same if you use the Community or Plugins nav bar items.

psobolewskiPhD:
Closes: https://github.com/napari/docs/issues/682
hopefully!

# Description
Based off of: https://www.sphinx-doc.org/en/master/development/html_themes/templating.html#helper-functions
Trying to use pathto() to fix the links.

### DIFF
--- a/docs/_templates/sidebar-nav-bs.html
+++ b/docs/_templates/sidebar-nav-bs.html
@@ -19,19 +19,19 @@
     </p>
     <ul class="nav bd-sidenav">
       <li class="toctree-l1">
-          <a class="reference internal" href="community/meeting_schedule.html">Community calendar</a>
+          <a class="reference internal" href="{{ pathto(community/meeting_schedule.html, 1) }}">Community calendar</a>
       </li>
       <li class="toctree-l1">
           <a class="reference external" href="https://napari.zulipchat.com/">Community chat</a>
       </li>
       <li class="toctree-l1">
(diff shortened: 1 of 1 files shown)


psobolewskiPhD:
Looks like it worked, at least for dev docs.
Not sure yet how to fix deployed 0.6.0

psobolewskiPhD:
I guess we could try to merge this and then re-deploy to 0.6.0 folder using manual trigger?

TimMonko:
You know more about how to correctly deploy this, but I have tested it and it does fix the sidebar links. Wonder how we let this slip through unnoticed.

Pretty neat that there are default Jinja functions in Sphinx

TimMonko (approved):

DragaDoncila:
Oh lit, I was just about to open an issue for this :laughing: I'ma just merge instead

PR #715, opened 2025-05-21
Update contributing docs page

melissawm:
# References and relevant issues
--

# Description
This is a suggestion only, and I took a heavy axe to this document 😅 But I just wanted to experiment with a different organization of things. I tried:
- Highlighting the common themes to all workflows first
- Including more signposting
- Using dropdowns for the specialized build - I think that section will need a more careful look tomorrow
- Grouping similar themes together so the information is less scattered.
This can obviously be very subjective so I'm definitely happy to address feedback and keep working on it. In particular I'll come back tomorrow to review the links and other internal references since they might have changed.

The main content is still all there, I just moved things around and maybe reworded things here and there.

PR #716, opened 2025-05-22
Update code of conduct committee members

### LINKED ISSUE #661: Code of Conduct notes
First, I want to say I've always been grateful for repos that have a Code of Conduct, it really helps to create a welcoming environment, even if it seems like a small thing. 

1. [Reporting Guidelines](https://napari.org/stable/community/code_of_conduct.html#reporting-guidelines) are the people to report to up to date? I know Nick and Kira aren't as active any more. Also, it may be helpful to state why Charlotte Weaver is an alternative contact (if they still are).

jni:
Closes #661

CC @DragaDoncila @psobolewskiPhD @TimMonko @willingc

### DIFF
--- a/docs/community/code_of_conduct.md
+++ b/docs/community/code_of_conduct.md
@@ -54,12 +54,13 @@ For clearly intentional breaches, report those to the Code of Conduct committee 
 You can report issues to the napari Code of Conduct committee, at napari-conduct@googlegroups.com . Currently, the committee consists of:
 
 * [Juan Nunez-Iglesias](https://github.com/jni) (chair)
-* [Nicholas Sofroniew](https://github.com/sofroniewn)
-* [Kira Evans](https://github.com/kne42)
+* [Draga Doncila Pop](https://github.com/DragaDoncila)
+* [Peter Sobolewski](https://github.com/psobolewskiPhD)
+* [Tim Monko](https://github.com/TimMonko)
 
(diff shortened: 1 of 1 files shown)


TimMonko (approved):

psobolewskiPhD:
Huh, Circle ran normally here, but not Melissa's PR https://github.com/napari/docs/pull/715

psobolewskiPhD:
It struck me that GitHub handles don't provide a means of communication (probably for the best, but still).
So if someone wants to contact the committee and it involves one of the members, there is no way for them to do so without using the email that would be expected to hit all the members?
I guess that's where the external person comes it who can be reached directly/single-ly

psobolewskiPhD (approved):

melissawm (approved):
I also agree with Peter that maybe we want to be better at giving contact information, but maybe the general email is enough. Not sure to be honest. In any case, thanks to all new members stepping up 🙏🏻

TimMonko:
Yes, I thought about that general contact too. But, I'm wondering is our ease of access on Zulip enough? Perhaps thats worth a mention? Certainly I don't mind if my e-mail gets directly added! I think part of the idea of there being a committee-type e-mail is that you can trust that reaching out should get you _some_ response from someone.

melissawm:
In general, communicating with the committee I think is better than communicating privately. That's because communications with the committee will naturally leave a paper trail which is good for transparency and accountability. However, I know for coc communications it can be hard to trust a faceless/nameless email. 

I think having the members named and maybe mention somewhere that they are on zulip can be ok, but stressing that official complaints must be sent to the committee email. (Sorry if this is confusing)

PR #720, opened 2025-05-24
Update docker build links

willingc:
Currently the general container registry link redirects to files from 2022. This PR updates to the package location of builds for napari and napri-xpra.

### DIFF
--- a/docs/howtos/docker.md
+++ b/docs/howtos/docker.md
@@ -4,7 +4,9 @@
 
 ## Build
 
-Builds are available in the [GitHub Container Registry](https://github.com/orgs/napari/packages).
+Builds are available in the GitHub Container Registry:
+- [napari](https://github.com/napari/napari/pkgs/container/napari)
+- [napari-xpra](https://github.com/napari/napari/pkgs/container/napari-xpra)
 
 A dockerfile is added to napari root to allow build of a docker image using official napari release.
 It contains two targets built on top of Ubuntu 20.04:


psobolewskiPhD:
Lol, i think the published date is when we started publishing packages. 🤯 
If you click the links there, it goes to the same place.

psobolewskiPhD (approved):

docs/howtos/docker.md
```diff
@@ -4,7 +4,9 @@
 
 ## Build
 
-Builds are available in the [GitHub Container Registry](https://github.com/orgs/napari/packages).
+Builds are available in the GitHub Container Registry:
+- [napari](https://github.com/napari/napari/pkgs/container/napari)
+- [napari-xpra](https://github.com/napari/napari/pkgs/container/napari-xpra)
 
 A dockerfile is added to napari root to allow build of a docker image using official napari release.
 It contains two targets built on top of Ubuntu 20.04:
```
psobolewskiPhD, line 12:
```suggestion
It contains two targets built on top of the official Python 3.11 slim-bookworm image:
```
I'm not sure this is important. Maybe we just delete it so it's not something that needs to be maintained?

Similarly, below it should say:
* `napari`: The result of `python -m pip install pip install "napari[all]", including all the ...

willingc: Yeah, I can go ahead and review the rest of the doc. I just put the PR together when I saw the links :wink:

psobolewskiPhD: Updating this is on my todo list 👀 (along with updating the xpra container), so I'd not worry about the whole document. But I just noticed those two lines as being out-dated. We could just remove them.

PR #723, opened 2025-05-28
Fix comment and manual dispatch triggered build jobs

psobolewskiPhD:
Currently, comments are not properly triggering jobs.
The primary  issue is that comment triggers are from the main branch context and not PR context.
For circleCI it's even harder, because if you checkout the PR and then run CircleCI it will still internal checkout the wrong branch.
Also reporting back to the PR was also not working.
I did some googling and sorted out a pathway using [CircleCI-Public/trigger-circleci-pipeline-action
](https://github.com/CircleCI-Public/trigger-circleci-pipeline-action) -- not sure how I missed it before?
Also consulted Claude a bit.

# Description
I ended up re-doing the whole action, so I gave it a new name. I think the diff would be atrocious, so this is easier.
The main change here are as follows. First for the Circle part:
1. To use https://github.com/CircleCI-Public/trigger-circleci-pipeline-action in the action, we need to adjust the circleci/config.yaml to add the pipeline parameters required by the action and also switch to using a job parameter for the make target. Then we can use workflows for circleCI that are conditional on the GH_Meta pipeline parameter. The net result is that the right make target is used.
2. Use https://github.com/CircleCI-Public/trigger-circleci-pipeline-action which lets us pass the PR branch (this isn't well documented, but works) via target-branch, as well as the desired make target in GH_Meta. The best part is the status reporting then works for both the Circle job and the redirector :tada:  
We get the target-branch from the "PR details" step of the first job `determine-make-target`  

Importantly, this also sets us up for a followup PR to check for a label and run full build. With this implementation that will be trivial by comparison.

Now for the artifact build:
1. The key is the job is again getting those PR details. Then we use the right branch for the reusable workflow and get the right artifact built. For this we add the extra workflow_call parameter so we can pass the PR detail to the job.
2. The issue is with the status reporting. Again the context being the main branch of the repo, getting the reusable action status is tricky. myrotvorets/set-commit-status-action ended up being very useful for this simplifying a ton vs. the old approach.
Fun part:
- Added the eyes reaction to show that the comment was received.

I made the PR branch my main branch in my fork and then opened a PR vs that, which you can find here:
https://github.com/psobolewskiPhD/napari-docs/pull/90
You can see that things came together and work!
Feel free to try a comment!


.circleci/config.yml
```diff
@@ -6,21 +6,34 @@
...
+  GHA_Meta:
+    type: string
+    default: ""
+
 # Orbs are reusable packages of CircleCI configuration that you may share across projects.
 # See: https://circleci.com/docs/2.1/orb-intro/
 orbs:
   python: circleci/python@3.0.0
 
-# parameterize the make target used for build, passed in from build_trigger.yml action
```
psobolewskiPhD, line 14: With the change to use https://github.com/CircleCI-Public/trigger-circleci-pipeline-action we will be triggering a workflow (below) and then passing the make target as a job parameter, rather than pipeline parameter. So here we remove this.

.circleci/config.yml
```diff
@@ -57,6 +70,21 @@ jobs:
           paths:
             - docs/docs/_build/html/
 workflows:
+  # default trigger on push, don't double up with actions
   build-docs:
+    when:
+      not: << pipeline.parameters.GHA_Meta >>
```
psobolewskiPhD, line 86: This change is needed to prevent double-triggers. We want the normal workflow to trigger on push/commit/PR status only, not when the comment/trigger is used.

.circleci/config.yml
```diff
@@ -57,6 +70,21 @@ jobs:
...
     jobs:
       - build-docs
+  # triggered by comment to PR with make target
+  triggered-by-issue-comment:
```
psobolewskiPhD: So now for the triggered job, we use a new workflow that checks the GHA_Meta pipeline parameter and kicks off our build-docs job with the right job parameter.

.github/workflows/build_docs.yml
```diff
@@ -5,19 +5,23 @@ name: Build PR Docs
...
-        default: 'slimfast'
+        default: "slimfast"
   workflow_call:
     inputs:
       make_target:
         description: "Enter make target: html html-noplot docs slimfast slimgallery"
         type: string
-        default: 'slimfast'
+        default: "slimfast"
+      pr_ref:
```
psobolewskiPhD, line 21: Add another input parameter so that we can pass the PR detail from the triggering job.

.github/workflows/build_docs.yml
```diff
@@ -33,6 +37,7 @@ jobs:
         with:
           # Check out to  '/home/runner/work/docs/docs/docs'
           path: docs
+          ref: ${{ inputs.pr_ref || github.ref }}
```
psobolewskiPhD, line 40: Here make sure to be able to use the branch info passed in from the triggering job or just the regular dispatch.

.github/workflows/build_docs.yml
```diff
@@ -70,7 +75,7 @@ jobs:
           GOOGLE_CALENDAR_API_KEY: ${{ secrets.GOOGLE_CALENDAR_API_KEY }}
           PIP_CONSTRAINT: ${{ github.workspace }}/napari/resources/constraints/constraints_py3.10_docs.txt
         with:
-          run:  make -C docs ${{ github.event_name == 'pull_request' && 'slimfast' || inputs.make_target }}
+          run: make -C docs ${{ inputs.make_target || 'slimfast' }}
```
psobolewskiPhD, line 78: Make sure the trigger works for both PR triggers and manual dispatch (both use inputs.make_target)

.github/workflows/triggered_target_build.yml
```diff
@@ -0,0 +1,117 @@
...
+  pull-requests: write
+  statuses: write
+
+jobs:
+  determine-make-target:
+    runs-on: ubuntu-latest
+    outputs:
+      target: ${{ steps.determine-make-target.outputs.target }}
+      pr_ref: ${{ steps.get-pr-info.outputs.pr_ref }}
+      pr_sha: ${{ steps.get-pr-info.outputs.pr_sha }}
```
psobolewskiPhD, line 31: These are the 3 key things we need to know about the trigger:
- what make target
- pr details, so we can get the right stuff to build and then report back to the right status.

.github/workflows/triggered_target_build.yml
```diff
@@ -0,0 +1,117 @@
...
+
+  trigger-circleci:
+    needs: determine-make-target
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run CircleCI pipeline
+        uses: CircleCI-Public/trigger-circleci-pipeline-action@v1.2.0
+        with:
+          GHA_Meta: ${{ needs.determine-make-target.outputs.target }}
+          target-branch: ${{ needs.determine-make-target.outputs.pr_ref || github.ref_name }}
```
psobolewskiPhD, line 95: Here we pass in the make target from the earlier job and then ensure that status reporting works by also passing in the branch information. If a manual dispatch is used, it will also work (GitHub.ref_name).

.github/workflows/triggered_target_build.yml
```diff
@@ -0,0 +1,117 @@
...
+        env:
+          CCI_TOKEN: ${{ secrets.CIRCLECI_TOKEN }}
+
+  trigger-artifact-build:
+    needs: determine-make-target
+    uses: ./.github/workflows/build_docs.yml
+    with:
+      make_target: ${{ needs.determine-make-target.outputs.target }}
+      pr_ref: ${{ needs.determine-make-target.outputs.pr_ref || github.ref_name }}
```
psobolewskiPhD, line 104: Again, this is the branch info we need to build the right stuff.

.github/workflows/triggered_target_build.yml
```diff
@@ -0,0 +1,117 @@
...
+    steps:
+      - name: Set build job status
+        uses: myrotvorets/set-commit-status-action@master
+        with:
+          sha: ${{ needs.determine-make-target.outputs.pr_sha }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+          status: ${{ needs.trigger-artifact-build.result }}
+          context: "Docs Artifact Build: ${{ needs.determine-make-target.outputs.target }}"
+          description: ${{ needs.trigger-artifact-build.result == 'success' && 'Documentation built successfully' || 'Documentation build failed' }}
+          targetUrl: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
```
psobolewskiPhD, line 121: Here's the reporting. We have the right PR info, so then we collect the pass/fail status and the make target so we can put that in the PR status.

TimMonko (dismissed):

TimMonko:
@psobolewskiPhD 
Great work! Your github code commenting is beautiful. Exemplified exactly the value we talked about at the last docs meeting! 
Can we test in this branch? Or only after merge? I think if I'm reading workflow file, bot should work on this PR?

TimMonko:
@napari-bot make slimfast

TimMonko:
My comment above did trigger an action, but does not add the eyes at the time I expected, which would be at the start of the action.

TimMonko:
Ah, not quite fully hooked up correctly yet:
https://github.com/napari/docs/actions/runs/15327362311?pr=723+%28comment%29

psobolewskiPhD:
@TimMonko this repo (napari/docs) is using the actions from the default branch (main), so your comment is using the not-correct original implementation.
If you want to test, as I noted in the PR description, I've made *this PR branch* the default branch of my fork--equivalent to merging it into main here in the docs repo. You can open a PR in my fork against the the `fix_comment_trigger` default branch to test. I've opened a PR against that branch in my fork and there you can test this, which i did put in the OP: https://github.com/psobolewskiPhD/napari-docs/pull/90

(it was these branch complications that led my original attempt astray, as can be seen by the needed changes)

TimMonko (approved):

TimMonko:
Somehow I didn't put it all together reading your (excellent) PR description. Thank you for the clarification! 
Most recently we've been setting up action chagnes to work within the given PR, so I naively thought the same would work here. I guess the reference to main makes this a bit more challenging. 

Thanks!

psobolewskiPhD:
> Ah, not quite fully hooked up correctly yet: [napari/docs/actions/runs/15327362311?pr=723+%28comment%29](https://github.com/napari/docs/actions/runs/15327362311?pr=723+%28comment%29)

If you click to this run, on the left you can click `workflow file` and see that it's the workflow from `main`

psobolewskiPhD:
> Most recently we've been setting up action chagnes to work within the given PR, so I naively thought the same would work here.

That can work if the action is triggered by PR push for example. But in this case, the triggers are not from the branch of the PR, but from an outside context (comment, label, etc.), which lives in the default branch of the repo.

TimMonko:
Thank you for the extra clarification! Really helping me to learn. The very reason workflow ran from main was why I was concerned, but I now realize how this 'outside context' makes this the issue, and why it is a challenge to troubleshoot this type of CI.

.github/workflows/triggered_target_build.yml
```diff
@@ -0,0 +1,119 @@
+name: Trigger target build of docs
+# this workflow is triggered by a comment on a PR or manually via workflow_dispatch
+# It allows you to specify a make target for building the docs for a branch/PR
+# When triggered by a comment, it will run on the PR branch and report status on the PR
+# # This is a one-time-run, it does not run on every push to the PR branch
```
willingc:
```suggestion
# This workflow can be started either by commenting on a pull request or by using workflow_dispatch manually.
# You can specify a `make` target to build the documentation for a specific branch or pull request.
# If started by a comment, the workflow runs on the PR branch and updates the PR status.
# Note: This workflow runs only once per trigger and does not execute on every push to the PR branch.
```

.circleci/config.yml
```diff
@@ -6,21 +6,34 @@
...
-parameters:
-  make_target:
-    type: string
-    default: "slimfast"
-
 jobs:
   build-docs:
     docker:
       - image: cimg/python:3.10.17
+    parameters: # job level setting of make target
```
willingc:
```suggestion
    parameters:
      # Set make target for the job
```

.circleci/config.yml
```diff
@@ -57,6 +70,21 @@ jobs:
           paths:
             - docs/docs/_build/html/
 workflows:
+  # default trigger on push, don't double up with actions
   build-docs:
+    when:
+      not: << pipeline.parameters.GHA_Meta >>
     jobs:
       - build-docs
+  # triggered by comment to PR with make target
```
willingc:
```suggestion
  # Run this workflow when a PR comment triggers a docs build for `make` target
```

.circleci/config.yml
willingc:
```suggestion
  triggered-by-pr-comment:
```

This is probably better renamed to PR.

.circleci/config.yml
```diff
@@ -57,6 +70,21 @@ jobs:
           paths:
             - docs/docs/_build/html/
 workflows:
+  # default trigger on push, don't double up with actions
```
willingc: This comment is confusing. It's unclear what "don't double up with actions" means.

.github/workflows/build_docs.yml
willingc, line 21: `pr_ref` or `pr_branch`?

.github/workflows/build_docs.yml
willingc: adding branch to `pr_ref` name would improve readability and clarity (or renaming to `pr_branch`)

willingc (approved):
Looks like this will work fine. I would consider rewording some of the comments and changing some variable names.

.circleci/config.yml
psobolewskiPhD: Thanks, I will try to rephrase.

.github/workflows/build_docs.yml
psobolewskiPhD: um, good question. let me try to look that up again. I think it's correct, but it's been a while since I looked at this.

.github/workflows/build_docs.yml
psobolewskiPhD: If i recall `ref` is the GitHub terminology. I will try to circle around to this and see if it can be changed.

.github/workflows/build_docs.yml
psobolewskiPhD: Ok, so this is analogous to the `github.ref` that would be normally used, which is using the `git` terminology. So i lean to keeping this the same, because it will make it easier for anyone diving into this, but keeping things consistent?

.github/workflows/build_docs.yml
psobolewskiPhD: Yeah, as noted in the other spot, `ref` is actually git terminology and it is also used by GitHub. So we either use the `github.ref` if using a workflow dispatch or the `pr...ref` which we have stored in the `pr_ref` variable. So i lean to leaving this to make it easier for someone searching/reading github/git docs to sort this out?

.circleci/config.yml
```diff
@@ -71,8 +71,12 @@ jobs:
           paths:
             - docs/docs/_build/html/
 workflows:
-  # default trigger on push, don't double up with actions
   build-docs:
+    # this workflow is triggered automatically by default
```
willingc, line 80: OK, I see what is confusing now. This wording is fine now. I do think that we should add to the top of this file the following:

```
# This GitHub Action workflow uses CircleCI-Public/trigger-circleci-pipeline-action.
# This allows triggering a CircleCI pipeline from any event on GitHub with GitHub Actions.
# See: https://github.com/CircleCI-Public/trigger-circleci-pipeline-action/blob/main/README.md
```

willingc (approved):
Add this suggested comment to top of file and :ship: Thanks @psobolewskiPhD

PR #726, opened 2025-06-01
Update installation.md to link to conda getting started not miniconda

psobolewskiPhD:
The current `conda` link on the installation page links to miniconda docs, which go to anaconda.com
anaconda.com can be a blocked domain due to licensing issues

# Description
Instead of linking to the miniconda docs specifically, link to the conda "getting started guide"
https://docs.conda.io/projects/conda/en/stable/user-guide/getting-started.html


jni (approved):

jni:
I'm a big fan of micromamba so I wouldn't mind pointing to that instead. But we can have that discussion later, this is already an improvement. 👍

TimMonko (approved):
Its ok by me. It does reference trying to get one of the minimal installers before anaconda, so that's a good thing.

PR #730, opened 2025-06-10
Updates to NAP-9: Multiple Views

brisvag:
Current (before this PR) rendered version of this NAP: https://napari.org/stable/naps/9-multiple-canvases.html

You can click on the "Check the rendered docs here!" below to see the rendered version of this update!

# Description
It's time to pick up this work again! A lot of the changes should hopefully be self-explanatory, but there are a few general things to note:

1. Naming change from "canvas" to "view": there were several discussions about this, which I tried to summarize in the `Alternatives > Terminology` section. Feel free to discuss and propose more!
2. There's now a lot more reliance on the Async work. When the NAP was first drafted, we pinpointed the potential overlap and mentioned it, but by now I think that work has advanced enough to see that we should simply finish it and rely on it.
3. Switching from a single shared layerlist (as proposed by @aganders3 in his original draft in #249) to a per-view layerlist: this was my main disagreement on that version of the implementation, and I tried to highlight some of the reasons in the `Alternatives > Single LayerList` section. I'd like particular attention/discussion on this part so we can be more comprehensive on the pros/cons and reach a more consensus decision.
4. I removed the architecture diagrams (sorry @aganders3 :sweat_smile:): they were outdated, and I found they didn't help me much with following the design logic due to the high level of detail and unrelated methods/attributes being mentioned. If you disagree and think they should be updated and added back in, let me know.

As usual for NAPs (as per [the NAP workflow](https://napari.org/stable/naps/0-nap-process.html#nap-workflow)), remember that this is just another iteration in the discussion, so this PR will be merged relatively liberally, after enough discussion and changes have accumulated. Declaring the NAP as `Accepted` is not part of this PR.


psobolewskiPhD:
I would strongly prefer anything other than `View`. Viewbox, viewport, canvas, ...

TimMonko:
> I would strongly prefer anything other than `View`. Viewbox, viewport, canvas, ...

Why is this @psobolewskiPhD? I think Lorenzo's layout of the reasoning to choose 'view' strongly corresponds with why I prefer view as well. So, I'd like to hear what you have against it.

docs/naps/9-multiple-canvases.md
```diff
@@ -10,27 +9,34 @@
...
+*Viewer* - Currently maps basically 1:1 to the napari application main window, including canvas, dims sliders, layer list, layer controls, and dock widgets. Related is the `ViewerModel`, a class in napari that maintains the state of the Viewer.
 
-*Canvas* - The main napari data view (2D rectangle) where 2D or 3D slice data is displayed, and additional data views displayed within a single Viewer window. This NAP is in part an attempt to formaize this concept, and as noted proposes a related `CanvasModel` class in napari to hold the state of a Canvas.
+*Canvas* - Currently usually used to refer to the central widget of the napari Viewer which renders the data from the current slice, as well as keeping track of all the visualisation scenegraph from vispy and various other napari-vispy interfaces.
 
-*Layer* - The base unit of the napari image data model. A `ViewerModel` maintains an ordered list of Layers that it may display on its Canvas.
+*View* - New term introduced by this NAP in order to formalize and disentangle the previous notions of *Canvas* and *Viewer*. This NAP introduces a `View` class in napari with its own `Layerlist`, `Dims`, and `Camera`.
```
TimMonko, line 22: Would this also introduce a `view.canvas`? or, do we want canvas elements (overlays, color, etc) to live directly on the view object (and by extent, remain on `viewer`)? I could see both implementations being sensible

docs/naps/9-multiple-canvases.md
```diff
@@ -43,210 +49,291 @@ Several plugins and examples have been created to address these limitations, for
...
 
 Providing native support in napari would allow developers to more easily create these experiences, enable interoperability between such plugins, and improve performance.
 
 ### Out of Scope
 * Improvements to VisPy to support multiple views of the same `SceneGraph` (sharing data, saving VRAM) - for relevant discussion start with [vispy/#1992](https://github.com/vispy/vispy/issues/1992).
-* For now, canvas arrangement (for example: tiling behavior) will be handled in the view only (left to Qt or custom Qt widgets). Making this state (de)serializable is out of scope for this project, but may be relevant when implementing a “savable viewer state” feature.
-* Normalizing slice data for different layer types, though this may benefit in the course of this work.
-* Specific UI implementations will be explored as part of this work, but I expect UX and UI will be formalized later (possibly in a separate NAP).
+* For now, view arrangement (for example: tiling behavior) will be handled in the viewer only (left to Qt or custom Qt widgets). Making this state (de)serializable is out of scope for this project, but may be relevant when implementing a “savable viewer state” feature.
```
TimMonko, line 67: Can you clarify two things:
1. which tiling behavior is this referring to? In that I could rearrange the various `view`s? 
2. what would it even mean for the views to be serializable or not? isn't serialization _generally_ out of scope for our current model?

docs/naps/9-multiple-canvases.md
```diff
@@ -43,210 +49,291 @@ Several plugins and examples have been created to address these limitations, for
...
+    status: Union[str, dict]
+    tooltip: Tooltip
+    theme: str
+    title: str
+    _overlays: EventedDict[str, Overlay]
+    _layer_slicer: _LayerSlicer
+```
 
-Fig. 1: napari architecture today.
+With this NAP and the above considerations, we expect to separate the above components roughly as follows:
```
TimMonko, line 155: This section is great, really helps me understand 🍻

TimMonko:
> Switching from a single shared layerlist (as proposed by aganders3 in his original draft in https://github.com/napari/docs/pull/249) to a per-view layerlist: this was my main disagreement on that version of the implementation, and I tried to highlight some of the reasons in the Alternatives > Single LayerList section. I'd like particular attention/discussion on this part so we can be more comprehensive on the pros/cons and reach a more consensus decision.

@brisvag your description of why not to do single layerlist has mostly convinced me, I think. Though there is perhaps some hybrid method that I haven't seen described, and would happy to be corrected. 
In this hybrid method there would be a full, single layerlist that has a tickbox (or equivalent) appear next to each layer corresponding to each view. Then, if the view is ticked, that layer is also moved/copied into the layerlist for that view. 

For me, toggling and dragging layers _only_ between views would be irritating. I think it also makes sense that if I adjust contrast on a single layer in the main layer list, then that layer would be updated the same in each view, but then if I only update view1's contrast of that layer, the other views are left alone.

docs/naps/9-multiple-canvases.md
```diff
@@ -43,210 +49,291 @@ Several plugins and examples have been created to address these limitations, for
...
+* [3D Slicer](https://www.slicer.org/)
+* [Orthogonal views in ImageJ and Imaris](https://www.youtube.com/watch?v=94d8sHMP_w8)
+* [OHIF/Cornerstone.js](https://www.cornerstonejs.org/live-examples/crosshairs)
+* [neuroglancer](https://neuroglancer-demo.appspot.com/#!%7B%22dimensions%22:%7B%22x%22:%5B8e-9%2C%22m%22%5D%2C%22y%22:%5B8e-9%2C%22m%22%5D%2C%22z%22:%5B8e-9%2C%22m%22%5D%7D%2C%22position%22:%5B2914.500732421875%2C3088.243408203125%2C4045%5D%2C%22crossSectionScale%22:3.762185354999915%2C%22projectionOrientation%22:%5B0.31435418128967285%2C0.8142172694206238%2C0.4843378961086273%2C-0.06040274351835251%5D%2C%22projectionScale%22:4593.980956070107%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%22precomputed://gs://neuroglancer-public-data/flyem_fib-25/image%22%2C%22tab%22:%22source%22%2C%22name%22:%22image%22%7D%2C%7B%22type%22:%22segmentation%22%2C%22source%22:%22precomputed://gs://neuroglancer-public-data/flyem_fib-25/ground_truth%22%2C%22tab%22:%22source%22%2C%22segments%22:%5B%2221894%22%2C%2222060%22%2C%22158571%22%2C%2224436%22%2C%222515%22%5D%2C%22name%22:%22ground-truth%22%7D%5D%2C%22showSlices%22:false%2C%22layout%22:%224panel%22%7D)
```
jni, line 119: I would add ITK-SNAP here, which defaults to orthogonal views + 3D view.

docs/naps/9-multiple-canvases.md
```diff
@@ -10,27 +9,34 @@
...
 
-Note that [VisPy](https://vispy.org/) (the current backend for all napari visuals) has its own specific definitions for some of these or related concepts, such as [`Canvas`](https://vispy.org/api/vispy.app.canvas.html#module-vispy.app.canvas) and [`ViewBox`](https://vispy.org/api/vispy.scene.widgets.viewbox.html#module-vispy.scene.widgets.viewbox). Where necessary to refer to these concepts in this NAP (or discussion), such concepts will be qualified accordingly (for example: "a VisPy Canvas").
+Note that [VisPy](https://vispy.org/) (the only current rendering backend in napari) has its own specific definitions for some of these or related concepts, such as [`Canvas`](https://vispy.org/api/vispy.app.canvas.html#module-vispy.app.canvas) and [`ViewBox`](https://vispy.org/api/vispy.scene.widgets.viewbox.html#module-vispy.scene.widgets.viewbox). Where necessary to refer to these concepts in this NAP (or discussion), such concepts will be qualified accordingly (for example: "a VisPy Canvas").
```
jni: It would be good here to have definitions and a distinction between these elements.

docs/naps/9-multiple-canvases.md
```diff
@@ -10,27 +9,34 @@
...
 
 ## Abstract
+Current napari architecture supports a single canvas/camera/view per viewer. Simultaneously showing multiple views of the same (or different) data generally necessitates opening an entirely new napari viewer window or low-level work with Qt widgets and private napari APIs. This wastes resources (primarily memory) and complicates interaction.
+This NAP establishes a plan to implement a builtin system for opening, visualizing and interacting with multiple views within a single viewer.
 
-Current napari architecture supports a single canvas (viewbox) per viewer (window). Simultaneously showing multiple views of the same data generally necessitates opening an entirely new napari viewer window or low-level work with Qt widgets and private napari APIs. This wastes resources (primarily memory) and complicates interaction.
+We propose to achieve this in two parts that can be implemented independently:
+1. Splitting of part of the current `ViewerModel` into a new `View` model holding a `Layerlist`, a `Dims`, and a `Camera`. `ViewerModel` will hold a list of `Views`, allowing for multiple independent views.
```
jni: Suggestion: replace "list" with "Sequence" (or "list/Sequence") to leave room for more complex containers that could include layout information, such as a 2x2 array that can be viewed as a ravelled array.

docs/naps/9-multiple-canvases.md
```diff
@@ -43,210 +49,291 @@ Several plugins and examples have been created to address these limitations, for
...
-* Normalizing slice data for different layer types, though this may benefit in the course of this work.
-* Specific UI implementations will be explored as part of this work, but I expect UX and UI will be formalized later (possibly in a separate NAP).
+* For now, view arrangement (for example: tiling behavior) will be handled in the viewer only (left to Qt or custom Qt widgets). Making this state (de)serializable is out of scope for this project, but may be relevant when implementing a “savable viewer state” feature.
+* Normalizing slice data for different layer types to a uniform protocol/system, though this may benefit in the course of this work.
+* Specific UI implementations will be explored as part of this work, but UX and UI will likely be formalized later. Since this NAP was first drafted, a lot of work and discussion a on UX and UI has already been carried on in [#5348](https://github.com/napari/napari/issues/5348), which may be in
```
jni: which may be in... ... ?

docs/naps/9-multiple-canvases.md
```diff
@@ -43,210 +49,291 @@ Several plugins and examples have been created to address these limitations, for
...
-    * Layer visibility
+    * Separation of rendering information (e.g: colormap) from the data (e.g: features) which currently both live on the layer model
+    * Window state restoration and "workspaces" (see [#4227](https://github.com/napari/napari/issues/4227) for further discussion)
+
+## Requirements
+* The application data model (`ViewerModel` + Layers) shall support multiple views.
+* The application shall natively display multiple views simultaneously.
+    * There shall be a minimum of one view (current status) per viewer.
+* Each view shall have independent:
+    * Layer list (necessary for visualizing different data)
```
jni: Can you expand on this use case (visualizing different data)? It's unclear to me that this is a *requirement*.

To clarify my comment: at the community meeting we discussed this approach vs "views" of a single shared layer list, and (iirc) we found that if we could express things both as subsets of a single layer list, or as separate layer list, *as long as* layers could be in more than one place at once. At this point, it becomes an engineering choice rather than a requirement.

However, if you have compelling use cases that completely independent lists can achieve better, then it makes sense to expand on them here.

docs/naps/9-multiple-canvases.md
```diff
@@ -43,210 +49,291 @@ Several plugins and examples have been created to address these limitations, for
...
-napari architecture is based on the MVC pattern. The model layer comprises a `ViewerModel` and a list of Layer models (subclasses of a base `Layer`). There are seven layer types, each with a corresponding view type. Currently models and views are paired 1:1, and the correlation is stored in a map (`layer_to_visual`) on the `VispyCanvas`. Figure 1 shows the class relationships for the base model types and the Image layer types (for brevity - other layer types have similar connectivity).
+## Implementation
 
-```{figure} _static/multicanvas-napari-architecture-today.png
-:name: nap-9-fig-1
+(part-1-view-model)=
+### Part 1: View model
+
+* Introduce a minimally disruptive `_views` attribute on the ViewerModel, and implement a `View` model to hold a `Layerlist`, a `Dims`, and a `Camera` (as well as some related concepts and models).
+* using a `SelectableEventedList` for `_views` lets us easily introduce the concept of an `active` View. This, together with some simple dispatch of current viewer elements to the active view, will make this part of the implementation a drop-in replacement.
```
jni, line 128: I am nodding vigorously reading this section. 😃

docs/naps/9-multiple-canvases.md
brisvag: Not in my mind. I think everything that so far was under the "canvas" umbrella can go directly in View (bgcolor, size, canvas overlays). I don't see the need for adding an extra nesting level.

docs/naps/9-multiple-canvases.md
brisvag: 1. yes
2. I wouldn't say viewer serialization is "generally" out of scope: it's a pretty big and old standing goal of napari. The reason why this line is here is to address some expecations that arised during discussion in NAP-3 and later conversations, where some folks wanted this work to allow saving a "viewer workspace setup" so it could be restored.

docs/naps/9-multiple-canvases.md
```diff
@@ -43,210 +49,291 @@ Several plugins and examples have been created to address these limitations, for
...
 
-Fig. 1: napari architecture today.
+With this NAP and the above considerations, we expect to separate the above components roughly as follows:
+
+```py
+class View:
+    camera: Camera
+    cursor: Cursor
+    dims: Dims
+    grid: GridCanvas
```
jni, line 162: I find this weird, because I thought that GridCanvas was the first implementation of multi(canvas/view). That is, the way I envisioned things instead is that grid remains in the ViewerModel and triggers the creation of many `View`s with linked `Camera`s. What am I missing?

jni: I guess you could compose orthoviews with grid mode. But that's getting very hairy very quickly... (But certainly now I'm curious to see it 😂)

docs/naps/9-multiple-canvases.md
```diff
@@ -43,210 +49,291 @@ Several plugins and examples have been created to address these limitations, for
...
-* `ViewerModel` will own a *list* of `CanvasModel` objects
-* `QtViewer` will own a *list* of `VispyCanvas` objects
-* `VispyLayer` subclasses will hold references to their `Layer` (for rendering information) as well as a `LayerSlice` (for data)
-* The LayerSlicer will need to update the sync and async callbacks[^async-only] to pass a canvas parameter where the resulting sliced data will be stored, rather than storing it on the layer itself. Slice task cancellation logic will need to be revisited accordingly.
-    * async callback: `LayerSlicer._on_slice_done`
-    * sync callback: `Layer._slice_dims`
-* Callbacks (interaction, events) will need to be specifically connected to individual `CanvasModel` objects where relevant (dims, camera) rather than the `ViewerModel`.
+:::{admonition} TODO
+Discuss whether the LayerSlicer should be singleton on the viewer or one per view.
+:::
```
jni, line 184: gut feeling is: one per view, with some form of linking. Orthogonal views become hard to reason about in a single object I think... A collection of SubSlicers? 🤔

docs/naps/9-multiple-canvases.md
```diff
@@ -43,210 +49,291 @@ Several plugins and examples have been created to address these limitations, for
...
+
+:::{seealso}
+An alternative implementation could maintain a single centralized `LayerList`, with `Views` only having control over layer visibility. See [#Alternative single LayerList](#single-layerlist) for pros and cons.
+:::
```
jni, line 188: This admonition probably belongs with the requirements, or wherever you first mention the independent layer lists.

brisvag:
> In this hybrid method there would be a full, single layerlist that has a tickbox (or equivalent) appear next to each layer corresponding to each view. Then, if the view is ticked, that layer is also moved/copied into the layerlist for that view.

I don't think this would be really different from @aganders3's proposal. The only difference is in the GUI, but implementation-wise it's still a single centralized layelist with `Views` only knowing whether a certain layer is visualised there or not.

> I think it also makes sense that if I adjust contrast on a single layer in the main layer list, then that layer would be updated the same in each view, but then if I only update view1's contrast of that layer, the other views are left alone.

I didn't consider something like this until now, but I can see the uselfulness of something like this! However, I'm pretty sure we're opening the can of worms of "separating data from visualisation", as mentioned in the [non-goals](https://output.circle-artifacts.com/output/job/b0ea6d36-26e1-4c91-98b5-f3f438a970dd/artifacts/0/docs/docs/_build/html/naps/9-multiple-canvases.html#out-of-scope).
Not to say this isn't worth doing, just that we might have to expand this NAP even more into a few sub-naps :P

docs/naps/9-multiple-canvases.md
brisvag: lol, something got lost there :P

docs/naps/9-multiple-canvases.md
brisvag: Yes, my point here was not meant to be about the single layerlist vs multiple layerlists. Rather about the necessity of *one of the two*. I'll phrase is differently.

docs/naps/9-multiple-canvases.md
brisvag: This has been a sticking point many times over... I guess I didn't explain myself very well about this. I added new sections describing why I think this are and should be separate. See the new `### Grid mode vs Multi-View` in design considerations and `Multi-View grid mode` in Alternatives.

brisvag: @psobolewskiPhD This should hopefully also answer [this comment of yours](https://github.com/napari/napari/pull/7870#issuecomment-2973965660).

docs/naps/9-multiple-canvases.md
brisvag: I agree, i think multiple slicers are easier both to reason about and implement. The only reason I brought this up was in case there are performance considerations (async, threading, etc).

docs/naps/9-multiple-canvases.md
brisvag: I added a link in that first mention to clarify (as mentioned in oyour other comment), so I think this more extensive one can live here!

docs/naps/9-multiple-canvases.md
TimMonko: For me grid mode should remain distinct in that it's, by it's implementation, relatively inflexible. This gives a more foolproof multi-layer gridded viewing experience. Sure, it uses viewboxes, but does not require its own view because of sharing dim and camera states (which, to me, is one of the largest differences about multi-view in general)

docs/naps/9-multiple-canvases.md
TimMonko: Juan brings up another way to look at my discussion about the 'hybrid' design. I.e., why is multiple lists _required_ for anything? 
The one thing you bring up in the recent commits (I think) is that a single list would require de-duplication of the camera and dims and such for each view. But, why can't this single layer list store only a reference to the data itself, and not the state. Then each view specifically handle the views and state of that data. Perhaps this goes beyond my understanding in how async and lazy data would be handled (since all the data would not live in memory at once). I think with in-memory data this seems entirely reasonable, and good design to keep separate state and data, but can't say I would fully grasp why this might not work

docs/naps/9-multiple-canvases.md
TimMonko: My impression was that creating a canvas object would help us untangle Qt and VisPy. By adding some of the VisPy things (like color and overlays) to Viewer/View we are mixing Qt and VisPy. If we create a canvas object we can start to separate these, right?

brisvag: I think the previous discussions about the "Canvas model" should apply to the `View` model here. The same is true here: by having explicit model-based API for things like "background" and "overlays" we remove the necessity of digging through vispy/qt-specific code from the napari/python side, and it's instead the callbacks who need to do the work, which allows us to detach the two parts. Detaching specifically Qt *from* vispy was never really part of this discussion (though still important, but harder because our vispy-side code in `Canvas` assumes everything happens in QT).

docs/naps/9-multiple-canvases.md
brisvag: I'm sorry but I don't really follow your explanation here :sweat_smile: 

> require de-duplication of the camera and dims and such for each view.

What do you mean with de-duplication, and where do I talk about this?

> But, why can't this single layer list store only a reference to the data itself, and not the state

What do you mean by data, and by state? If you mean `Layer` and `SliceResponse`, then yes: `Then each view specifically handle the views and state of that data.`, and this was the proposal from @aganders3. I don't think this is really a big difference between the two versions, ultimately they behave almost identically to the user (though this version might be *slightly* more unintuitive to implement because there's no 1-1 mapping, but it's not a big deal :P)

> and good design to keep separate state and data

Again, assuming I understand what you mean above: this is true for both implementations as well!

psobolewskiPhD:
@napari-bot make slimgallery

brisvag:
There's a few unanaswered comments above, but we could also merge this already and continue iterating in a later PR with.

TimMonko:
> There's a few unanaswered comments above, but we could also merge this already and continue iterating in a later PR with.

If you could pop those into a new issue, I'd be happy to merge as is. (I think) the whole point of NAPs is to update them as needed, as we've done here.

TimMonko (approved):

docs/naps/9-multiple-canvases.md
brisvag: @TimMonko it was mostly this one that I'm not sure I addressed!

PR #735, opened 2025-06-18
Add active roadmap document

DragaDoncila:

# Description
This PR adds our active roadmap to the website under the `Roadmaps` section and links to it from the `Roadmaps` index page.

### DIFF
--- a/docs/roadmaps/active_roadmap.md
+++ b/docs/roadmaps/active_roadmap.md
@@ -1,52 +1,8 @@
 # **Napari Roadmap**
 
-## April 10, 2025
+## June 22, 2025
 
-[**Project vision and foundation	1**](#project-vision-and-foundation)
-
-[Vision	2](#vision)
-
-[Founding principles	2](#founding-principles)
-
-[Mission and purpose	2](#mission-and-purpose)
-
-[**Roadmap strategy	3**](#roadmap-strategy)
-
-[Approach and philosophy	3](#approach-and-philosophy)
-
-[Potential challenges	3](#potential-challenges)
-
-[Risk mitigation strategies	4](#risk-mitigation-strategies)
-
-[**Roadmap focus areas	4**](#roadmap-focus-areas)
-
-[Core technology	4](#core-technology)
-
-[User experience	5](#user-experience)
-
-[Project sustainability and community	5](#project-sustainability-and-community)
-
-[**Strategic deliverables	5**](#strategic-deliverables)
-
-[Core technology	5](#core-technology-1)
-
-[API evolution	6](#api-evolution)
-
-[Core library	6](#core-library)
-
-[Performance	7](#performance)
-
-[User experience	7](#user-experience-1)
-
-[Project sustainability and community	8](#project-sustainability-and-community-1)
-
-[**Timeline	8**](#timeline)
-
-[2025	8](#2025)
-
-[2026 milestones and beyond	9](#2026-milestones-and-beyond)
-
-# **Project vision and foundation** {#project-vision-and-foundation}
+# **Project vision and foundation**
 
 [napari](https://napari.org/) is a fast, interactive, multi-dimensional image viewer for Python. It's designed for browsing, annotating, analyzing, and processing large multi-dimensional images.
 
@@ -70,7 +26,7 @@ The napari team came together around a shared vision for a Python-based image vi
 
 * provides native 2D and 3D views.
 
-## Mission and purpose {#mission-and-purpose}
+## Mission and purpose
 
 napari aims to be the multi-dimensional image viewer for Python and to provide GUI access to a plugin ecosystem of image analysis tools for scientists to use in their daily work. We hope to accomplish this by:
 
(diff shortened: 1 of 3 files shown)


docs/roadmaps/active_roadmap.md
```diff
@@ -0,0 +1,310 @@
...
+
+* native (no browser required),
+
+* blazing fast (GPU-powered),
+
+* layered, allowing easy visualisation of data in a common space, and
+
+* provides native 2D and 3D views.
+
+## Mission and purpose {#mission-and-purpose}
```
DragaDoncila: would be nice to just include this section somehow from the actual docs page so we're not maintaining two spots?

docs/roadmaps/index.md
```diff
@@ -2,7 +2,11 @@
 
 # Roadmaps
 
-The [**napari global roadmap**](https://github.com/orgs/napari/projects/24/views/2?pane=info) is hosted on a Github Project board. This roadmap captures the current priorities of the napari core developer team. It is revisited every six months.
+The [**napari global roadmap**](active_roadmap.md) is a living document, and we revisit it every three months. It contains a summary of our vision, our mission and values, and our key focus areas
+for the project. It also details strategic deliverables and key priorities for each focus area within a horizon of 3-5 years.
+
+As a companion to the roadmap, we manage a [Github Project board](https://github.com/orgs/napari/projects/24/views/2?pane=info) that contains more technical details, issues and Pull-Requests
```
DragaDoncila: do we want to just remove the bit about the project board? or do we want to commit to keeping it updated?

willingc: Remove it. It can always be added later if needed.

docs/roadmaps/active_roadmap.md
```diff
@@ -0,0 +1,266 @@
...
+
+### Project sustainability and community
+
+* **Bug Fixes:** Addressing reported issues and ensuring software stability.
+
+* **Documentation:** Maintaining comprehensive and up-to-date documentation.
+
+* **Website:** Providing a central hub for information and resources.
+
+* **Community Engagement:** Fostering communication, outreach, and collaboration through conferences and core team meetings.
```
melissawm, line 111: I think this is a strong point of napari and its community, and we could potentially highlight the approach to community engagement as part of the project's mission. One of the best things about a community-led project (and what give it an edge over proprietary systems, in general) is that engagement and a committment to listening to feedback from the community. I think napari does it well and we should highlight it.

docs/roadmaps/active_roadmap.md
```diff
@@ -0,0 +1,266 @@
...
+* is n-dimensional first,
+
+* native (no browser required),
+
+* blazing fast (GPU-powered),
+
+* layered, allowing easy visualisation of data in a common space, and
+
+* provides native 2D and 3D views.
+
```
brisvag, line 28: Isn't extensibility worthy of living here?

In general, i notice now that I'm nto sure what 's the difference between "foundign principles" and "mission and purpose".

docs/roadmaps/active_roadmap.md
DragaDoncila: Yeah I think it would be good to update these sections on the main Mission & Values page and then just include it here somehow? This is currently lifted from those sections, and I would hate to have to maintain it in two spots. Do we think it's better to tweak this PR to automagically include the current version from docs and then I can follow up with a PR updating that section? Or do we think we should make the desired updates here and then copy across? I think it's important they stay in sync. @jni @kephale @kevinyamauchi thoughts?

docs/roadmaps/active_roadmap.md
DragaDoncila: Great point @brisvag. See [my comment above](https://github.com/napari/docs/pull/735#discussion_r2161249548) about maybe just including the current copy here from the original page and then updating there?

docs/roadmaps/active_roadmap.md
willingc: @DragaDoncila You may be able to use a `literalinclude` with line numbers using Sphinx/myst

willingc:
@DragaDoncila and all, Let's focus on landing this document as is. It's already way better than what we have today. It is possible in future PRs that we may be able to use a `literalinclude` to pull from other pages.

While I don't disagree with @melissawm and @brisvag on their points, I think the changes can come in a future PR (either soon, in 3 months, or when using the `literalinclude`).

Draga, I also think it is **very** important to add the Roadmap to the list of resources in the left-hand sidebar.

docs/roadmaps/active_roadmap.md
willingc: See my other comment re: merging as is for now.

TimMonko (approved):
LGTM. Yeah agree with Carol this is WAY better than what we have. Sidebar would be stellar to show that we are working on being rockstars!

willingc (approved):

willingc:
Added #741 so next steps do not fall through the cracks.

docs/roadmaps/active_roadmap.md
jni: Just catching up here. 😅 I think the founding principles are important historical context, ("why is napari not in Jupyter!!!") and I don't think that plugins were really a part of it. However, they are of course very much a part of napari now. So I think it's worth expressing this as "founding principles" followed by, "napari today and in the future" or somesuch — things that were added later but are now part of the mission.

PR #736, opened 2025-06-19
Use darker blue for community meetings in napari calendar

TimMonko:
# References and relevant issues

Inspired by a comment from Draga in a core-dev thread on Zulip: 
> I think it's totally fine to have non-community-meeting events in the calendar, I'd love to see it. Would be nice to give the community meetings their own colour though, maybe? Not sure if that's possible on the website. Just wanna make sure they're easy to spot

# Description

Based on Ashley's suggestion (https://github.com/napari/napari-sphinx-theme/pull/189#issuecomment-2985525867) and coinciding with my own searches. This was the alternate strategy I found to what I proposed in #189, but I was going to overly complicate things in comparison...

Overrides calendar color with `--napari-primary-blue` if the events contain "community" or "working" in the title, since these are consistent names we use. Therefore, these events will be emphasized compared to the (current) lighter color that any other event will have. 

image

PR #738, opened 2025-06-20
Test passing PR number instead of ref on triggered build

aganders3:
See https://github.com/napari/docs/pull/736#issuecomment-2989423609

# Description
Testing a hunch that @actions/checkout can't find the ref because it's looking for it in _this_ repo, but it belongs to the fork the PR is coming from.

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -18,8 +18,8 @@ on:
         description: "Enter make target: html html-noplot docs slimfast slimgallery"
         type: string
         default: "slimfast"
-      pr_ref:
-        description: "PR branch reference"
+      pr_number:
+        description: "PR number"
         type: string
         required: false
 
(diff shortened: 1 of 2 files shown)


aganders3:
@napari-bot make slimfast

psobolewskiPhD:
Yes! I was looking at the same thing!
Forks man...can't live with em can't live without em.

aganders3:
Well, that didn't work, but also it looks like the bot uses the workflow file(s) from `main` so it wouldn't...

psobolewskiPhD:
let's see if dispatch works, cause I think it should also be affected.

Edit: just letting me choose main or tags :(

psobolewskiPhD (approved):

psobolewskiPhD:
I'm going to merge to see if we get this fixed. I think this is the correct solution.

psobolewskiPhD:
It won't get any more broken 😜

aganders3:
I think it's working? https://github.com/napari/docs/actions/runs/15769048532

I guess now I'm a little confused how this is different from a normal "build PR docs" that runs on PR sync.

Edit: ahh, it lets you specify a more full docs build where the default is `slimfast`

psobolewskiPhD:
Yeah, it's complicated 😦 
I think i have a fix for the circle tho https://github.com/napari/docs/pull/739

PR #745, opened 2025-06-26
[Enh] When `ready-to-merge` is applied, do a full build of docs

psobolewskiPhD:
Followup to https://github.com/napari/docs/pull/723 and https://github.com/napari/docs/pull/669
This came up in a meeting and zulip, plus maybe a PR reviews?

# Description
Adds an action to check for ready-to-merge label and trigger a full build when tagged.
This is simpler than the comment action because it should use the PR context by default.
fingers crossed.
Note it's just a one-time full build. Pushes will still use the default, slimfast.

I did testing in my fork:
https://github.com/psobolewskiPhD/napari-docs/pull/89
But as we saw before isn't guaranteed to work here! So we may need to merge optimistically and then fix if it doesn't trigger correctly. 😬

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@
 version: 2.1
 
 # in addition to the default circleCI triggering, we have a GitHub Action:
-# .github/triggered_target_build.yml 
+# .github/triggered_target_build.yml
 # This Action has a job that uses CircleCI-Public/trigger-circleci-pipeline-action.
 # This allows triggering a CircleCI pipeline from any event on GitHub with GitHub Actions.
 # Right now we are using comments on PRs to trigger the workflow "triggered-by-pr-comment"
@@ -98,3 +98,9 @@ workflows:
     jobs:
       - build-docs:
           make_target: << pipeline.parameters.GHA_Meta >>
+  triggered-by-label:
+    when:
+      equal: ["ready-to-merge", << pipeline.parameters.GHA_Meta >>]
+    jobs:
+      - build-docs:
+          make_target: "html"
--- a/.github/workflows/label_triggered_build.yml
+++ b/.github/workflows/label_triggered_build.yml
@@ -1,4 +1,4 @@
-name: Trigger Full Build on ready-to-merge label
+name: Trigger Full Build on ready to merge label
 
 on:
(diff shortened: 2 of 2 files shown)


psobolewskiPhD:
whoa, since it's from the branch context, it actually triggers here?
https://github.com/napari/docs/actions/runs/15915678461/job/44892779998
Not sure why getting a 403 from circle. I will look into it. But otherwise it seems to be working!

psobolewskiPhD:
It's like the token is missing?
https://github.com/napari/docs/actions/runs/15915678461/job/44892779998#step:2:5

Oh! because it's running from my branch, which doesn't have access to the secrets.
So once this is merged I think it will all work.

TimMonko:
> So when the tag is applied, it kicks off a full build. Progress is updated in the 
`ci/circleci: build-docs` status thingie. And when the job finishes, the redirector works properly.
If you then push to the branch, this will trigger the default build.
I think this is OK? Changing the default behavior doesn't seem obvious, because the push is the trigger, so not obvious how to set the parameter.
The idea behind the read-to-merge label full run is that review applies the label and then a full build kicks off just make sure everything is fine. IF CI green, then you merge. if not, then you end up removing ready-to-merge anyways.

_Originally posted by @psobolewskiPhD in https://github.com/psobolewskiPhD/napari-docs/issues/89#issuecomment-2910459319_
            
Agreed!

TimMonko (approved):

TimMonko:
@psobolewskiPhD Triggered, but failed :O

image

psobolewskiPhD:
@TimMonko See https://github.com/napari/docs/pull/745#issuecomment-3011141679
Basically this is still not merged so running from my fork, which doesn't have the right token.
Im actually surprised it's working at all before being merged, hence 403 error:
https://github.com/napari/docs/actions/runs/15917574006/job/44897991205?pr=745#step:2:19

melissawm (approved):
I love it!!! And also agree that we can't really debug now, so I think we should merge and address remaining issues later.

TimMonko:
failed in #774 @psobolewskiPhD 

https://github.com/napari/docs/actions/runs/15928380029/job/44931156980?pr=744

PR #749, opened 2025-06-28
Only build artifact on ready-to-merge label

psobolewskiPhD:
Alternative to https://github.com/napari/docs/pull/748

# Description
To avoid using the less secure pull_request_target trigger, we can just avoid building the CircleCi preview.
The idea is that when reviewing, you would do the bot trigger to build the needed preview. The ready-to-merge build would just be there to verify that a full build doesn't fail.
Note: because a reusable action is used, it will be triggered from the fork, but that should be ok as long as the fork branch is up to date with napari/docs `main`, which is easy to ensure with the update button in PRs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,9 +98,3 @@ workflows:
     jobs:
       - build-docs:
           make_target: << pipeline.parameters.GHA_Meta >>
-  triggered-by-label:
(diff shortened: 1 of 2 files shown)


TimMonko (approved):
Approving in favor over #748

PR #751, opened 2025-06-29
Final 0.6.2 Release Notes

TimMonko:
# Description

Final release notes for 0.6.2
You can also read my ramblings on trying and failing to get a video in release notes, before realizing on Github the notes need a link and not something relative to the napari-docs repo.


TimMonko:
I've done goofed both the video and image aren't showing
I'll circle back later maybe someone will know the problem

jni:
See [this change](https://github.com/napari/docs/pull/696/files#diff-ebc7a3be656d701d432bc948d19514f77f8e6a9d441074608eeab588c27dce56L103-R114) in #696:

1. You have to use the same link that is in the markdown source of the comment, rather than right click->copy link. It starts with `github.com/user-attachments` rather than `private-user-images.githubusercontent.com`. They are, for some weird github technical reason, completely different links, and work differently when linked in markdown. But also:
2. If you do that, only some browsers will render it correctly. Firefox will not. Therefore, you should instead download the videos, convert them to webm with ffmpeg, add them to `_static/images`, and use the `video` html5 tag, as in #696. Here is my ffmpeg command for converting mp4 to webm (replace `mov=` with your own file name):
  `mov="command-palette"; ffmpeg -i ${mov}.mp4 -c:v libvpx-vp9 -crf 40 -b:v 0 -c:a libopus ${mov}.webm`

TimMonko:
Ahh! I was wondering why the image link was so odd. 
I recently converted the video to webm for #725, so it's good to know I can just link to things in _static hopefulyl the build looks good now

TimMonko:
@jni what have I goofed.. it works in the viewer tutorial but not in the release notes
tutorial: https://output.circle-artifacts.com/output/job/24154297-0d1e-4015-ab7e-7797a1257876/artifacts/0/docs/docs/_build/html/tutorials/fundamentals/viewer.html
release notes: https://output.circle-artifacts.com/output/job/24154297-0d1e-4015-ab7e-7797a1257876/artifacts/0/docs/docs/_build/html/release/release_0_6_2.html

TimMonko:
oh I see, mayhaps its because we've never actually linked any videos from release notes, only gifs. the video think you linked is actually from non-release note files.

Czaki:
I have opened PR with troubleshooting of mixed installation. You may want to add it to this release.

Czaki (approved):

jni (approved):
You can definitely do video links from release notes, they'll just need a different relative path. I think we should do that because gifs are a huge waste of bandwidth for a less-smooth playback. But we can fix that in later PRs, in the meantime, let's get this in for release! 🚀

PR #755, opened 2025-07-01
Update Release Guide

TimMonko:
# Description

Just some changes after managing the last two releases, while fresh in my mind.

PR #757, opened 2025-07-01
Add empty release notes for 0.6.3

Czaki:

# Description

To avoid problems with lack of release notes on first napari 0.6.3 alpha just add empty release notes.


jni (approved):

TimMonko:
@jni @Czaki do you think it would be better to instead encourage Release Managers to make not empty release notes, but run the script on the next milestone? 

Thinking ahead to what we can add to #755 ...

Thinking about it more, it's better that release notes don't include speculative PRs, because it would make something like an early alpha confusing when reading the notes because of uncertainty on what's actually been changed.

Czaki:
> @jni @Czaki do you think it would be better to instead encourage Release Managers to make not empty release notes, but run the script on the next milestone?

Since at least 3 releases, the release manager forgot to do this and the first alpha was not released. 



> Thinking about it more, it's better that release notes don't include speculative PRs

I agree

PR #764, opened 2025-07-03
Update triggered_target_build.yml regex to ensure we match on hyphen

### LINKED ISSUE #763: Fix the bot make trigger workflow to have `html-noplot` work properly

psobolewskiPhD:
Closes https://github.com/napari/docs/issues/763

# Description
Fixes the regex to include `-` so we can match `html-noplot`


TimMonko (approved):
Thanks!

PR #766, opened 2025-07-08
Explore homepage reorg to bring cards further up

DragaDoncila:

# Description
In discussion with @TimMonko we noted that the homepage cards are very far down the page and some folks don't even know they're there - on Tim's machine it looks like the homepage just ends.

This PR explores moving the video and the text into a two-column layout so we can bring the cards further up the page, above the fold.


DragaDoncila:
This is janky on mobile, it's just a POC

DragaDoncila:
Closing for #767

PR #780, opened 2025-07-13
Update napari.org homepage to remove the `imshow` "button"

psobolewskiPhD:
At SciPy 2025, Harry, a UI specialist, asked about the non-functional button on the homepage. We were confused and realize he was referring to the `imshow` command that is in a box that renders like a button:
image

The idea is that this is the code you can run to get the visual, but it doesn't look like a code command (left justified with prompt) and the code doesn't actually run either.

As a result, it was suggested to remove this and recover the real-estate.

# Description

This PR removes the `figcaption` that adds a box with the `imshow` command under the video animation.


psobolewskiPhD:
Not sure why Circle isn't running. I assume it's using my account or something again

DragaDoncila (approved):
I totally agree the code block doesn't provide any meaningful information, not to mention the issues with formatting and how much space it takes up. I'm for removing it.

psobolewskiPhD:
image

DragaDoncila:
So much more room for activities :heart_eyes:

jni:
Hi folks,

just as a bit of context, that code block was there to emphasise that napari is usable within Python, rather than (just) a standalone application, because several folks at SciPy over the years have been confused thinking that napari is a standalone app and thus not for them. Having the "code" below the video was inspired my the matplotlib home page, which I love:

matplotlib.org

screenshot_fDqKi3Dd@2x

Each refresh yields a different example with the relevant function call. I'd love to get to something like that, even if the existing home page design didn't really achieve that.

DragaDoncila:
@jni maybe let's open an issue about how to achieve this?

I think as a minimum, in order for it to be useful in exhibiting the python-ness of napari we'd need it to be clearly formatted as code and linking to a basic example of getting napari going with an image.

psobolewskiPhD:
Yeah the matplotlib page doesn't render it as a button and the text is a link to docs, so then it makes much more sense!

Aside: it also looks like a carousel or something because it's different each time, which is also something that folks at SciPy 2025 suggested would be valuable. I have notes, but not yet written up in human readable format.

TimMonko:
It could be a win-win to have highlighted examples, and then have those displayed on a home page carousel. Then linking the example to it will bring people to the delightful code!

Issue #785, opened 2025-07-17
Documentation on napari Events Reference


ian-coccimiglio:
## 🐛 Bug
I've been involved in learning about the napari Event Loop over SciPy 2025, and have made the [Napari Event Monitor](https://github.com/ian-coccimiglio/napari-event-monitor) plugin to help with both interactive diagnostics and documentation of napari events. As such, here are some of the things I'm interested in improving.

The [napari Events Reference](https://napari.org/stable/guides/events_reference.html) should be:
1) complete for all non-private events
- See issue for text overlays https://github.com/napari/napari/issues/5599
- All interactive elements in napari should have a corresponding event reference, like the 'display text' button in points layer.
2) have no repeated entries. 
- See issue https://github.com/napari/napari/issues/7221
3) have associated descriptions for existing events 
- This might mean updating the docstrings

I'm posting this issue to document areas for improvement as well as provide progress updates. A related aim is to expand the narrative/explanations of how the napari events system works, targeted towards intermediate napari users.

willingc:
I'm working on a pr that would deduplicate the repeated items and add a column for which files emit the event.

PR #787, opened 2025-07-18
Add module docstrings to scripts

willingc:
Partially addresses #785 

# Description

I'm not sure how I feel about this PR since AI was used. The module docstrings were generated by Claude Code, which I have been testing for another project. I'm opening the PR since it may help @ian-coccimiglio and others understand our scripts which generate the events page and other auto-generated content.

I'm fine if we wish to close this and never merge into the codebase. In the short-term, I do think it gives useful information.

I have added an attribution statement to each docstring as well.


TimMonko:
To me, this is a prime use case of AI for code maintenance, and is generally how I use it myself (figuring out what's going on in a file without having to parse the code). So, the only reason I would not have to support this is that sometimes module docstrings go out of date and we don't uupdate them well enough. I'm in favor of how you've done this.

TimMonko (approved):

PR #788, opened 2025-07-18
Update pre-commit config to add some python checkers

willingc:
This PR adds some additional pre-commit checkers for our `.py` files.

@jaimergp I know you are working through automation stuff so adding you here as well.


TimMonko (approved):
I laughed reading #787 seeing all the things pre-commit would have fixed, get fixed

jaimergp (approved):

PR #789, opened 2025-07-20
Group event docs in order

willingc:

Partially addresses #785.

# Description

This PR groups event related docs to be in order so that they are easier to discover.

No content is changed in this PR.

FYI @ian-coccimiglio


TimMonko (approved):

TimMonko:
Looks good to me, makes sense. Thanks! 
image

PR #795, opened 2025-07-22
Minor roadmap update

jni:
At the SC we had a minor discussion about whether to use "multidimensional
gridded data" vs "multidimensional image data", and in the end #735 went with
just "multidimensional data" which I think is too general — we are not and
never will be about interacting with high-dimensional tabular data, for
example.

Therefore, I add back "image" as our primary focus. (I think those that
understand the word "gridded" also understand that image straighforwardly
generalises to gridded?)

I also removed "experts" from "Python experts" since we also very much target
Python newbies. Could also do "*from* Python experts to non-Python users"
instead of "*both*" — thoughts welcome.

Finally, the file had a bunch of trailing whitespace that was causing vim to
yell at me so I removed it. 😜

### DIFF
--- a/docs/roadmaps/active_roadmap.md
+++ b/docs/roadmaps/active_roadmap.md
@@ -51,9 +51,9 @@ Our documentation provides more information about [napari's mission and values](
 
 ### Approach and philosophy
 
-**This roadmap is a living document** that the team adapts as needed to meet changing priorities, technology shifts, and scientific opportunities. napari's roadmap focuses on expanded core functionality, performance improvements, new features, and compatibility with other scientific tools. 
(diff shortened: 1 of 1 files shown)


willingc (approved):
Definitely fine as suggested. At the risk of bike-shedding, perhaps: 

Napari aims to simplify the exploration, annotation, and analysis of multidimensional image data, ensuring that the experience is smooth for everyone, whether they are experts in Python or not. This inclusive approach allows a broader audience to engage with complex image data effectively.

psobolewskiPhD (approved):
Love this.
thanks for the thoughtfulness @jni

jni:
@willingc thanks for the suggestion! I like it but after some thought, I think I prefer to keep it one-sentence statement. We can of course continue to iterate, but in the meantime (and with thanks to @psobolewskiPhD for chiming in), I'll merge this in time for 0.6.3 (though I hope this doc is in the unversioned part of the docs, no time to check rn 😅)

PR #797, opened 2025-07-22
Add introductory paragraph to starting an event loop api doc

willingc:

Partially addresses #785

# Description

This adds a brief paragraph giving more context on `napari.run` for starting the event loop.

cc/ @ian-coccimiglio

### DIFF
--- a/docs/api/event_loop.rst
+++ b/docs/api/event_loop.rst
@@ -1,6 +1,10 @@
 Starting the Event Loop
 -----------------------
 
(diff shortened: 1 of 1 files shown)


willingc:
image

TimMonko:
Very useful. Thanks! 
Maybe in the future we change the name of it to just be `napari.run()` because it seems to make it harder for SEO being named the way it is -- and would be more useful for people searching for the function.

TimMonko (approved):

PR #799, opened 2025-07-23
Update title of event loop api page for discoverability of napari.run

willingc:
Update the page title per Tim's suggestion.

Follow-up to #797. Related to #785.


TimMonko (approved):
:heart:

PR #807, opened 2025-07-31
Update announcement and version switcher

### LINKED ISSUE #801: When switched to old docs, there is no banner to suggest stable
## 🐛 Bug
Go to:
https://napari.org/0.5.4/index.html

See no banner at the top for link to stable
(it works for dev docs)

willingc:

Closes #801

# Description

This PR makes the following changes:
- updates announcement url to use github content
- fix a trailing comma in the version_switcher.json that local build was erroring on
- fix an invalid pygments import statement
- updates app-model intersphinx urls to https

# Testing

I wasn't sure a good way to test the version switcher with PRs. Checked the pydata sphinx theme repo but their CI seems to also redirect to live versions.


TimMonko:
Has 2 banner headers, which I thought we turned off at one point. But I might not be used to it, given that I guess its been quietly broken for a while?
image

docs/_static/announcement.html
```diff
@@ -1,6 +1,6 @@
 <!-- This will appear in the announcement banner at the top of the site.
 It needs to be activated by passing the link:
-"https://napari.org/dev/_static/announcement.html"
+"https://raw.githubusercontent.com/napari/docs/main/docs/_templates/announcement.html",
```
TimMonko, line 3: Can you explain to me this change? I don't understand why its different

in part because I don't see an announcment html in the _templates

willingc: Ugh. Because I pulled the was moving the announcement.html into _templates where it belongs since it is currently in `_static`. And then I changed my mind. I will update the comment.

willingc: Changed my mind again and am moving announcements.html into _templates to be consistent with pydata sphinx themes approach

willingc:
@TimMonko Check out the pydata sphinx theme docs for examples of two banners. We could style the colors with css.

TimMonko:
Thanks! Excited to have the banner back. 
 It trips me up so often to see I'm on like 0.4.18

TimMonko (approved):

melissawm:
I'm sorry I was too slow to review 😅 - I am undoing some of this PR in #811 since (I think) we don't want the announcement banner. We can also think about ways to document these decisions in the docs working group if it helps. Thanks!

PR #811, opened 2025-08-01
Remove announcement banner

melissawm:
Related to #807 and #801

# Description

Deactivates the announcement banner and moves the template back to _static.

The reason the templates needs to live in _static is that we want to be able to do updates to the banner without rebuilding the full docs site (see https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/announcements.html#update-or-remove-announcement-banner)

Also remove redundant logic for the version switcher.

### DIFF
--- a/docs/_static/announcement.html
+++ b/docs/_static/announcement.html
@@ -1,8 +1,11 @@
 <!-- This will appear in the announcement banner at the top of the site.
 It needs to be activated by passing the link:
-"https://raw.githubusercontent.com/napari/docs/main/docs/_templates/announcement.html",
+"https://napari.org/dev/_static/announcement.html",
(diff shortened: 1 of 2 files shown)


willingc:
@melissawm Curious about what the downside is for doing a full docs rebuild when changing the announcement?

willingc (approved):

psobolewskiPhD:
@willingc I think this PR the the "source of truth" https://github.com/napari/docs/pull/341

The idea is that all the banners on all of the versions that were built with the flag will read from the same source that they can all access and is simple to edit. Then those old versions do not need to be rebuilt for a specific file.

psobolewskiPhD (approved):

PR #813, opened 2025-08-02
Reorganize bundle instructions page to make it easier to navigate and provide download links

### LINKED ISSUE #714: Improve the bundle app download experience
## 🧰 Task

The main installation guide has a section for the bundled app:
https://napari.org/stable/tutorials/fundamentals/installation.html#install-as-a-bundled-app
There is also a more detailed guide:
https://napari.org/stable/tutorials/fundamentals/installation_bundle_conda.html

But neither of these provides the actual download link directly. They make you navigate to the Releases page, which is really annoying and confusing if there has been a pre-release recently.

We should provide direct links or tiles to the specific version of napari matching the docs version.
Note that we *do* this in the Quick Start:
https://napari.org/stable/tutorials/fundamentals/quick_start.html#installation

psobolewskiPhD:
Closes: https://github.com/napari/docs/issues/714
Note: this is a reorganization borne out of a review by Harry Barng at SciPy 2025

# Description
This PR makes it simpler to navigate the bundle installation:
1. adds tiles to jump to the instructions for a chosen OS
2. adds direct links to the proper GitHub download for the version. Note I tried to use tiles for this, but the version substitution didn't work. I'm open to other suggestions for how to do it. E.g. it might be possible to use an image as the click target. After discussing with Harry we decided that best was the enemy of the good and this was still worth implementing.

I also made some copy edits, mostly to just standardize on "app" for this, since bundle isn't at all obvious to the target audience, while everyone is familiar with apps.


psobolewskiPhD:
Huge thanks to @harrybarng (didn't ping in the opening post to reduce spam).
You can see the updated page built from this PR here:
https://output.circle-artifacts.com/output/job/39cb1ea6-f5fd-4e1e-bc00-04c4615885b7/artifacts/0/docs/docs/_build/html/tutorials/fundamentals/installation_bundle_conda.html

Note: we still need to improve the Installation landing page, but at the docs meeting we decided that this PR would be rather benign and uncontroversial, while changes to the Installation page might be more radical.

willingc (approved):
I love these changes @psobolewskiPhD. Nice work and collaboration at SciPy.

jni:
Just found this in my GH notifications. Very cool!!! 🤩

PR #814, opened 2025-08-02
Simplify installation guide & better highlight bundle

psobolewskiPhD:
Note: this is a reorganization borne out of a review by Harry Barng at SciPy 2025

# Description
The goal of this PR is to make the installation guide more clearly target the 3 possible user groups: experts, newcomers, and those in between. 
It encourages experts to just use the tools they already use, while highlighting the bundle for those who want to get started quickly. 
I also remove the old bundle section, since it wasn't too helpful, in favor of just directly linking to the dedicated bundle page, which is also being updated, see #813 

I also change the ToC layout a bit, by nesting the Qt backends and constraints under a new heading "Advanced Installation Options".

### DIFF
--- a/docs/tutorials/fundamentals/installation.md
+++ b/docs/tutorials/fundamentals/installation.md
@@ -48,9 +48,10 @@ You may also want:
 New to Python or uncertain about conda, pip, and virtual environments?
 Here are some resources we recommend:
 
+
+- [Talley Lambert's "Getting started running Python" slides](https://hms-iac.github.io/bobiac/content/02_getting_started_with_python/getting_started_with_python.html#slides)
 - [Scientific Python: Getting started with Python for science](https://lectures.scientific-python.org/intro/index.html)
-- [Talley Lambert's Python environments workshop](https://hackmd.io/@talley/SJB_lObBi)
-- [Carpentries GIT workshop](https://swcarpentry.github.io/git-novice) (for those interested in contributing)
+- [Detailed comparison of Python environment management tools](https://www.nijho.lt/post/python-environments/)
 ```
 
 Python package distributions of napari can be installed via `pip`, `conda-forge`, or from source.
--- a/docs/tutorials/fundamentals/installation.md
+++ b/docs/tutorials/fundamentals/installation.md
@@ -18,8 +18,10 @@ kernelspec:
 
 This guide will teach you how to do a clean install of **napari** and launch the viewer.
 
-```{note}
-If you want to contribute code back into napari, you should follow the [development installation instructions in the contributing guide](dev-installation) instead.
+If you're already familiar with Python, napari can be installed from PyPI or conda-forge using your favorite virtual environment and package manager. Otherwise, keep reading for instructions on how to install napari as a Python package.
+
+```{tip}
+If you just want to get started quickly with napari as an app, check out our [napari app installation instructions](installation_bundle_conda.md).
 ```
 
 (install-python-package)=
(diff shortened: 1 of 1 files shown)


psobolewskiPhD:
Why did circle not run on this PR? worked fine on #813

psobolewskiPhD:
@napari-bot make slimfast

docs/tutorials/fundamentals/installation.md
```diff
@@ -39,14 +45,12 @@ You may also want:
 [venv](https://docs.python.org/3/library/venv.html) **(Highly recommended)**
 
 ```{note}
-New to Python or uncertain about conda, pip and virtual environments? You can use our
-[Bundled App](#install-as-a-bundled-app) to get started or first look at some
-resources we recommend:
+New to Python or uncertain about conda, pip, and virtual environments?
```
willingc, line 48: Let's include uv and pixi as modern, user-friendly options.

```suggestion
New to Python or uncertain about pip, conda, uv, pixi, and virtual environments?
```

docs/tutorials/fundamentals/installation.md
```diff
@@ -39,14 +45,12 @@ You may also want:
...
+Here are some resources we recommend:
```
willingc, line 49: Let's drop the following:
- Talley's doc which is dated
- Carpentries GIT workshopp which is off topic

Add:
- Scientific Python: Intro to development https://learn.scientific-python.org/development/tutorials/dev-environment/
- pyOpenSci's Python packaging resources for scientists https://www.pyopensci.org/python-packaging-science.html

willingc (commented):
Love this! A few suggestions:
- Add modern packaging/installation tools which simplify installing Python too
- Update the resource list to include information from pyOpenSci and Scientific Python and drop dated resource and pure Git lesson.

docs/tutorials/fundamentals/installation.md
psobolewskiPhD: Both of those are great, but seem aimed at developers rather than end users so I don't think this is the best place to link them? Seem better in the plugin docs? 
The reason I like Talley's is that it really explains all the basic concepts. It was so crucial for me to wrap my brain around all this stuff. Other than mentioning mamba it doesn't seem to be that dated? and mamba still works technically. 
I will try to find some time to look for some more recent getting started with virtual environments guide that targets newcomers.

docs/tutorials/fundamentals/installation.md
psobolewskiPhD: I left them out because our installation instructions don't yet use those, so I didn't want to further confuse things?
I think in the long run we should re-think the Python installation part and then standardize across our repos.

willingc: It's likely more confusing to leave them out (uv is rapidly becoming a defacto standard for end users to install Python packages). pixi provides some benefits to users with workflows similar to Brian's.

docs/tutorials/fundamentals/installation.md
willingc: I think it is important to have both of the suggested additions. Both are the best places for scientific researchers to find information and support about Python. Building awareness is helpful for users to find not just documents but also communities to support them.

As for Talley's doc, I would be ok with keeping it if we acknowledge that it has good explanations, though it refers to Python 3.7 and 3.8 (Python versions which are no longer supported).

docs/tutorials/fundamentals/installation.md
psobolewskiPhD: But we haven't yet updated any of the downstream instructions to use them and the immediate copy above that section only mentions conda and venv. So I really do think adding other concepts here is too much. 
I think that's the next step is to update the Python package install section to cover the newer tools.

docs/tutorials/fundamentals/installation.md
psobolewskiPhD: I updated to the latest slides from Talley, which are still fantastic for a new comer and then added the broader resource you had shared in zulip. I don't want to overload here with development things, but we should add those two resources in the Contributing section and Plugins guides.

TimMonko (approved):
LGTM adding RTM to double check that the screenshot spawns correctly

PR #816, opened 2025-08-04
Update codespell config and minor corrections

willingc:

Make more consistent with Scientific Python recommended practice:
https://learn.scientific-python.org/development/guides/style/#spelling

# Description

This PR simplifies the codespell configuration to be more explicit. It adds `make spelling` option for easier testing in development. It adds an `ignore-words.txt` since it is more readable than the regex in the current config. It skips some additional files and folders.

### DIFF
--- a/.codespell/ignore-words.txt
+++ b/.codespell/ignore-words.txt
@@ -1,0 +1,5 @@
+nd
+recuse
+Ganes
+Tru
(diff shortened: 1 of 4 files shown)


melissawm (approved):
LGTM and CI is green, ship it!

PR #827, opened 2025-08-22
Fix version switcher URL to prevent problems with unversioned pages

melissawm:
Addresses https://github.com/napari/napari.github.io/issues/427

# Description
Because of this setting, the following happens:

- The latest (dev) version gets build by a PR to napari/napari or napari/docs;
- It is deployed to napari.github.io
- The unversioned action runs, copies over a few pages from the latest version into stable. These pages were built as "dev" so they point to a relative location of version_switcher.json

This relative location does not get updated in the future, and thus the version switcher used for this version of the pages is frozen to be the current version.

I tried adding a comment to explain this problem, hopefully this helps.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -74,12 +74,14 @@ html_sourcelink_suffix = ""
 # Check version and set version_match which is used by the version switcher
 if version == "dev":
     version_match = "dev"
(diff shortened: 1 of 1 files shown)


jni (approved):
Great comment!

PR #829, opened 2025-08-30
Update documentation for dependencies groups

Czaki:

depends on [#8227](https://github.com/napari/napari/pull/8227)

# Description

Update installation instructions for development installation for using dependency groups.  

Currently this PR does not update workflows itself, as some decision may be made in the main PR.


README.md
```diff
@@ -17,7 +17,7 @@ Please read our [contributing guide](https://napari.org/dev/developers/contribut
 ### quickstart: local setup
 * __Check the prerequisites__
     1. Create a clean Python (>=3.10) environment (e.g., with conda).
-    1. In that environment, create an editable `napari` installation with the `docs` dependency group and a Qt backend. For example, after first forking and cloning the main `napari` project if you've not previously done so, run `python -m pip install -e ".[pyqt, docs]"` from your `napari/napari` clone directory. This will use the default Qt backend.
+    1. In that environment, create an editable `napari` installation with the `docs` dependency group and a Qt backend. For example, after first forking and cloning the main `napari` project if you've not previously done so, run `python -m pip install -e ".[pyqt]" --group docs` from your `napari/napari` clone directory. This will use the default Qt backend.
```
psobolewskiPhD:
```suggestion
    1. In that environment, create an editable `napari` installation with the `docs` [dependency group](https://packaging.python.org/en/latest/specifications/dependency-groups/) and a Qt backend. For example, after first forking and cloning the main `napari` project if you've not previously done so, run `python -m pip install -e ".[pyqt]" --group docs` from your `napari/napari` clone directory. This will use the default Qt backend.
```
Looks like the copy here was actually wrong, since it referred to a dependency group! I'm adding a link to the python docs just in case there is some confusion.

docs/developers/contributing/dev_install.md
```diff
@@ -60,31 +60,31 @@ In order to make changes to `napari`, you will need to [fork](https://docs.githu
...
     a Qt backend installed or if you use an experimental backend like PySide6:
     ```sh
-    pip install -e ".[dev]"  # (quotes only needed for zsh shell)
+    pip install -e . --group dev  # (quotes only needed for zsh shell)
     ```
 
     Note that in this last case you will need to install your Qt backend separately.
 
     ```{note}
     In all of the above cases, you may wish to include the `optional` dependency group,
```
psobolewskiPhD:
```suggestion
    In all of the above cases, you may wish to include the `optional` extra (optional dependency),
```
Correcting the copy.

docs/developers/contributing/dev_install.md
```diff
@@ -60,31 +60,31 @@ In order to make changes to `napari`, you will need to [fork](https://docs.githu
...
-    in addition to `dev`, by using `pip install -e ".[dev, optional]"`, for example.
+    in addition to `dev`, by using `pip install -e . --group dev --group optional`, for example.
```
psobolewskiPhD:
```suggestion
    in addition to `dev`, by using `pip install -e ".[optional]" --group dev`, for example.
```
`optional` is an extra

psobolewskiPhD: Can't suggest but below in line 88, please change:
> The `optional` dependency group includes `numba` and other performance packages that are

to

> The `optional` extra (optional dependency) includes `numba` and other performance packages that are

psobolewskiPhD:
i can't suggest, but lets also update the README
https://github.com/napari/docs/blob/583bfffcd3cf86049184281e0d7e384173df19f5/README.md?plain=1#L20
here is another place where the copy was "dependency group" which is/was wrong, but will be correct with this change overall. just need to move the `docs` out of the brackets.

TimMonko (approved):
Approving as if Peter's changes were included.

brisvag:
ping @Czaki, seems pretty small to fix and get in!

brisvag (approved):

PR #835, opened 2025-09-07
Fix @jaimergp's last name

jaimergp:
# Description

I noticed this typo in the release notes. Is there a database somewhere where I need to change the "source" or are these copy pasted across releases? I couldn't find any with a cursory "rodraguez" search across the org.

### DIFF
--- a/docs/release/release_0_6_3.md
+++ b/docs/release/release_0_6_3.md
@@ -163,7 +163,7 @@ Napari now has improved support for PySide6 [(#7887)](https://github.com/napari/
 - [Filippo Balzaretti](https://github.com/napari/napari/commits?author=FilBalza) ([docs](https://github.com/napari/docs/commits?author=FilBalza))  - @FilBalza +
 - [Grzegorz Bokota](https://github.com/napari/napari/commits?author=Czaki) ([docs](https://github.com/napari/docs/commits?author=Czaki))  - @Czaki
 - [Ian Coccimiglio](https://github.com/napari/docs/commits?author=ian-coccimiglio) - @ian-coccimiglio +
-- [jaime rodraguez-guerra](https://github.com/napari/docs/commits?author=jaimergp) - @jaimergp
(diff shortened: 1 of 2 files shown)


psobolewskiPhD (approved):
Gah! Sincere apologies! Sorry we missed that. Not sure how the names are obtained from the handles for the release notes...

I think to fix the display in the 0.6.2 and 0.6.3 versions of the docs we also need a PR to this branch to update the html:
https://github.com/napari/napari.github.io/tree/gh-pages
But this is quite annoying because that branch is massive so getting it checked out is a PITA, but if I recall it works using the web UI?

There are 3 spots:
0.6.2/release/release_0_6_3.html
0.6.3/release/release_0_6_3.html
0.6.3/release/release_0_6_4.html

I think merging this will update the live website dev and stable (0.6.4) and future versions.

Before doing so, maybe @melissawm can confirm that i'm barking up the right tree!

TimMonko:
@jaimergp it looks like in the past, release manager have run the [release making tool](https://github.com/napari/napari-release-tools/blob/main/generate_release_notes.py) with an arg for `name_correction.yaml` -- you can see you are included in it https://github.com/napari/napari-release-tools/blob/9a52e8184bc5c46d1f72d91d55ee2e66eb4fa737/name_corrections.yaml#L9. I haven't been, and having been looking, so I apologized. When Juan was teaching me he hadn't had to use the file in a long while, so he told me to not pay attention. 

The release tooling does pull from your Github info, so maybe you can check there? It may just be swapping the í for an a, so maybe we can check the . Obviously, other names are not correct, so it may not be something you can fix. 

~~@brisvag be sure to run the name correction with `generate_release_notes.py` so this doesn't happen again. :)~~

TimMonko:
Actually I'm confused, maybe @Czaki or @jni has more info? 

By default `generate_release_notes.py` uses the name corrections file...

https://github.com/napari/napari-release-tools/blob/main/generate_release_notes.py#L105-L109

melissawm:
I will take care of the old artifacts in napari.github.io, thanks for the ping!

PR #838, opened 2025-09-12
Auto generate release index with highlights and timeline

TimMonko:
# Description

This PR auto-generates a nifty index page for release notes. In my experience, release notes are one of the most enticing things to read, and made especially more so when I get to be presented with something cool (like a timeline, "what I missed", etc). This tries to do a few things that to me are very compelling:
1. Uses our highlights to make compelling quick reads
2. Uses dropdowns to make a user feel more in control of the otherwise lengthy file
3. Its much more searchable than multiple files in terms of highlights. (i.e. I can search "Labels" and find relevant release highlights)
4. Includes a frame of reference for time, which is often _very_ hard to comprehend as someone who is not following on top of all the releases. Helps much more with the "whats changed since I've last used napari"

I accomplish this by:
1. autogenerating index.md for releases with `update_release_docs.py`
2. ~~hook this in with our stub system, to reduce build time. this does build in about a second though, so I'm not sure this is needed? just being consistent with other files I used as examples~~ its hooked up to stubs, but is currently set to always generate because it seriously takes minimal time. I'm open to whatever we think is best for this
3. changes the sidebar link to _always_ link to this nifty page, rather than the meh experience that is current when clicking on latest release note.

I made this because I was procrastinating a demo, but I really wanted to make it much easier for past napari users to "catch up" outside of us having to pick out _all_ the pieces in our demo.

image


TimMonko:
I added the highlight label because I think this is worth showing to people / having them be able to share with others more easily.

jni (approved):
I haven't looked at the code in detail, but the proof is in the pudding, and the pudding is friggin delicious! 😂 

One thought I had is that we could get the release date from PyPI instead of trying to parse it. But maybe this is worse since it doesn't seem to be a problem for any recent releases, and we autogenerate the date now, and currently the script doesn't need to use any remote APIs which is nice.

TimMonko:
> One thought I had is that we could get the release date from PyPI instead of trying to parse it. But maybe this is worse since it doesn't seem to be a problem for any recent releases, and we autogenerate the date now, and currently the script doesn't need to use any remote APIs which is nice.

Yes, I thought about this too! My  conclusion was the no remote is nice -- less likely to flake on CI. 
Agreed, these can be follow-ups.

Draga also gave this a look over in person, so I'm gong to merge.

PR #840, opened 2025-09-19
Enable dark mode for website

melissawm:
Opening in draft since it depends on new release of the napari-sphinx-theme!

Paired with https://github.com/napari/napari-sphinx-theme/pull/195

# Description
Enables dark mode for the website, including a new switcher to turn dark mode on/off. Also fixes an emoji that was not rendered correctly previously and removes some no longer needed CSS customization.


psobolewskiPhD:
Can't suggest but this will also require editing line 154 in conf.py. Right now we hard-set Light mode.

melissawm:
Great catch @psobolewskiPhD , done!

psobolewskiPhD:
Actually, maybe we don't need to change that? I guess that setting should set the default of the page, but even with the old hard-set light, with the switcher someone could still manually switch to dark if they want, right?

melissawm:
Yes - but as far as I understand it, setting it to auto is best practice for accessibility, so I think it's worth doing!

psobolewskiPhD:
Sounds good -- it's my preferred behavior too 😉

brisvag (approved):
<3

TimMonko (approved):

PR #870, opened 2025-10-15
Add pixi setup for building docs

melissawm:
# References and relevant issues
N/a

# Description
This allows you to run 'pixi run slimfast' to build the docs, without having to clone and install napari from source directly.

Nothing changes if you decide not to use it.

I of course nuked the pixi.toml I had worked on with @jaimergp so hopefully this is correct 😅 

This PR is more of a PoC but I wanted to share with folks.

PR #876, opened 2025-10-26
Enhance documentation build process with pixi integration and Windows…

rensutheart:
### References and relevant issues
Supersedes #870
N/A for issues (no behavior change to deployed docs; optional local build support only)

### Description
This PR adds an optional pixi-based setup to build the documentation using the existing Makefile targets—specifically the fast, no-gallery path—without requiring a local clone or editable install of the napari source.

### What’s included

Add a minimal [pixi.toml](vscode-file://vscode-app/c:/Users/rptheart/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) so contributors can run `pixi run slimfast` to build docs quickly.
Pin Python to 3.13 to align with feedback on #870.
Reuse the Makefile as the source of truth (tasks.slimfast = "make slimfast"); no custom commands or duplicated logic.
Document a simple “Option 2: Using pixi” flow in [README.md](vscode-file://vscode-app/c:/Users/rptheart/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) for cross‑platform slim builds.

### Scope and intent

This is intentionally narrow: it only supports the slimfast build (docs repo content only; no gallery; no notebook cell execution).
Nothing changes if you don’t want to use pixi; the standard manual path remains the default and supports full builds.
The [pixi.toml](vscode-file://vscode-app/c:/Users/rptheart/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is kept single‑platform to avoid lock inconsistencies; users switch the platforms = ["<their-os>"] entry before pixi install.

### Why supersede #870

Incorporates reviewer suggestion to use Python 3.13 rather than a version range.
Keeps risk and surface area low by delegating to the existing Makefile targets instead of introducing new workflows.
Provides clear, minimal documentation for contributors who prefer pixi for a quick local preview.

### Final checklist

Minimal changes for desired functionality
Uses existing Makefile targets; consistent with current CI and local builds
Optional path only; no changes to deployed docs behavior


README.md
```diff
@@ -32,6 +35,77 @@ Please read our [contributing guide](https://napari.org/dev/developers/contribut
...
+
+
+[Pixi](https://pixi.sh) provides a simple, cross-platform way to quickly build the documentation using the `slimfast` target without manually installing dependencies or cloning the napari source repository.
+
+**Note: This option only supports `make slimfast` builds. For full builds with the gallery (`make html`) or other build options, use Option 1.**
+
+1. **Install pixi** (if not already installed):
+   ```bash
+   # Windows (PowerShell)
+   iwr -useb https://pixi.sh/install.ps1 | iex
```
melissawm, line 47: Is `iwr` a tool that will be always present in windows?

melissawm (commented):
Thank you @rensutheart !

@jaimergp would you mind taking a look?

README.md
rensutheart: Yes, `iwr `(Invoke-WebRequest) is a built-in PowerShell command (cmdlet), which should always be available if the user uses PowerShell in Windows.

README.md
```diff
@@ -32,6 +35,77 @@ Please read our [contributing guide](https://napari.org/dev/developers/contribut
...
+   ```
+
+3. **Configure platform (REQUIRED):** In `pixi.toml`, set only ONE platform for your OS:
+   ```toml
+   platforms = ["win-64"]      # Windows
+   platforms = ["linux-64"]    # Linux
+   platforms = ["osx-64"]      # macOS Intel
+   platforms = ["osx-arm64"]   # macOS Apple Silicon
+   ```
+   Only one platform should be present in the array at a time.
```
psobolewskiPhD: Why is this the case? 
My understanding was that pixi supported multiplatform solving, so that projects were truly portable, see also https://pixi.sh/dev/workspace/multi_platform_configuration/
Is it something about our docs build process that breaks this? or am I missing something?

Makefile
```diff
@@ -15,17 +15,29 @@ docs_dir := $(current_dir)docs
 clean:
 	echo clean
 	echo $(current_dir)
+ifeq ($(OS),Windows_NT)
+	@python -c "import shutil, pathlib, glob; shutil.rmtree('$(docs_dir)/_build', ignore_errors=True); [pathlib.Path(f).unlink(missing_ok=True) for f in glob.glob('$(docs_dir)/api/napari*.rst')]"
+else
 	rm -rf $(docs_dir)/_build/
 	rm -rf $(docs_dir)/api/napari*.rst
+endif
```
jni, line 23: rather than having if/else/endif everywhere, I wonder if maybe we should have two different makefiles? Is that even possible? 🤔 It's just strange to have this for every command...

jni: @TimMonko how do you deal with using the makefile to build docs in Windows?

README.md
rensutheart: When I initially tested this, it kept failing for multiplatform use. But it turns out that I had a lock file issue. So once I deleted that it seems to work now. I've updated the PR commits to incorporate this.

Makefile
psobolewskiPhD: I believe if you install git on windows you can launch a shell with bash emulation, so then the regular makefile would work.

jni: (Discussed this with Tim at the hackathon, and this is indeed the standard way to deal with this. Ignore me!)

TimMonko: I discussed with Juan at hackathon, but currently you must run docs with something that allows you to do unix-like (i.e. Git Bash or WSL). Its a HUGE pain as a new contributor, and you can see the amount of docs that are Windows specific already... https://napari.org/stable/developers/contributing/documentation/index.html#building-the-documentation-on-windows

README.md
```diff
@@ -32,6 +35,68 @@ Please read our [contributing guide](https://napari.org/dev/developers/contribut
 These steps should set you up to build and preview your docs contributions on your local machine.
 For more detailed instructions and tips, please visit the relevant sections of our [contribution guide](https://napari.org/dev/developers/contributing/documentation/index.html).
 
+#### Option 2: Using pixi (for slimfast builds only)
+
+
+[Pixi](https://pixi.sh) provides a simple, cross-platform way to quickly build the documentation using the `slimfast` target without manually installing dependencies or cloning the napari source repository.
+
+**Note: This option only supports `make slimfast` builds. For full builds with the gallery (`make html`) or other build options, use Option 1.**
```
melissawm:
```suggestion
**Note: This option only supports `make slimfast` builds, and it will clone the main napari repo from the main branch on GitHub into the pixi environment. For docstring changes, full builds with the gallery (`make html`), or other build options, use Option 1.**
```

Makefile
TimMonko: Two make files both feels approachable and also like a maintenance pain

README.md
```diff
@@ -32,6 +35,68 @@ Please read our [contributing guide](https://napari.org/dev/developers/contribut
...
 For more detailed instructions and tips, please visit the relevant sections of our [contribution guide](https://napari.org/dev/developers/contributing/documentation/index.html).
 
+#### Option 2: Using pixi (for slimfast builds only)
+
+
+[Pixi](https://pixi.sh) provides a simple, cross-platform way to quickly build the documentation using the `slimfast` target without manually installing dependencies or cloning the napari source repository.
+
+**Note: This option only supports `make slimfast` builds, and it will clone the main napari repo from the main branch on GitHub into the pixi environment. For docstring changes, full builds with the gallery (`make html`), or other build options, use Option 1.**
+
+1. **Install pixi** (if not already installed):
```
TimMonko: lets also link the pixi install documentation here, incase the commands change or a user needs to modify a bit (or inspect the source command)

Makefile
psobolewskiPhD: git is pretty much required for contributing via local setup though right?

psobolewskiPhD: The bigger issue is `make` ...

melissawm:
I think we should merge this as-is, since it does not remove any options from the docs building process, but just adds a new alternative. Since it is still incomplete, I propose we follow-up with
1. adding all the other build options from the makefile
2. amend the contributor guide to describe the pixi workflow.

Makefile
psobolewskiPhD: Wait! with pixi we can install make!
https://github.com/conda-forge/make-feedstock

psobolewskiPhD:
The convo was closed above, but adding `make` to the pixi project at least on windows would be very cool.
Then someone on windows with git (and thus git-bash) and pixi would be all set, they'd have the bash emulation and pixi would install make for them!

willingc:
@rensutheart Nice work! 

@melissawm and team, ~~Is there any reason not to add pixi for Linux and macOS builds?~~ Nevermind, I see the `pixi.toml`, now.

Makefile
```diff
@@ -15,17 +15,29 @@ docs_dir := $(current_dir)docs
 clean:
 	echo clean
 	echo $(current_dir)
+ifeq ($(OS),Windows_NT)
```
willingc, line 18: Is this indented correctly relative to the echo statements above?

README.md
```diff
@@ -32,6 +35,68 @@ Please read our [contributing guide](https://napari.org/dev/developers/contribut
 These steps should set you up to build and preview your docs contributions on your local machine.
 For more detailed instructions and tips, please visit the relevant sections of our [contribution guide](https://napari.org/dev/developers/contributing/documentation/index.html).
 
+#### Option 2: Using pixi (for slimfast builds only)
+
+
```
willingc:
```suggestion
```

README.md
```diff
@@ -32,6 +35,68 @@ Please read our [contributing guide](https://napari.org/dev/developers/contribut
...
+     ```powershell
+     $env:Path = [System.Environment]::GetEnvironmentVariable("Path","Machine") + ";" + [System.Environment]::GetEnvironmentVariable("Path","User")
+     ```
+   
+   - **macOS/Linux**: Restart your terminal, or run:
+     ```bash
+     # For bash
+     source ~/.bashrc
+     
+     # For zsh (macOS default since Catalina)
```
willingc:
```suggestion
     # For zsh (macOS default since version 10.15 Catalina)
```

README.md
```diff
@@ -32,6 +35,68 @@ Please read our [contributing guide](https://napari.org/dev/developers/contribut
...
+     source ~/.zshrc
+     
+     # For fish shell
+     fish_add_path ~/.pixi/bin
+     ```
+     
+     **Note:** If you're using fish shell and pixi was just installed, you may need to manually add it to your PATH with the command above.
+   
+   Verify pixi is available by running: `pixi --version`
```
willingc: Let's bold this or make it a separate step.

README.md
```diff
@@ -32,6 +35,68 @@ Please read our [contributing guide](https://napari.org/dev/developers/contribut
...
+4. **Build the docs** (slimfast only):
+   ```bash
+   pixi run slimfast
+   ```
+
+5. **Preview**: Open `docs/_build/html/index.html` in your browser.
+
+**Notes:**
+- The `pixi.toml` is multi-platform; no manual platform selection is required and it should “just work” on Windows, Linux, and macOS.
+- If you switch OS/architecture or update pixi, simply run `pixi install` again to refresh the environment.
```
willingc:
```suggestion
- If you update pixi, run `pixi install` again to refresh the environment.
```

willingc (commented):
Looks good. I suggested a few clarifications.

README.md
willingc: Specifically, call out the verification step.

jni (approved):
Tested on macOS and it works a treat!

TimMonko (approved):
This is AMAZING 🤩 
I tested on Windows. As we can see in contributing docs, Windows is a flipping pain https://napari.org/stable/developers/contributing/documentation/index.html#building-the-documentation-on-windows

As a follow-up (or as part of this PR) we need to  update this documentation page, along with the README like in this file. 

I have tested this on Windows, where I have lots of experience. For example, with current make commands
1. `cmd` (the "older" Windows default terminal) -- doesn't work
2. `powershell` (the older Windows terminal that was never defualt -- doens't work
3. `pwsh` (newest version of Powershell, which is some user's default) -- works (due to Unix command interoperability

NOW with this PR, `pixi run {build}` works flawlessly in all these situations.

The other follow-up we will need to consider for switching to pixi is whether or not to ship a lock file. These lock files, in my expeirence, aren't the easiest to build cross platform, mostly due to Qt, but it _is_ possible. When doing `pixi install` initially the creation of the lock file creates much longer than I would like.

However, I'd also like to reinforce how much better this workflow is than the previous one. When testing I did the only way and didn't have a docs install, so I had to cd out, `pip install -e .[docs]` in napari, and then cd back, and then do it. 🤮 🤮 🤮 

This is so much better!

willingc (approved):

willingc:
@melissawm Let's land this now so folks can use at the hackathon :smile:

Makefile
rensutheart: Yes, the indentation is intentional. `ifeq` is a Make directive, not a recipe command, so it needs to be aligned with the target (`clean:`), not tab-indented like the echo lines.

TimMonko:
Hoooooooraaaaaaaaaaaaaaaaaaay! Congrats on your first PR! 🥳 🎈 🎉 🤩 

cute-lion

PR #878, opened 2025-10-27
Add mdformat to pre-commit config

### LINKED ISSUE #821: Find and apply a suitable linter for documentation pages
## 🧰 Task

Previous discussion exists in PR #54 

Currently waiting on a suitable myst linter for special admonitions.

melissawm:
Closes #821 

# Description
This PR lints the entire repo. I am using mdformat with myst, and there are some limitations and caveats:
- Colon fence syntax is not currently supported by the linter. I have tried to separate out these changes in a specific commit so they're easy to revert if we want to (see https://github.com/executablebooks/mdformat-myst/issues/13 for context)
- There are some escapes enforced here (like `\[`) that we don't _really_ need but I guess they don't hurt and the benefit in the end is a net positive (having the linter turned on by default)
- I have excluded the naps folder from the linter because the rst syntax for the file metadata creates issues. We can explore if we want to touch those in the future but I don't think they are critical to lint.

Most of the motivation for this is to catch syntax issues early before a contributor actually submits a PR or has to build the docs locally.

~~Run mdformat with pre-commit. This is a test to see how the formatter behaves.~~


psobolewskiPhD:
Just for my sanity: you first converted all the colon fencing manually or using another tool, and then the normal mdformat was able to work basically?

melissawm:
Yes, exactly!

willingc (approved):
Looks great @melissawm. I fixed the small merge conflict. Once this is green, I will mark ready to merge.

TimMonko (approved):

melissawm:
Yay thanks folks!

PR #883, opened 2025-10-30
Modernize type annotation to use builtins

Czaki:

# Description

We should not suggest using `List` when all supported Python versions support `list[]`.


brisvag (approved):

willingc (approved):
Agree. Thanks @Czaki

docs/plugins/advanced_topics/npe2_migration_guide.md
```diff
@@ -148,7 +148,7 @@ command for an `npe2` reader.
 #### napari_hook_spec
 
 ```python
-def napari_get_reader(path: Union[str, List[str]]) -> Optional[ReaderFunction]
+def napari_get_reader(path: Union[str, list[str]]) -> Optional[ReaderFunction]
```
psobolewskiPhD, line 151: Does it make sense to change this? it's from the hook_spec, so that would be older napari and older python, right? So maybe we should leave them to match the original code? 
I'm not sure anyone is still migrating anything, so it might be moot...

docs/plugins/advanced_topics/npe2_migration_guide.md
```diff
@@ -235,16 +235,16 @@ has to be imported or run until a compatible writer is found.
 
 ```python
 def napari_get_writer(
-    path: str, layer_types: List[str]
+    path: str, layer_types: list[str]
```
psobolewskiPhD, line 238: Same comment as above, maybe we should leave this since it's from an era when List was the correct type hint.

docs/plugins/advanced_topics/npe2_migration_guide.md
```diff
@@ -235,16 +235,16 @@ has to be imported or run until a compatible writer is found.
...
 ) -> Optional[WriterFunction]
 ```
 
 Where the `WriterFunction` is something like:
 
 ```python
 def writer(
-    path: str, layer_data: List[Tuple[Any, Dict, str]]
-    ) -> List[str]
+    path: str, layer_data: list[tuple[Any, dict, str]]
```
psobolewskiPhD, line 246: same as above.

psobolewskiPhD:
Not a blocker at all, but maybe we shouldn't change the hook_spec code? It's from a different era and was correct at the time. Someone migrating -- though I doubt anyone is doing that now? -- would encounter that code.

willingc:
@psobolewskiPhD If you are at all concerned, perhaps the best thing would be to add a note to the doc that this applies to Python 3.9 and later. If someone wants to use 3.8 or earlier, they would need the older typing format.

psobolewskiPhD:
Eh.
I'm not concerned really, I don't think anyone is using this documentation anymore. I just find it odd to change something retroactively. I do wonder if it was an intentional change or search-n-replace 😜

PR #889, opened 2025-11-28
Remove outdated mentions about PySide2 in documentation

Czaki:

Depends on https://github.com/napari/napari/pull/8450

# Description

As we drop PySide2, we need to update documentation to not suggest using it.


brisvag (approved):
LGTM!

PR #900, opened 2026-01-07
Allow to use local copy of napari when using pixi

Czaki:

# Description

The current pixi configuration works with the main branch of napari. So it cannot be easily used to debug problems with failing PR to the napari repo (like https://github.com/napari/napari/pull/8051#issuecomment-3700953859)

This PR adds an environment that installs napari from a local clone where the proper PR could be checked out.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -7,12 +7,24 @@ name = "napari-docs"
 platforms = ["win-64", "linux-64", "osx-64", "osx-arm64"]  # All platforms
 version = "0.1.0"
 
-[dependencies]
+[feature.base.dependencies]
 python = "3.13.*"
 make = "*"
 
 [pypi-dependencies]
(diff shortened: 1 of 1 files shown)


melissawm (approved):
Can you add a comment to the toml file about how to use the local napari clone when building? Does the pixi command change?

Czaki:
> Can you add a comment to the toml file about how to use the local napari clone when building? Does the pixi command change?

What did you mean? just add `-e env_name` to command. Like

```
$ pixi run -e local-napari python docs/_scripts/update_ui_sections_docs.py
```

melissawm:
Exactly, I had no idea about that 😅 I assume other people won't know either. Would you mind adding that line as a comment to the pixi.toml file?

Czaki:
@melissawm Could you check now?

melissawm (approved):
Perfect, thank you!

PR #904, opened 2026-01-12
Fix pixi tasks for different environments

melissawm:
Follow-up to #900

# Description
After we merged https://github.com/napari/docs/pull/900, I can't build the docs locally with pixi. No tasks are shown when I run 

```
➜  napari-docs git:(main) pixi task list
Tasks that can run on this machine:
-----------------------------------

Task  Description
➜  napari-docs git:(main) 
```

which means pixi doesn't see any of the tasks defined in our pixi.toml file. Doing some research on the docs, it looks like removing the `no-default-feature = true` option from the environment solves it. Another option is to list each task again for each of the two environments, but I believe this PR is the simplest fix.


Czaki (approved):
Yes.

melissawm:
Thanks!

PR #881, opened 2025-10-30
Reorg of the Usage section of the docs

melissawm:
# References and relevant issues
N/A

# Description
At the napari hackathon @ gloBIAS, we had the chance to discuss the docs organization with a few users and came up with ideas for potential improvements to our Usage section. We also discussed some changes to our API docs, but those will be handled separately.

~~This is a first very rough draft that is clearly incomplete, but I wanted to stash it somewhere so the work can continue.~~

EDIT: This is now ready for review. It only touches the first sections of the user guide but I think this is reasonable for now. This was discussed at the docs meeting this week (nov 20/21).

PR #908, opened 2026-01-13
Remove mention of outdated plugin from quick start

TimMonko:

Follow-up to #881 

# Description

While reading through the new quickstart, I noticed this mention of an un-maintained plugin. For a while, the plugin did not work, but I do think it does appear to be fixed now https://github.com/MouseLand/cellpose-napari/pull/51#issuecomment-2351146003

I propose we remove this very small mention.

We can consider adding back plugins once we have our review system in place. My preference would be for us to make a "starter plugins" guide, rather than a short quip (that's hard to notice because of its placement!) in this quick start. There already is good detail here pointing people towards the hub :)

### DIFF
--- a/docs/getting_started/quick_start.md
+++ b/docs/getting_started/quick_start.md
@@ -170,10 +170,6 @@ the default keybinding {kbd}`m`.
 </figure>
 ```
 
-Several plugins can perform automatic segmentation that takes image layers as input and generates labels layers as output.
-
-Try [cellpose-napari](https://www.napari-hub.org/plugins/cellpose-napari) if you have cell images.
-
 +++
 
 ## Get the cell area measurement


TimMonko:
? not sure I understand the fail. I don't see anything useful in the circleci output

psobolewskiPhD:
I don't have time right now, but here's my process:
search for warning
then google the warning
image

TimMonko:
Yeah I did but that's irrelevant to my PR, right?I assume its about the mermaid's but I dont get it

melissawm:
sphinxcontrib-mermaid had a new release with breaking changes, and we don't pin it. So you need to change this line

https://github.com/napari/docs/blob/1f9759bc61c0697cdbf9f805e334dbbc4b2fca28/docs/conf.py#L200C1-L200C25

to read 

`mermaid_include_elk = False`

and the build works (tested locally)

You can see the config value here: https://sphinxcontrib-mermaid-demo.readthedocs.io/en/stable/ (old version) and here https://sphinxcontrib-mermaid-demo.readthedocs.io/en/latest/#mermaid-include-elk (new version)

psobolewskiPhD:
We do have constraints though, so if this is due to a package update, then this suggests they aren't working properly.

melissawm:
Well, it looks like we should be using the constraints: https://github.com/napari/docs/blob/1f9759bc61c0697cdbf9f805e334dbbc4b2fca28/.circleci/config.yml#L69

which do pin sphinxcontrib-mermaid https://github.com/napari/napari/blob/326d0144b22b7808cd16dfeb49c48590d94869d9/resources/constraints/constraints_py3.12_docs.txt#L710

But this doesn't take effect in circleci, so sphinxcontrib-mermaid which is not pinned here https://github.com/napari/napari/blob/326d0144b22b7808cd16dfeb49c48590d94869d9/pyproject.toml#L197 is installed to the latest version 😬 

Maybe uv doesn't take in to account the PIP_CONSTRAINTS argument? https://app.circleci.com/pipelines/github/napari/docs/1752/workflows/adef5c30-e6be-41c9-9701-1d9caa83e8c2/jobs/1753?invite=true#step-106-3450_31

TimMonko:
@Czaki

psobolewskiPhD:
Yeah it's something like that, but not quite because here's the GA run which has the older version, so presumably from constraints and it's also uv
https://github.com/napari/docs/actions/runs/20971807220/job/60276799171?pr=908#step:7:259

psobolewskiPhD:
If you look at the working action it uses:
https://github.com/napari/shared-workflows/blob/b5f068979222e2929547fc42226bd83122be4568/.github/workflows/build_docs.yml#L72

PR Inc.

TimMonko:
@psobolewskiPhD and @melissawm true 🧙s
thanks friends 🥳

psobolewskiPhD (approved):

PR #912, opened 2026-01-14
increase stack size to solve import recursion problem

Czaki:

unlock napari/napari#8051

# Description

Currently napari/napari#8051 is failing the docs build because the `pydeps` graph being built during the "generating docs for UI section: Application status bar" step hits a recursion error. Although that PR doesn't really touch our code that much (and certainly doesn't touch the Application status bar), it **does** introduce a couple of new gallery dependencies, and their presence in the docs environment may be causing this error.

As I have problems with finding the best way to improve pydeps, this PR is just increasing the stack limit to merge the example in 0.7.0 release.

### DIFF
--- a/docs/_scripts/update_ui_sections_docs.py
+++ b/docs/_scripts/update_ui_sections_docs.py
@@ -780,6 +780,9 @@ def main(stubs=False):
 
 if __name__ == '__main__':
     import argparse
+    import sys
+
+    sys.setrecursionlimit(5000)
 
     parser = argparse.ArgumentParser(description='Update UI sections docs.')
     parser.add_argument(


DragaDoncila:
@Czaki have you run this locally with napari/#8051? When I tried increasing the recursion limit I still hit the same error, but maybe I was doing something wrong. I've updated the description and I'll approve and merge, but I'm just curious if you saw a successful build :laughing:

DragaDoncila (approved):
Happy to try this so we can get that PR in.

Czaki:
> I've updated the description and I'll approve and merge, but I'm just curious if you saw a successful build 😆

Yes. Multiple times. Using isolated venvs. Maybe your environment contains even more packsges.

PR #920, opened 2026-01-23
Remove mention of double click in macOS app instructions

dstansby:
It's possible to open the macOS app without double clicking (e.g, cmd+o on the keyboard, or single clicking in the downloads folder), so change the wording to simply "Open".


docs/getting_started/installation.md
```diff
@@ -299,7 +299,7 @@ class: admonition
 ```
 
 Once you have downloaded the appropriate macOS package file, you will have a file with a name like `napari-0.5.0-macOS-x86_64.pkg`, but the version number will be {{ napari_version }}.
-Double click this file to open the installer.
+Open this file to open the installer.
```
psobolewskiPhD: how about this? I think double-click is the most idiomatic and some people could find just "open" confusing? (I for one never think about it that way)
```suggestion
Double-click or open this file to launch the Installer.
```

TimMonko: Hey @dstansby curious your thoughts on Peter's feedback.
If you don't get around to this, then I'll commit the suggestion within the week so it's released with 0.7.0.

TimMonko (approved):

psobolewskiPhD (approved):

PR #922, opened 2026-01-26
Update release notes for 0.7.0a2

DragaDoncila:
# Description

Expand on some highlights and pull in latest 0.7.0 PRs

### DIFF
--- a/docs/release/release_0_7_0.md
+++ b/docs/release/release_0_7_0.md
@@ -47,11 +47,11 @@ our [Plugins Zulip chat
 channel](https://napari.zulipchat.com/#narrow/channel/309872-plugins) or by
 coming to one of our [community meetings](meeting-schedule).
 
-### Grid overlay - bigger, better, faster 📈
+### Grid mode - bigger, better, faster 📈
 
-If you've been playing with our new grid overlays since 0.6.5, you 
+If you've been playing with our new grid mode since 0.6.5, you 
 may have stumbled into performance issues when progressively adding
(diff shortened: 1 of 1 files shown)


docs/release/release_0_7_0.md
```diff
@@ -42,32 +47,91 @@ our [Plugins Zulip chat
 channel](https://napari.zulipchat.com/#narrow/channel/309872-plugins) or by
 coming to one of our [community meetings](meeting-schedule).
 
-### Grid Overlay
+### Grid overlay - bigger, better, faster 📈
```
brisvag: Not an overlay 🤔  Grid view is what we usually call it, or grid mode. 

We *will* have a grid overlay soon (napari/napari#7827) so better be clear :P

brisvag (changes requested):
Looking good to me, a part from the grid section!

docs/release/release_0_7_0.md
DragaDoncila: Derp, yep. Thanks @brisvag should be fixed now

brisvag (approved):
Nice! I do wonder if we really need to say so much detail about the angles here (it's a big section for just "highlights"), or if instead we could put the technical details elsewhere (or just point to the PR) and simply say that this is a breaking change but they are better?

DragaDoncila:
@brisvag good question. I'm going to merge for now and tag the next alpha, but I'm happy to make it a bit shorter - I think it's one that a few users/contributors have been following, so I dedicated a bit of detail to it, but I can revise in an upcoming PR!

PR #928, opened 2026-02-05
Update scripts to pydantic v2

brisvag:
- part of napari/napari#8509

# Description
Updating the scripts to work with pydantic v2.

### DIFF
--- a/docs/_scripts/update_event_docs.py
+++ b/docs/_scripts/update_event_docs.py
@@ -86,8 +86,8 @@ class Ev:
 
         if issubclass(self.model, ViewerModel):
             return f'viewer.events.{self.name}'
-        for name, field_ in napari.Viewer.__fields__.items():
-            if field_.type_ is self.model:
+        for name, field in napari.Viewer.model_fields.items():
+            if field.annotation is self.model:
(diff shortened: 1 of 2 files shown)


brisvag:
@Czaki since we needed to modify this anyways, I also removed the dependency on `napari._pydantic_compat`, so I think we can remove that from there as well?

psobolewskiPhD:
This depends on the napari PR right? so we need to merge that to unbreak CI in this PR?
Of course that means docs builds will be broken in the napari PR until this is merged?
🥶

brisvag:
Yeah we're in a bit of a loop here I think :sweat_smile:

Czaki:
We might think about moving part of these scripts to the  main repository. 

Or just do cross checks against PRs and do do yolo merge

DragaDoncila:
I think if we check locally that docs build with the branch from napari/napari#8509 then it's fine to merge this as is - but maybe we merge it when #8509 is also close so that we don't have docs broken for a long time?

melissawm (approved):
Builds fine locally with the branch indicated above, so LGTM! But I agree we should wait and not break the docs build.

PR #929, opened 2026-02-06
Remove npe1-related docs

DragaDoncila:
Partially addresses napari/napari#8447

# Description
This PR removes references to the napari plugin engine, hook specifications, and npe1 generally. I've removed the hook specification reference entirely (it'll still be available on old docs versions). I've kept the migration guide more-or-less unchanged, so folks should still have all the info they need to update their plugins. I've also updated various examples using hook specifications to use the npe2-style contributions.


docs/roadmaps/0_3_retrospective.md
```diff
@@ -66,7 +66,7 @@ Once the above goals are met, we will develop napari's capabilities for image pr
 
 - Support for reader plugins [#937](https://github.com/napari/napari/pull/937) and writer plugins [#1068](https://github.com/napari/napari/issues/1068) to allow **viewing of domain-specific data and saving of annotations**. For more details see the [`plugins` label](https://github.com/napari/napari/labels/plugins) on our repository.
 
-  > We have a stable specification for adding both [reader and writer plugins](hook-specifications-reference) that is being used in the community.
+  > We have a stable specification for adding both reader and writer plugins that is being used in the community.
```
DragaDoncila, line 69: Docs wouldn't build using my branch from napari/napari#8622 without removing this, so I've removed it. Not sure if there's a better way to handle a historical reference like this.

docs/further-resources/glossary.md
```diff
@@ -5,9 +5,7 @@
...
-| event              | Events and connections are a way to tell napari "If this event occurs, then call this function". [Hooking up your own events](connecting-events).                                                                                                                                                                                                                                                                                                                                                                                                 |                                                                    |
```
DragaDoncila, line 8: Maybe we want to keep these (without the reference links) for historical reasons? But I didn't think it was worth it.

DragaDoncila:
Hmm I don't understand [why it says](https://app.circleci.com/pipelines/github/napari/docs/1810/workflows/055abcee-0e29-41de-83fd-1c6e2554750b/jobs/1811?invite=true#step-107-39124_128) `napari-plugin-template` is undefined...

DragaDoncila:
Drat, the other failure is about a missing hookspec in a docstring. I [remove that line](https://github.com/napari/napari/pull/8579/changes#diff-39f76adc7daef9346eb47709be873ec4914aae7624431035e25cb3ab7d3bbb32L501) in napari/napari#8579... So I guess we need to:

1. ~~Fix the issue about undefined label napari-plugin-template from above~~ done
2. Merge napari/napari#8579
3. That should get this PR passing, merge it
4. Then napari/napari#8622 should be ok to build docs?

jni:
> Hmm I don't understand [why it says](https://app.circleci.com/pipelines/github/napari/docs/1810/workflows/055abcee-0e29-41de-83fd-1c6e2554750b/jobs/1811?invite=true#step-107-39124_128) `napari-plugin-template` is undefined...

search for napari-plugin-template in the docs? a label is another word for a link. It's a labeled link. So I suspect you've deleted the place where the label was defined but you are still linking to it. I think?

jni:
> So I guess we need to:

Yes I think you have the order of operations right

psobolewskiPhD:
Sounds like the right plan @DragaDoncila 
The remaining warning is the link from napari/napari you already have removed in the other pr.
this is the price we pay for the two-repo-solution.

DragaDoncila:
@napari-bot make html

docs/roadmaps/0_3_retrospective.md
brisvag: You could hard-link it using a full URL, since it doesn't need to update with new versions... But IMO this is fine, we don't really need people to go here.

docs/further-resources/glossary.md
brisvag: Agree, not worth keeping.

docs/plugins/advanced_topics/npe2_migration_guide.md
```diff
@@ -439,17 +439,19 @@ contributions:
...
-When this option is enabled and a plugin using the legacy plugin manager API
+When a plugin using the legacy plugin manager API
 is loaded for the first time, the plugin will be imported as usual and
 contributions will be discovered. A "shim" npe2 manifest representing the
 plugin's contributions will be created and cached locally. On all future
 launches of napari, that cached manifest will be used and the plugin will
 *not* be imported immediately when napari boots.
 
+With napari v0.7.0, legacy npe1 plugins will **only** be usable if they
+can be "shimmed" via this adaptor.
```
brisvag:
```suggestion
With napari v0.7.0, legacy npe1 plugins will **only** be usable through
this adaptor; plugins relying on import-time behaviour may malfunction. 
```

docs/plugins/building_a_plugin/best_practices.md
```diff
@@ -379,51 +379,46 @@ If you are vendoring other projects, please add an acknowledgement in your READM
...
 class MyWidget(QWidget):
     def do_something_amazing(self):
         return something_amazing()
 
-@napari_hook_specification
-def napari_experimental_provide_dock_widget():
-    return MyWidget
+class FastWidget(QWidget):
+  def do_something_fast(self):
+    return np.zeros((10, 10))
```
brisvag:
```suggestion
    def do_something_fast(self):
        return np.zeros((10, 10))
```

docs/plugins/building_a_plugin/first_plugin.md
```diff
@@ -303,6 +303,8 @@ which you can upload to [PyPI](https://pypi.org/) or share with others.
 However, you may wish to automate distribution of your plugin using Github,
 see below and the [Publishing](plugin-deploy) guide for more details.
 
+(napari-plugin-template)=
+
```
brisvag:
```suggestion
```

brisvag (approved):
Some suggestions, but otherwise good!

docs/plugins/building_a_plugin/first_plugin.md
DragaDoncila: @brisvag I added this here on purpose because we otherwise have nowhere to point people to the template. This label is used in a couple of spots. I'd be open to adding a different, better section on the template elsewhere, but for now I thought it was better to go here than nowhere.

Although I note now that we refer to the template explicitly by full URL in other places, so maybe we just do that everywhere? idk

brisvag: Uhm, I deleted only the white space I think, or is the suggestion wrong? Anyways, it was just an aesthetic suggestion, not to remove the reference :)

TimMonko (approved):

PR #943, opened 2026-02-17
Add info about new points/shapes layer button behavior

### LINKED ISSUE #938: Add explanation to Viewer tutorial about how new layer buttons work
## 🧰 Task
We updated the new layer button behaviour in napari/napari#8357, this needs to be documented in the viewer tutorial, with a screenshot of the button change visual once that's merged napari/napari#8649

TimMonko:

Closes #938 

# Description

This PR was worked on during today's code cafe (hmm, I guess if I'm doing these I should make a hackmd for it) with @ClementCaporal, whom gave excellent language suggestions.

We discussed diataxis and how this may be considered reference material, but because of the user facing visualization and #938 source (a docs discussion) we opted to add to the viewer tutorial.

This adds a small admonition description the visuals and also adds an image of the layer buttons, which does not exist in the viewer tutorial.

I think we could bikeshed more pieces of the viewer tutorial, but this is better reserved for a follow-up (like why this header is nested where it is).

Follow-ups will include keeping an eye on us adding functionality to the labels layer.


DragaDoncila (approved):
LGTM! Thanks @TimMonko and @ClementCaporal!

docs/getting_started/viewer.md
```diff
@@ -293,12 +293,22 @@ and these changes will instantly propagate to the GUI. For more information abou
...
 New empty `Points`, `Shapes`, and `Labels` layers can be added to the viewer using the layer buttons between the layer controls and layer list. This is equivalent to, for example, the following code to make an empty `Points` layer:
 
 ```python
 viewer.add_points()
 ```
 
+```{admonition} Tip: New layer button behavior
+---
+class: tip
+---
```
psobolewskiPhD, line 307: I believe you can just use:
````{tip}`
(3 backtick, {tip})

File: docs/getting_started/viewer.md [hunk 2]
```diff
@@ -293,12 +293,22 @@ and these changes will instantly propagate to the GUI. For more information abou
...
+Creating a new points or shapes layer with the layer buttons will inherit the dimensions and scale of the selected layer(s), as shown by the highlight around the buttons.
+To create layers that inherit the full dimensions and mixed scale of all the layers in the layer list, select/deselect all layers.
```
psobolewskiPhD: Minor clarification -- that's assuming I understand the behavior correctly (i've not used it much!).
```suggestion
To create layers that inherit the full dimensions and mixed scale of all the layers in the layer list, either select all layers or ensure no layers are selected.
```

TimMonko: You are correct, and I believe that is better wording, thanks!

docs/getting_started/viewer.md
TimMonko: Interesting, I was just looking at https://github.com/TimMonko/docs/blob/5938867e00e40eecac4de8036004bef5c209dbfc/docs/getting_started/viewer.md#L116-L120
which used this format. But it does seem like you are correct for myst admonitions https://mystmd.org/guide/admonitions#admonition-titles

Maybe as silly of a reason as it is, the vscode extension MyST-Markdown at least properly renders the format I used, but butchers this syntax. Not that such should be the driving factor

(I'm having trouble building the docs because of all our changes to pydantic, I think)

psobolewskiPhD: Im not sure which is more proper 😬

psobolewskiPhD (approved):

docs/getting_started/viewer.md
DragaDoncila: The tip looks rendered fine to me on the built docs so I'm gonna merge but maybe @melissawm can tell us what the preferred formatting is for the admonitions? I've always used the backticky one

PR #946, opened 2026-02-22
Add redirects to pages moved in Usage section

melissawm:
Follow-up to #881, depends on napari/napari#8690

# Description
Adds the sphinx-rediraffe extension to manage redirects of the old pages, so the reader flow is not broken between versions (i.e. if you click on docs/fundamentals/installation, it will redirect you to docs/getting_started/installation


DragaDoncila:
@melissawm [looks like](https://github.com/napari/docs/actions/runs/22606456988/job/65499466825?pr=946#step:9:937) an actual failure here?

melissawm:
Oh yeah, looks like redirecting to an anchor doesn't work so I'll just add a redirect to the installation page itself.

DragaDoncila (approved):
LGTM! Thanks @melissawm

Issue #793, opened 2025-07-22
Emphasise running napari from Python on homepage with examples


jni:
I would like the homepage to emphasise that napari is usable within Python, rather than (just) a standalone application, because several folks at SciPy over the years have been confused thinking that napari is a standalone app and thus not for them. Having the "code" below the video, removed in #780, was inspired my the matplotlib home page, which I love:

matplotlib.org

<img width="2052" height="1964" alt="screenshot_fDqKi3Dd@2x" src="https://github.com/user-attachments/assets/f8891d04-bfc1-4a07-87c1-02fbf3571d09" />

Each refresh yields a different example with the relevant function call. I'd love to get to something like that, even if the previous home page design didn't really achieve that.

_Originally posted by @jni in https://github.com/napari/docs/issues/780#issuecomment-3067320194_
            

Ideas based on discussion in #780:
- have several examples that rotate on the home page
- maybe pull them straight from the gallery using the tag "highlight" or "homepage" or something like that
- ideally, figure out generation of videos in gallery so that it's not just static images on the home page
- discussion in #780 is of a carousel, but that idea has been shot down before (#415), so maybe random choice +/- a link to all the choices (e.g. link to the highlight tag in the gallery) is better? 🤷

psobolewskiPhD:
For me the key to this is rendering one of a selection of examples on the homepage, and making it clickable to the example page with the code.
Personally, I'm fine without a video if the image is compelling, but napari-animation has videos generated from examples, so it should be possible: https://napari.org/napari-animation/

PR #953, opened 2026-03-13
Restructure homepage and add rotating featured examples

TimMonko:

Address #793, I think?
~~Also, I think it will need to wait until 0.7.0 releases because it will look for the example images in stable static ... right? They won't be there until after 0.7.0 releases.~~ This is no longer true since we use versioned homepage now

# Description

Revamp of homepage, will detail out later. Having a bit of trouble full building locally so I'm pushing to make the bot build for me lol. This makes the homepage MUCH easier and quicker to parse _and_ I think actually provides more information. I prompted copilot with Claude Opus to set this up and it did a REALLY bad job aesthetically (I still have css cleanup to do), BUT it use this whole .js stuff that I'm clueless about and that actual works (and is what matplotlib uses). So it was sort of a win and sort of a fighting a toddler moment. In the end, I combed over most of the code myself lol.

1. Gets ride of sidebar on homepage only
2. Adds a search button in the navbar persistent on all pages, to continue to allow searching from home page
3. Gets rid of many cards, especially those that are redundant with navbar links (like community and contributing)... **TODO: realizing I should probably drop the plugins card too...? But this one is more of a "heres how napari works" compared to community and contributing**
4. Takes the quicklinks from the sidebar and makes them prevalant, but still less downplayed parts of the homepage
5. Adds a rotating example from the gallery -- `featured_examples.json` allows us to control which examples. We could probably do this as a gallery tag, but I thought it was nice to control here (rather than in napari/napari) how examples are rendered and we can be more precise about home page wording for the individual example . Also notice the dropshadow around the media, so we won't have the really sad white border around it like with the above viewer example.

image

## Some more ideas

1. have a copy-pastable `uvx "napari[all]" remote_path_to_example.py` to show it running python files?
7. Instead of the "really nice" examples, we could do more basic ones to demonstrate API like `viewer.add_image()` leading to the add image example? I don't really like this though
8. We could also just ref any page we want and any image, the `featured_examples.json` just takes keys that are title, description, doc path, and image path. So if we have in dpeth tutorials or other things to highlight (or we can make another dynamic section with the same model that highlights other parts of the docs)
9. Ignore the example part, but do keep the sidebar removal + moving those key links over + the install + download cards.

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -69,7 +69,7 @@ html[data-theme="dark"] div.sphx-glr-download a:hover {
   display: grid;
   align-items: center;
   gap: 1.5rem;
-  grid-template-columns: 3fr 2fr;
+  grid-template-columns: 1fr 1fr;
   margin: 1.5rem 0 2rem;
 }
 
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -119,23 +119,26 @@ html[data-theme="dark"] div.sphx-glr-download a:hover {
   }
 }
 
-/* ── Homepage quicklinks (icon row) ─────────────────────────────────── */
+/* ── Homepage quicklinks (icon columns) ──────────────────────────────── */
 
 .homepage-quicklinks {
   border-bottom: 1px solid var(--pst-color-border);
   display: flex;
   flex-wrap: wrap;
-  gap: 0.5rem 2rem;
+  justify-content: center;
+  gap: 0.75rem 4rem;
   margin: 1rem 0 1.5rem;
   padding-bottom: 1rem;
 }
 
 .homepage-quicklinks a {
   align-items: center;
-  display: inline-flex;
-  font-size: 1rem;
+  display: flex;
+  flex-direction: column;
+  font-size: 0.9rem;
   font-weight: 500;
-  gap: 0.4rem;
+  gap: 0.5rem;
+  text-align: center;
   text-decoration: none;
 }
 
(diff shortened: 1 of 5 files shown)


TimMonko:
@napari-bot make html

TimMonko:
@napari-bot make html

TimMonko:
@napari-bot make html

brisvag:
Nice stuff! Love the rotating example thing. Some comments:

- IMO taking up half the "card real estate" for installation is a waste. I think the simple `Installation` card was nicer, with the split between bundle and python happening on the page itself. If you're worried about non-python folks getting intimidated by the beginning of [installation](https://napari.org/stable/tutorials/fundamentals/installation.html#napari-installation), I'd rather reshuffle that one so the bundle info is higher up.
- not sure I love the position of the quick links (calenda, chat, etc): have you tried below the rotating example?
- we could have even two rotating examples IMO! Push that boring funding header lower down xD

DragaDoncila:
This is fully lit @TimMonko :star_struck:. I love the example, I love that the sidebar is gone. 

Unlike @brisvag I like dedicating the two different cards to installation, and in fact I'd want to highlight them even more - maybe as their own different coloured buttons above the row of cards -- then we'd have more room for more cards!

I agree the calendar links etc. look a little out of place. I love the icons, but I almost think the icons should be bigger and sit centered on top of the title of the link, to give more breathing room to each of the items. 

I'm ambivalent about having two examples. I think one is ok, and we should take up more room with other stuff, like separating the installation cards and giving more room to the socials links

jni:
- I also like two install cards and I love the wording of the bundled app card
- I also prefer a single example (simplicity is very important in the home page) and I might go a bit further: what if we replace the main video with the examples? Videos are great: we can make video for each curated example, and/or we can autogenerate videos (or first manual, then autogenerate)
- I also do want to keep the funding panel prominent because we really need to get to work on adding more icons there — and the more prominent the more people will want to pay to put an icon there, maybe? 😂

TimMonko:
Thank you everyone for such great feedback! 
1. Instead of adding "two examples", we could add 1 examples and 1 link to more narrative documentation -- but I agree that for now we need to keep the homepage simpler and we dont have great narrative documentation (like the tutorials) for modern napari. 
2. I agree about using this to replace the video in a future follow-up, but right now we don't have the infrastructure.
3. I am also very much in favor of the two different install and download buttons, and completely agree with Draga that they should be highlighted.
4. Also we _need_ to push this for 0.7.0. I'm still in "unversioned" pages brain, but if we don't then we won't get the homepage lol. It also means all the examples will be there on the latest branch, so we need not worry (ni other words we will see what circle ci shows)

Here is where I've landed
I am least tied to the placement of the community quick links, after playing for a bit, I do feel ok with them being below the example. (see below)
Two things I can't figure out: the css causing the dark mode Examples Gallery button to not be quite the same as the cards and I can't for the life of me figure out the css to make the quicklink icons bigger... 
I've tried incredibly hard and need to give up, so if people are really torn by this you can fix it :P

image
image


Here is where we _could_ land, based on everyone's feedback. But I will list why I'm making different design decisions.
1. We don't have four high priority cards that "sell" napari, we really just have two: quick start (which is now changed to not go to launch... what were we doing on the old homepage?) and plugins (and even that is redundant with a header. We could link to thinks like howtos and such, but someone would also have the pathway of "usage" to get there. Remember we want the **highest value** things on our homepage
5. As such, I like have a 1x4 row where we keep a visual hierarchy due to color, but have a nice natural "left to right" progression of material importance. 
6. In that light, I originally placed the icons between the cards and the examples because I'd argue we want the objective hierarchy that such placement implies -- these ARE important to the napari project, but have slightly less visual hierarchy compared to the cards. Furthermore, I think this creates a nice visual boundary between the examples and the cards. As in the below screenshots, I think it results in them being too unimportant but we realyl dowant to highlight them.

image
image

@napari-bot make html

psobolewskiPhD:
I like the vibe and getting rid of the side bar.
Radical suggestion: what if we nuke the video (it's always the same, static) and put the example there front and center? 
(Also: dark mode looks pretty bad with the white around the video! wow! 😮 has it always been like that?!)

I strongly prefer the text tiles 4 side by side, rather than the 2x2 box.

Edit: So my vision is something like:
napari: a fast ...
the example block moved up
the tiles in a row
the bullet points? optional

--

Killer followup feature would be to have the example image rendered using the matching theme (so we'd need to make 2 at build, one with each theme and then swap em with some 🪄 )

brisvag:
> (Also: dark mode looks pretty bad with the white around the video! wow! 😮 has it always been like that?!)

Yes xD

DragaDoncila (approved):
@TimMonko I think this is absolutely gorgeous, and agree would want it in for 0.7.0. There are things I think we can improve, but I think they should be done in a follow-up PR:

- Add a separator before and after the socials/calendar links, to give them a "row of their own". We have a separator underneath, but that's basically invisible in light mode, and I think we should have one on top too. We should also improve the look of the separator in light mode (again, separate issue). Possibly also add faint vertical separators between each item? Might be useful for mobile
- Potentially get rid of the video and bump the examples, but I will say the video is very beautiful, and the examples are static and sometimes not that impressive, so I still think there's a reason to keep the video around. Again, can be discussed in a follow-up PR.

I don't want to bikeshed this PR too much because afaic this is a strict improvement on what we currently have, and we can always iterate on it in future.

TimMonko:
For me, this is ready to go now

> I like the vibe and getting rid of the side bar. Radical suggestion: what if we nuke the video (it's always the same, static) and put the example there front and center? (Also: dark mode looks pretty bad with the white around the video! wow! 😮 has it always been like that?!)

Yes the video has been like that on dark mode, its why I emphasized the drop shadow for the Example, because Juan has said the reason he liked the video is because it includes the drop shaodw. I also do like keeping the video front and center like Draga said, because it really emphasizes what napari is capable of.
 
> I strongly prefer the text tiles 4 side by side, rather than the 2x2 box.

👍  


> Edit: So my vision is something like: napari: a fast ... the example block moved up the tiles in a row the bullet points? optional

We can iterate in the future. I personally like having introduction text on a website. I don't like going to them and having to scratch my head what the "product" is.

> Killer followup feature would be to have the example image rendered using the matching theme (so we'd need to make 2 at build, one with each theme and then swap em with some 🪄 )

Oof the build is already so slow, but I do like the idea.


>     * Add a separator before and after the socials/calendar links, to give them a "row of their own". We have a separator underneath, but that's basically invisible in light mode, and I think we should have one on top too. We should also improve the look of the separator in light mode (again, separate issue). Possibly also add faint vertical separators between each item? Might be useful for mobile

Add seperators everyone, and a nice mobile layout

image

image



>    Potentially get rid of the video and bump the examples, but I will say the video is very beautiful, and the examples are static and sometimes not that impressive, so I still think there's a reason to keep the video around. Again, can be discussed in a follow-up PR.
> 
> 
> I don't want to bikeshed this PR too much because afaic this is a strict improvement on what we currently have, and we can always iterate on it in future.

👍

PR #958, opened 2026-03-18
Refresh installation guide, with focus on bundled app instructions

TimMonko:
# Description

In preparation for the Introduction workshops, I focused on updating the bundled app instructions. 
#953 starts improving the bundle install process, by directly linking to the bundle section from the homepage. 
It also adds cards to the top, now that we've consolidated the two files into the same file. These cards are a direct copy from #953 and will shop up with the blue coloring with the css from #953. I would also be ok if we did not choose the css coloring for this (`:class-card: homepage-card-accent`)

image

This improves upon #953 that by making the bundled apps completely one click to install. Rather than going through the instructions and then the links, as before.

- I reordered the bundle OS's in order of downloads according to the weather report.
- I now added to conf.py and use the full release replacement, so now dev docs should correctly link to pre-release bundles instead of a 404 to a no-existent full release

image

- This lightly edits the `Install as Python package` section to bring up to date.
- I also copy-edit the bundle instructions just to simplify the language and use bold for emphasis on instructions.
- I removed some admonitions/notes that I felt like drew too much attention when the relavant materials were mostly found in the main body.

Note, there is a **singular** code block that gets run in this file. I would be in favor of removing it (it just shows the napari viewer) in order to save on build time, but I haven't yet


psobolewskiPhD:
I broadly like the changes!
...except for the tiles at the top like that? The left on (install with Python) links to the header that is literally right below:
image

I think the admonition approach like on main is more elegant.

docs/getting_started/installation.md
```diff
@@ -16,25 +16,34 @@ kernelspec:
 
 # Install napari
 
-This guide will teach you how to do a clean install of **napari** and launch the viewer.
+`````{grid} 1 2 2 2
 
-If you're already familiar with Python, napari can be installed from PyPI or conda-forge using your favorite virtual environment and package manager. Otherwise, keep reading for instructions on how to install napari as a Python package.
```
psobolewskiPhD, line 21: I'm partial to this sentence about being able to use your favorite virtual environment and package manager. We do want the docs to be opinionated, but if someone is more advanced and has their own preferences they can go with that!

docs/getting_started/installation.md
psobolewskiPhD, line 87: no longer really relevant.
```suggestion
If you prefer to manage packages with conda, napari is available on the
conda-forge channel. You can install it with:
```

docs/getting_started/installation.md
```diff
@@ -246,188 +253,173 @@ pip install napari[pyqt6, optional] -c constraints_py3.10.txt
 
 ## Install as an application
 
-napari can be installed as a bundled app on [MacOS](#macos-bundle), [Windows](#windows-bundle), and [Linux](#linux-bundle) with a simple one click download and guided installation process. This installation method is best if you mainly want to use napari as a standalone GUI app. However, certain plugins may not be supported.
+napari can be installed as a standalone application on macOS, Windows, and Linux — no Python knowledge required. This is the simplest way to get started and works best if you want to use napari as a standalone GUI app. Note that some plugins may not be available without a full Python environment.
```
psobolewskiPhD, line 256: How about this phrasing?
```suggestion
napari can be installed as a standalone application on macOS, Windows, and Linux — no Python knowledge required. This is the simplest way to get started and works best if you want to use napari as a standalone GUI app. Note that some plugins may not be fully supported by just the GUI.
```
(very apropos I got an email today from a user confused by something from plugin docs which assumed CLI usage -- we got it working via the console, so it was a nothing burger)

TimMonko:
> I broadly like the changes! ...except for the tiles at the top like that? The left on (install with Python) links to the header that is literally right below: image
> 
> I think the admonition approach like on main is more elegant.

Appreciate the comments and suggestions! I was also unsure about the two cards at the top. My idea was to make them feel like they are of equal value, rather than with an admonition which gives the sense of "if you're not good enough for the Python env, you can try the bundle" (at least to me).
Yes, its slightly redundant from a spatial perspective, but from an information hierarchy perspective I think it holds value. 
Would like additional thoughts! Will update the PR in the morning

psobolewskiPhD:
Are they of equal value? I don't think they are to the reader. To a python person the app is useless and to a non-python person the Python stuff will be overwhelming, so giving them the clear escape hatch seems worth it.

That said, after scipy we talked about promoting the bundle (app) more or making it more prominent and it was shot down because we're Python first, so I may be off base here.

TimMonko:
> Are they of equal value? I don't think they are to the reader. To a python person the app is useless and to a non-python person the Python stuff will be overwhelming, so giving them the clear escape hatch seems worth it.

Can you explain what you mean? With the cards there is a clear escape hatch. I find adominitions to be much less eye catching, perhaps because we tend to use them a bit too much (IMO) in our docs.

> That said, after scipy we talked about promoting the bundle (app) more or making it more prominent and it was shot down because we're Python first, so I may be off base here.

I think this sentiment may have shifted a bit more after GloBIAS, but I've always been pro-bundle. It gets folks foot in the door

psobolewskiPhD:
I mean that to very few people reading this page are the two options of equal value.
So to me it makes more sense to be opinionated and pick one route -- the Python one right now -- and then provide the other route -- the bundle/app right now -- as the clear alternative.

Having two tiles like that makes sense on the homepage or in on a quick-start page or some other page where other content will continue (so the tiles are the "go install, here's two options, come back when done"). But this is already the installation page that the user actively selected, so I think having a tile that does nothing is just weird. In general I think sometimes we try to present too many options at once, each as equivalent, rather than a clear main way, with perhaps an alternative.

(I love the admonitions because I love color. In Caldeum color means wealth!)

DragaDoncila (approved):
Discussed at Pacific community meeting. We want this for the workshops, specifically the application install buttons, so we're going to merge as is. The python package section needs a lot of work so we'll do that in a follow-up PR.

TimMonko:
The cards were also discussed as ok for now, but we could consider a follow-up where the cards are on an index page that links to two separated install files.

psobolewskiPhD:
> The cards were also discussed as ok for now, but we could consider a follow-up where the cards are on an index page that links to two separated install files.

Let's not do that? I'm very happy we started to un-nest a lot of the Usage docs. Let's not put stuff behind extra pages unnecessarily, making people have to read and click more before they get to where they want.

In my world we would have the python install instructions and the app install instructions.
Homepage has a card/tab/tile/whatever for each.
And each of them would have an admonition pointing to the other one near/at the top.

DragaDoncila:
@psobolewskiPhD I think I like that suggestion

PR #973, opened 2026-03-27
Use stable version for dev docs bundle links

### LINKED ISSUE #969: `release` should (sometimes?) point to `latest` when on dev version
## References
- seen here: https://napari.zulipchat.com/#narrow/channel/212875-general/topic/Napari.20workshop/with/581611672

## 🐛 Bug
Because `release` is set to the current napari version and this value is passed to the html context [here](https://github.com/napari/docs/blob/main/docs/conf.py#L155), this results in links like this:

Image

Which go nowhere. We should make sure these links point to the latest actual release instead, or at least some real artifact.

TimMonko:

Closes #969

# Description

Instead of just the raw release link, which can be broken for dev releases (i.e. `0.7.1dev#sha`) this changes  it so that
It will be the latest stable release.

To do this, I have it check whether the current ref is a prerelease, and if it is, then it checks `version_switcher.json` for the entry with "preferred" and returns it's "version". Thus both 0.7.1rc0 and 0.7.1dev+hash will point back to the preferred version (0.7.0). 

Then, when a new stable release is pushed, the ref is no longer a pre-release and will instead use the version string (which I think in all cases is X.Y.Z from how I read the code) so releasing 0.7.1 should use 0.7.1 for the links, despite the version_switcher not yet being updated.

I also slightly updated the Mac buttons and install section so that it cleared up uncertainty re: the intro-workshop download

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -15,6 +15,7 @@ import re
 from datetime import datetime
 from importlib import import_module
 from importlib.metadata import distribution, version as metadata_version
+from pathlib import Path
 from urllib.parse import urlparse, urlunparse
 
 import napari
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -267,10 +267,6 @@ min_python_version = min(napari_supported_python_versions)
 max_python_version = max(napari_supported_python_versions)
 
 version_string = '.'.join(str(x) for x in __version_tuple__[:3])
-bundle_release = release
-if napari_version.pre and napari_version.pre[0] in {'a', 'rc'}:
-    pre_label, pre_number = napari_version.pre
-    bundle_release = f'{napari_version.base_version}{pre_label}{pre_number}'
 # when updating the version below, ensure to also update napari/napari README
 python_version = '3.11'
 python_version_range = f'{min_python_version}-{max_python_version}'
@@ -279,7 +275,6 @@ myst_substitutions = {
     'napari_conda_version': f'`napari={version_string}`',
     'napari_version': version_string,
     'release': release,
-    'bundle_release': bundle_release,
     'python_version': python_version,
     'python_version_range': python_version_range,
     'python_version_code': f'`python={python_version}`',
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -15,7 +15,6 @@ import re
 from datetime import datetime
 from importlib import import_module
 from importlib.metadata import distribution, version as metadata_version
-from pathlib import Path
 from urllib.parse import urlparse, urlunparse
 
 import napari
@@ -267,6 +266,10 @@ min_python_version = min(napari_supported_python_versions)
 max_python_version = max(napari_supported_python_versions)
 
 version_string = '.'.join(str(x) for x in __version_tuple__[:3])
+bundle_release = release
+if napari_version.pre and napari_version.pre[0] in {'a', 'rc'}:
+    pre_label, pre_number = napari_version.pre
+    bundle_release = f'{napari_version.base_version}{pre_label}{pre_number}'
 # when updating the version below, ensure to also update napari/napari README
 python_version = '3.11'
 python_version_range = f'{min_python_version}-{max_python_version}'
@@ -275,6 +278,7 @@ myst_substitutions = {
     'napari_conda_version': f'`napari={version_string}`',
     'napari_version': version_string,
     'release': release,
+    'bundle_release': bundle_release,
     'python_version': python_version,
     'python_version_range': python_version_range,
     'python_version_code': f'`python={python_version}`',
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,6 +12,7 @@ individual extension documentation for more information on specific settings.
 import logging
 import os
 import re
+import subprocess
 from datetime import datetime
 from importlib import import_module
 from importlib.metadata import distribution, version as metadata_version
@@ -267,6 +268,26 @@ min_python_version = min(napari_supported_python_versions)
 max_python_version = max(napari_supported_python_versions)
 
 version_string = '.'.join(str(x) for x in __version_tuple__[:3])
+
+# For bundle download links, always use a version that has published assets.
+# On dev builds the current version_string refers to an unreleased version,
+# so fall back to the latest stable tag from git instead.
+def _latest_stable_tag() -> str:
+    """Return the most recent vX.Y.Z git tag (no pre-release suffix)."""
+    try:
+        result = subprocess.run(
+            ['git', 'tag', '--sort=-version:refname'],
+            capture_output=True, text=True, check=True,
+        )
+        for tag in result.stdout.splitlines():
+            tag = tag.strip()
+            if re.fullmatch(r'v\d+\.\d+\.\d+', tag):
+                return tag.lstrip('v')
+    except Exception:
+        pass
+    return version_string  # fallback
+
+bundle_version = _latest_stable_tag() if napari_version.is_prerelease else version_string
 # when updating the version below, ensure to also update napari/napari README
 python_version = '3.11'
 python_version_range = f'{min_python_version}-{max_python_version}'
@@ -274,6 +295,7 @@ python_version_range = f'{min_python_version}-{max_python_version}'
 myst_substitutions = {
     'napari_conda_version': f'`napari={version_string}`',
     'napari_version': version_string,
+    'bundle_version': bundle_version,
     'release': release,
     'python_version': python_version,
     'python_version_range': python_version_range,
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -10,9 +10,9 @@ individual extension documentation for more information on specific settings.
 # -- Imports --------------------------------------------------------------
 
 import logging
+import json
 import os
 import re
-import subprocess
 from datetime import datetime
 from importlib import import_module
 from importlib.metadata import distribution, version as metadata_version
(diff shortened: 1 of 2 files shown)


docs/getting_started/installation.md
```diff
@@ -255,39 +255,39 @@ pip install napari[pyqt6, optional] -c constraints_py3.10.txt
...
 :text-align: center
 
-{{ '[Download (.exe)](https://github.com/napari/napari/releases/download/vREL/napari-REL-Windows-x86_64.exe)'.replace('REL', release) }}
+{{ '[Download (.exe)](https://github.com/napari/napari/releases/download/vREL/napari-REL-Windows-x86_64.exe)'.replace('REL', bundle_release) }}
 +++
 {ref}`Installation steps <windows-bundle>`
 ```
 
-```{grid-item-card} macOS Apple Silicon (arm64)
+```{grid-item-card} macOS Apple Silicon (arm64) —  M1 and later
```
psobolewskiPhD: Hmm, this isn't correct for the new MacBook which is Apple Silicon but not "M1 and later"
Maybe we can link apple docs?
https://support.apple.com/en-us/116943

docs/getting_started/installation.md
```diff
@@ -338,8 +338,8 @@ Next, check out our [tutorial on the viewer](viewer-tutorial) or explore the [Us
 
 Download the installer for your Mac:
 
-- {{ '[napari-REL-macOS-arm64.pkg](https://github.com/napari/napari/releases/download/vREL/napari-REL-macOS-arm64.pkg)'.replace('REL', release) }} — Apple Silicon (M1 and later)
-- {{ '[napari-REL-macOS-x86_64.pkg](https://github.com/napari/napari/releases/download/vREL/napari-REL-macOS-x86_64.pkg)'.replace('REL', release) }} — Intel
+- {{ '[napari-REL-macOS-arm64.pkg](https://github.com/napari/napari/releases/download/vREL/napari-REL-macOS-arm64.pkg)'.replace('REL', bundle_release) }} — Apple Silicon (M1 and later)
```
TimMonko: @psobolewskiPhD here also is where I mention arm64 and Apple Silicon. 

I've never used mac so I genuinely don't know the correct answer here.

docs/getting_started/installation.md
TimMonko: is new new MacBook not M2? I genuinely don't know anything about mac architecture so I'd be happy to take sgugestiosn from you.

Basically, a user tried to install for mac but chose the Intel bundle when they had an M2 mac

psobolewskiPhD: M2 is from 2022
The new MacBook Neo uses an A-series processor similar to some of the apple iPhones and tablets.

Not sure why anyone would pick Intel for anything these days 😜 🥁

docs/getting_started/installation.md
psobolewskiPhD: Yeah same issue. I think linking the apple docs is probably the best, but otherwise we could try
`(M- or A-series)` or something like that.

docs/getting_started/installation.md
psobolewskiPhD: Could say "(2020 or newer)" that's not quite 100% correct either, but probably works for the vast majority of cases and makes it really easy to understand.

Czaki: Maybe: "macOS ARM" or "Apple ARM"

docs/conf.py
```diff
@@ -267,6 +267,10 @@ def get_supported_python_versions(project_name):
 max_python_version = max(napari_supported_python_versions)
 
 version_string = '.'.join(str(x) for x in __version_tuple__[:3])
+bundle_release = release
+if napari_version.pre and napari_version.pre[0] in {'a', 'rc'}:
+    pre_label, pre_number = napari_version.pre
+    bundle_release = f'{napari_version.base_version}{pre_label}{pre_number}'
```
Czaki: For me this logic is not fixing pointed issue. 

I think that we need to use function like this:

```python
def get_latest_released_version(version_: packaging.version.Version) -> str:
    version_ = version_.__replace__(local=None)
    if not version_.is_devrelease:
        return str(version_)
    if version_.pre is None:
        # we are on path with no pre-release
        assert version_.micro > 0, "dev release must have micro > 0"
        return f'{version_.major}.{version_.minor}.{version_.micro - 1}'

    # we are on path with pre-release
    pre_label, pre_number = version_.pre
    return str(version_.__replace__(pre=(pre_label, pre_number - 1), dev=None))
```

docs/getting_started/installation.md
willingc:
```suggestion
```{grid-item-card} macOS Apple Silicon (ARM)
```

docs/getting_started/installation.md
```diff
@@ -300,7 +300,7 @@ For earlier versions of the napari app, scroll below the latest release on the [
 
 ### Windows
 
-Download from Github: {{ '[napari-REL-Windows-x86_64.exe](https://github.com/napari/napari/releases/download/vREL/napari-REL-Windows-x86_64.exe)'.replace('REL', release) }}
+Download from Github: {{ '[napari-REL-Windows-x86_64.exe](https://github.com/napari/napari/releases/download/vREL/napari-REL-Windows-x86_64.exe)'.replace('REL', bundle_release) }}
```
willingc:
```suggestion
Download from GitHub: {{ '[napari-REL-Windows-x86_64.exe](https://github.com/napari/napari/releases/download/vREL/napari-REL-Windows-x86_64.exe)'.replace('REL', bundle_release) }}
```

docs/getting_started/installation.md
willingc:
```suggestion
- {{ '[napari-REL-macOS-arm64.pkg](https://github.com/napari/napari/releases/download/vREL/napari-REL-macOS-arm64.pkg)'.replace('REL', bundle_release) }} — Apple Silicon (ARM)
```

docs/getting_started/installation.md
```diff
@@ -383,7 +383,7 @@ Next, check out our [tutorial on the viewer](viewer-tutorial) or explore the [Us
 
 ### Linux
 
-Download from Github: {{ '[napari-REL-Linux-x86_64.sh](https://github.com/napari/napari/releases/download/vREL/napari-REL-Linux-x86_64.sh)'.replace('REL', release) }}
+Download from Github: {{ '[napari-REL-Linux-x86_64.sh](https://github.com/napari/napari/releases/download/vREL/napari-REL-Linux-x86_64.sh)'.replace('REL', bundle_release) }}
```
willingc:
```suggestion
Download from GitHub: {{ '[napari-REL-Linux-x86_64.sh](https://github.com/napari/napari/releases/download/vREL/napari-REL-Linux-x86_64.sh)'.replace('REL', bundle_release) }}
```

willingc (commented):
Thanks @TimMonko

Czaki:
https://napari.zulipchat.com/#narrow/channel/585967-napari-workshop-2026-04-01/topic/Installation.20Questions.20and.20Troubleshooting/near/583551942

willingc:
@TimMonko A general question. How would you feel about dev docs displaying the stable released bundle? Perhaps adding a comment that pre-release bundles may be found in GitHub releases page.

It may be less confusing for readers.

TimMonko:
> @TimMonko A general question. How would you feel about dev docs displaying the stable released bundle? Perhaps adding a comment that pre-release bundles may be found in GitHub releases page.
> 
> It may be less confusing for readers.

I switched it back to that, which is what Peter had implemented prior to my #958 . 
There is, at L297 a reference to the releases page already :)
This should be ready to go now :)

Czaki (dismissed):
Image

There is no 0.7.1 release yet. 

Links should lead to 0.7.0 not 0.7.1

TimMonko:
Woah thanks for looking at the build @Czaki , because locally I saw it correctly: 
image

This did make me totally switch up the logic to what I think will work, but definitely need careful thought. I'll update PR description.

I also did add mention about pre-releases on the releases page on github, and not just earlier builds.

Edit:
CI Looks good
image

docs/getting_started/installation.md
TimMonko: I  added a link to the documentation in the mac section

docs/conf.py
```diff
@@ -267,13 +268,29 @@ def get_supported_python_versions(project_name):
...
+    switcher_path = Path(__file__).parent / '_static' / 'version_switcher.json'
+    switcher_entries = json.loads(switcher_path.read_text(encoding='utf-8'))
+    return next(
+        entry['version']
+        for entry in switcher_entries if entry.get('preferred')
+    )
+
+bundle_version = (
+    _stable_version_from_switcher()
+    if napari_version.is_prerelease
```
Czaki, line 283:
```suggestion
    if napari_version.is_devrelease
```

`napari==0.7.1a2` is prerelease, but not devrelease

TimMonko: I think with this current code we want 0.7.1a2 to point to 0.7.0, not the pre-release bundle (even though that was the original goal), so we want to use `is_prerelease`. 

But you are right, if we do `is_devrelease` than I think 0.7.1a2 would point to the 0.7.1a2 bundle.

Can you confirm this logic?

willingc: @TimMonko I believe that if you want to use the stable version to avoid confusion for end users (i.e. dev docs version will always display the stable bundle) then we should be using bundle_version is `stable_version = _stable_version_from_switcher()`.

TimMonko: @willingc can you check the logic I have in the description? https://github.com/napari/docs/pull/973#issue-4155548101
A new stable release would not have the version_switcher updated, so if we did just that then it would be one version behind on each stable release. 
But, admittedly, my brain is struggling to wrap up the logic

Czaki: You are right, the getting version from switcher will provide wrong version during deploy stable docs. 

And there are two paths. User is opening `napari.org/dev` then we might want to provide stable release. 

But when user enter `napari.org/0.7.1` when `0.7.1` is not released yet, but have some alpha or beta then it should link to latest bundle, even alpha one.

willingc (approved):

docs/conf.py
TimMonko: We have decided in today's docs meeting that we want the dev docs to always point to the latest stable release, not alphas or rcs.

PR #976, opened 2026-04-01
Remove mention of unmaintained plugin in quick start

### LINKED ISSUE #974: Do not recommend unmaintained plugin?
The `napari-aicsimageio` plugin repo has been archived : https://github.com/AllenCellModeling/napari-aicsimageio, but is still mentioned in the docs:
https://github.com/napari/docs/blob/57a2f8d3f5b066cab5f764472b75120fcd423f2a/docs/getting_started/quick_start.md?plain=1#L82

cnstt:

Closes #974 

# Description

Remove mention of unmaintained plugin in quick start. Fixes #974 for now. If you have any idea of a replacement plugin we could suggest, feel free to mention it!

### DIFF
--- a/docs/getting_started/quick_start.md
+++ b/docs/getting_started/quick_start.md
@@ -79,7 +79,6 @@ napari natively supports tiff and many other formats supported by [skimage.io.im
 Try with your own images or download [this ome tiff file](https://downloads.openmicroscopy.org/images/OME-TIFF/2016-06/MitoCheck/00001_01.ome.tiff).
 
 Additional input file formats may be supported [by plugins](https://www.napari-hub.org/).
-Try [napari-aicsimageio](https://www.napari-hub.org/plugins/napari-aicsimageio) if you have czi, lif, or nd2 files.
 
 Once you have the proper plugin installed, use File > Open Files(s)
 and select the image file, or simply drag and drop the image into napari.


TimMonko:
Thank YOU! I couldn't find the notification for #974, I think I accidentally cleared it. Anyways, in the open images guide we recently added / approved using `ndevio` which is a successor to napari-aicsimageio. I develop it so I don't want to bias...
Here is where it was introduced https://github.com/napari/docs/blob/57a2f8d3f5b066cab5f764472b75120fcd423f2a/docs/getting_started/open_images.md?plain=1#L130-L136

psobolewskiPhD:
https://github.com/imaging-formats/napari-bioformats
is another good option for a Fiji-like experience

cnstt:
Thanks a lot for your recommendations, it's exactly what I was missing! I don't have enough experience with those plugins, to know which one is best to recommend in this case ;)
Or perhaps we could also just link to `getting_started/open_images.html#using-a-plugin-reader`, but it might be a long read for someone just looking for a quick way to open their files.🤔 Maybe the best would be to recommend one that works, and mention the specific docs section for further information.

TimMonko:
I actually quite like the idea for linking to the documentation.
1. We don't have to maintain two places (like we failed to here)
2. It gives more detail on what is actualyl going on, especially if someone is in quickstart they don't even really necessarily know what a plugin is yet

Something like "To read additional file formats (e.g. .lif, .czi, .nd2), read more about using plugin readers in the [open images guide](link)"

I'm not super keen on that language so hopefully you can figure it out :)

TimMonko:
Also @cnstt since I feel like I have no better place else to put it. I always appreciate seeing notifications from you. You do very nice clean up and maintenance work around things. Don't be afraid to pop into the community meetings or zulip chat -- would love to meet you (unless I already have??)! -- and maybe pull you even farther into the project ;)

willingc (approved):
Thanks for fixing this up @cnstt

Issue #981, opened 2026-04-09
Consider how to maintain pixi.lock better for local dev


TimMonko:
## 🧰 Task

I have run twice now into problems developing for the docs where I continually use pixi tasks (usually `pixi run slimfast-live`) to build.

However, it's been twice now that dependencies need updated such taht they are working either on CI or in a different but similar repo (in this case it was napari-sphinx-theme) but not locally. The solution has been `pixi update` to bring in newer versions of packages.

Perhaps we should be committing the lockfile to this repo, so that pulls from main give the most up to date lockfile. I do always think its a chore to have the diffs show up in PRs though, so I'm not sure if people are maintaining lockfiles via CI? Generally good stuff to think about.

On CI this isn't an issue currently because it uses constraints. So this causes divergence between CI and local as well.

TimMonko:
@willingc and @Aniketsy say we should commit the lock file 😁 

@melissawm, @psobolewskiPhD, and @rensutheart  do you have any thoughts?

melissawm:
I feel like @jaimergp is the right person to have this conversation with 😄

Czaki:
Did you try to use `UV_CONSTRAINTS` environment variable that will point to main repo docs constraints?

jaimergp:
This may be useful if you end up submitting the lock file: https://github.com/Quansight-Labs/conda-metadata-app/blob/main/.github/workflows/update-lockfiles.yml. Example PR: https://github.com/Quansight-Labs/conda-metadata-app/pull/65

willingc:
@Czaki Is https://github.com/napari/napari/blob/main/resources/constraints/constraints_py3.12_docs.txt the file that would be used to set `UV_CONSTRAINT`?

A few questions:

1. Does UV_CONSTRAINT work if you are doing local development with Python 3.13 or 3.14?
2. Does UV_CONSTRAINT only work with uv or can pixi pull it in too?

willingc:
It doesn't look like the current `pixi.toml` is using the docs constraints for local builds. I see 3.13 as the Python version in the `pixi.toml`. Would it make sense to make the `pixi.toml` consistent with the docs constraints file?

Czaki:
> Does UV_CONSTRAINT work if you are doing local development with Python 3.13 or 3.14?

It depends. If the latest version of package that support python 3.12 is supporting also python 3.13 and 3.14 it should work.

[...]

PR #982, opened 2026-04-09
Add "Edit on Github" to secondary sidebar

TimMonko:

xref https://github.com/napari/napari-sphinx-theme/pull/216

# Description

This adds an "Edit this Page" button to the secondary (right) sidebar on our docs. 
Minimal setup was done via here: https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/source-buttons.html

And `edit_page_url_template` points not to the `edit` URL but instead to `blob`. I noticed from pydantic sphinx theme that clicking through leads me to the "You need to fork this" and not a nice clean document. I think being met with the fork screen would be a WTF moment as a new contributor, so I'd prefer something more welcoming (and where you can actually see the code right away)

Also ran into #981 while developing

image

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -126,7 +126,8 @@ html_theme_options = {
     'show_version_warning_banner': True,
     'navbar_persistent': ['search-button'],
     'header_links_before_dropdown': 6,
-    'secondary_sidebar_items': ['page-toc'],
+    'use_edit_page_button': True,
+    'secondary_sidebar_items': ['page-toc', 'edit-this-page'],
     'pygments_light_style': 'napari',
     'pygments_dark_style': 'dracula',
     'announcement': '',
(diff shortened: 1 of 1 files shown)


brisvag:
Can it also be renamed to "propose an edit" or something like that? Might be less intimidating, as "edit on github" might sound like it's off-limits to the casual visitor.

TimMonko:
Hmm, at first my counter argument was that this language is "standardized", but I think you are on to something. 

To do this, it does require a customized html template. If I do this I will add it to napari-sphinx-theme, so all repos benefit. So, in that sense this PR can be merged and then the upstream napari-sphinx-theme will override it later.

willingc (approved):

TimMonko:
We collectively decided at the docs meeting, that it should remain "Edit on Github" since that is a standard. @brisvag if you feel strongly you can open an issue in napari-sphinx-theme ❤️

Issue #979, opened 2026-04-08
Add search-as-you-type


TimMonko:
## 🧰 Task

Follow-up to napari/napari-sphinx-theme#217

1. release napari-sphinx-theme
2. update pin in napari/napari
3. do the changes modeled in napari/napari-sphinx-theme#217 to expose in napari/docs

PR #991, opened 2026-04-16
Use flexible search field for navbar

TimMonko:

Follow-up to #979 

Will only be styled as below once napari/docs constraints get updated. (should we bump min pydata-sphinx-theme in napari/napari?)

# Description

Uses pydata-sphinx-theme's flexible search-field -> search button width-dependent navbar search. 
You can check https://napari.org/napari-sphinx-theme to see it in action, for now

image

image


docs/conf.py
```diff
@@ -125,7 +125,6 @@
         'version_match': version_match,
     },
     'show_version_warning_banner': True,
-    'navbar_persistent': ['search-button'],
```
TimMonko, line 125: interestingly, using `['search-field']` here seems to pull in a different html template, so it works better by just removing this and I guess defaulting to pydata-sphinx-theme's route. I _thought_ that was the flexible one, but the docs are unclear. Just leaving a breadcrumb here

https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/search.html

TimMonko:
@napari-bot make slimfast

the constraints have been bumped so it should have the new search field

TimMonko:
Hmm doesn't look like circle pulled in the updated napari-sphinx-theme yet

TimMonko:
Yay it's showing now after merging in a commit from main :) There is a small section of width where the search field dissapears and the search icon is not present. This does not seem to be the case for [pydata-sphinx-theme](https://pydata-sphinx-theme.readthedocs.io/en/stable/)

I suspect it has something to do with https://github.com/napari/napari-sphinx-theme/pull/198

image

brisvag (approved):

TimMonko:
After napari/napari#8949 updated constraints, this should be working and ready

TimMonko:
This dynamic search field re-introduces a wonky navbar at a small range of widths, that are not "typical" (i.e its search field at full width 1080p, and search icon at half width both without wonkiness), but because its the right side items, it's not as bothersome to me. For the accessibility benefit this provides, it's worth it. The main fix would be to left-align the navbar links rather than center align, but I'm not the biggest fan of that stylistically.

image

PR #993, opened 2026-04-20
Add info about color and intensity to surface guide

TimMonko:

depends on napari/napari#8909

# Description

Adds information about how colors are handled in surface layers, especially as it relates to contrast and colormaps.
Adds info relevant to napari/napari#8909 where color controls are disabled when `vertex_colors` is passed. 
I tried also to clarify how omitting the vertex values resulted in 1s being used to color the surface, resulting in "white" and thus a bit of magic (to me) seemed to be involved in contrast controls and such. I still don't feel like I got the words out correctly.


docs/howtos/layers/surface.md
```diff
@@ -33,9 +33,10 @@ layer first, then explore the GUI controls.
 
 The surface layer allows you to display a precomputed surface mesh that is
 defined by an `NxD` array of `N` vertices in `D` coordinates, an `Mx3` integer
-array of the indices of the triangles making up the faces of the surface, and a
-length `N` list of values to associate with each vertex to use alongside a
-colormap.
+array of the indices of the triangles making up the faces of the surface, and,
```
willingc: Since there are several long components and one optional, changing to a bulleted list would reduce cognitive load for reader.

docs/howtos/layers/surface.md
```diff
@@ -79,6 +80,13 @@ viewer.close()
 Once you have created a `surface` layer programmatically, the following GUI
 controls are available in the viewer:
 
+Controls related to `contrast limits`, `auto-contrast`, `gamma`, and
+`colormap` are only available when the surface color is computed from
+`vertex_values`. If you provide `vertex_colors`, napari uses those colors
```
willingc, line 83:
```suggestion
When the surface color is computed from `vertex_values`, controls related
to `contrast limits`, `auto-contrast`, `gamma`, and
`colormap` are unavailable. If you provide `vertex_colors`, napari uses those colors
```

Is vertex_colors a vertex_value?

docs/howtos/layers/surface.md
```diff
@@ -116,12 +124,27 @@ help(napari.Viewer.add_surface)
...
-list of values to associate with each vertex to use alongside a colormap. This
-3-tuple is accessible through the `layer.data` property.
+The data for a `surface` layer can be given as either a 2-tuple
+`(vertices, faces)` or a 3-tuple `(vertices, faces, vertex_values)`. The
+vertices are an `NxD` array of `N` vertices in `D` coordinates. The faces are
+an `Mx3` integer array of the indices of the triangles making up the faces of
+the surface. The optional vertex values are a length `N` list of scalar values
+to associate with each vertex for colormap-based rendering. If you omit
+`vertex_values`, napari fills them with ones, so the default underlying surface
+color is white.
```
willingc: Consider a bullet list for sentences after the first.

```suggestion
The data for a `surface` layer can be given as either a 2-tuple
`(vertices, faces)` or a 3-tuple `(vertices, faces, vertex_values)`.
- The vertices are an `NxD` array of `N` vertices in `D` coordinates. 
- The faces are an `Mx3` integer array of the indices of the triangles
  making up the faces of the surface. 
- The optional vertex values are a length `N` list of scalar values
  to associate with each vertex for colormap-based rendering. 
  If you omit `vertex_values`, napari fills them with ones, so the default
  underlying surface color is white.
```

docs/howtos/layers/surface.md
TimMonko: so this is IMO a bigger problem than I can address in the PR. 
`vertex_values` is really the 3rd element of the data tuple, but it seems to be used throughout the docs and examples as if it were a real argument.
On the other hand, `vertex_colors` is an argument to `Surface` that overrides anything set in that third element of the data tuple. 
I'll take another overall swing at the wording -- I think bulleting the tuple elements like above will probably help with this :)

willingc: No problem. I just didn't know the answer. Keep it simple now with just the bullets.

willingc (approved):

docs/howtos/layers/surface.md
```diff
@@ -33,9 +33,10 @@ layer first, then explore the GUI controls.
 
 The surface layer allows you to display a precomputed surface mesh that is
 defined by an `NxD` array of `N` vertices in `D` coordinates, an `Mx3` integer
-array of the indices of the triangles making up the faces of the surface, and a
-length `N` list of values to associate with each vertex to use alongside a
-colormap.
+array of the indices of the triangles making up the faces of the surface, and,
+optionally, a length `N` list of values to associate with each vertex to use
+alongside a colormap. If you omit the vertex values, napari uses a length `N`
+array of ones.
```
brisvag: Actually, surfaces values can be n-dimensional:

https://github.com/napari/napari/blob/93ef1fbc8de2758fd1b381a5ef7e5508323547a4/src/napari/layers/surface/surface.py#L55-L58

docs/howtos/layers/surface.md
```diff
@@ -116,12 +124,27 @@ help(napari.Viewer.add_surface)
...
-and vertex values. The vertices are an `NxD` array of `N` vertices in `D`
-coordinates. The faces are an `Mx3` integer array of the indices of the
-triangles making up the faces of the surface. The vertex values are a length `N`
-list of values to associate with each vertex to use alongside a colormap. This
-3-tuple is accessible through the `layer.data` property.
+The data for a `surface` layer can be given as either a 2-tuple
+`(vertices, faces)` or a 3-tuple `(vertices, faces, vertex_values)`. The
+vertices are an `NxD` array of `N` vertices in `D` coordinates. The faces are
+an `Mx3` integer array of the indices of the triangles making up the faces of
+the surface. The optional vertex values are a length `N` list of scalar values
```
brisvag: same here as above.

docs/howtos/layers/surface.md
TimMonko: Can you please provide a suggestion :) Reading the docstring is not bringing clarity to us in the docs meeting 💔 We should probably update that docstring too

TimMonko: RIP Me: 
image

Can you please provide a suggestion :) Reading the docstring is not bringing clarity to us in the docs meeting 💔 We should probably update that docstring too

brisvag: Ouch, sorry about this one! I'll come up with something now.

brisvag:
```suggestion
array of the indices of the triangles making up the faces of the surface, and,
optionally, a length `N` array of per-vertex values that can be used alongside a colormap to color the surface. Vertex values can also have additional leading dimensions (to represent, for example, a time series), in the form `(D1, D2, ..., N), which will be treated as extra dimensions by the viewer.
```

brisvag (approved):
Added suggestion. To not make this too messy, I only wrote it there, but I think @willingc's bullet-point idea should be done in both places, and then my text should be added to both!

brisvag (approved):

PR #997, opened 2026-04-21
Update deprecation warning guidance to use `FutureWarning`

### LINKED ISSUE #983: Add `FutureWarning` as preferred deprecation method to contributing guides
See: for how we actually want to do this

Future warning as deprecation warning is invisible for users (python silence it by default)

_Originally posted by @Czaki in https://github.com/napari/napari/pull/8852#discussion_r3052404244_
            
For now, I think the easiest place to add it will be in this section https://napari.org/stable/developers/contributing/index.html#api
Because it discusses `.. versionadded::` so its related in the sense of docstrings. But we should clarify how it goes in both the docstring and in the code (with `stacklevel=2`) and how `FutureWarning` should not be used

cc: @aniketsy -- let me know if you need any assistance while contributing to the docs ❤️

TimMonko:

Closes #983

# Description

We want to use `FutureWarning` because it is not silenced by Python

### DIFF
--- a/docs/developers/contributing/index.md
+++ b/docs/developers/contributing/index.md
@@ -147,9 +147,10 @@ Please also consider documenting any major features/changes in our
 
 #### Deprecation Warnings
 
-When deprecating a feature, use `DeprecationWarning` instead of `FutureWarning`.
-`FutureWarning` is silenced by Python's default warning filters, making it invisible
-to library users. `DeprecationWarning` is the correct signal for developer-facing deprecations.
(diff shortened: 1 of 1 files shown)


TimMonko:
cc: @Aniketsy, as I misled you and I'd like us both to learn 😄

Aniketsy:
>  I'd like us both to learn 😄

yes 😁 , thanks! for fixing, I was about to ask like, should i fix but then i saw, your PR 😊

melissawm (approved):

PR #1008, opened 2026-05-05
Remove "auto author assign" workflow

TimMonko:
# Description

Getting a ping a few seconds after I make a PR has consistently bothered me.
This removes the auto author assign action.

@melissawm you are the commit for its addition, could you share context on why you added it?
Don't want to remove something that is useful for others, but it is noise for me.


TimMonko:
Why did the action still run??? I thought actions use what is in the PR, not main?
image

melissawm:
Looking at the commit, I think this was just the result of us copying the docs over from the main repo and the action was already there. I did not create it and also don't like it 😅

Czaki:
> Why did the action still run??? I thought actions use what is in the PR, not main? image

The `pull_request_target` is executed based on `main` not on `pull_request` So it is correct that is it runs.

brisvag (approved):

PR #1014, opened 2026-05-11
Update circle ci workflow with PyQt6 and `docs` dependency group

TimMonko:

napari/napari#8704

# Description

1. Uses `docs` PEP735 dependency group for Circle CI config
2. Uses pyqt6 instead of pyqt5 for docs build, to bring in line with the napari workflows


TimMonko:
superseded by #1104

Issue #852, opened 2025-09-25
Consider updating the examples gallery to suggest using `napari <url of script>` to run


psobolewskiPhD:
## 🧰 Task
Motivated by #847

jni:
I'm anxious about this because it increases the distance between "using napari" and "just using Python".

The remote execution is cool and useful but also a risky habit to get into…

Issue #678, opened 2025-04-24
Add better instructions for running examples, e.g. to the Gallery landing page


psobolewskiPhD:
## 🧰 Task

For a new comer, it's not really so obvious I don't think how to use the examples in the Gallery.
The landing page says:

> All examples in this gallery can be downloaded as Python scripts or Jupyter notebooks to be executed locally. Check out [Jupyter notebook usage](https://napari.org/stable/tutorials/fundamentals/getting_started.html#launch-jupyter) for more details on using napari in Jupyter notebooks.

But actually the notebooks from the examples are kinda meh and they really are meant to run as scripts, which isn't described.

In the Quick Start
https://napari.org/stable/tutorials/fundamentals/quick_start.html

It says
> Test some [examples](https://github.com/napari/napari/tree/main/examples) to see how to add different layer types and add your own widgets to napari.

(note the link should be fixed to Gallery) But that doesn't help at all.

The "How to launch guide" does describe how to run a script:
https://napari.org/stable/tutorials/fundamentals/getting_started.html#python-script-usage
So at the very least we should link to that from the Gallery and maybe mention a Gallery example there?

The Viewer tutorial doesn't mention running Examples at all (and the https://napari.org/stable/tutorials/fundamentals/viewer.html#launching-the-viewer section should probably link to the How to launch guide https://napari.org/stable/tutorials/fundamentals/getting_started.html )

psobolewskiPhD:
Now they can be run by drag and drop, but you still need to download the script...

Issue #540, opened 2024-12-17
Binder runnable gallery notebooks


melissawm:
Would be really cool to have binder runnable gallery notebooks!
Related:
https://github.com/napari/napari/issues/7331

_Originally posted by @psobolewskiPhD in https://github.com/napari/docs/issues/422#issuecomment-2436337387_

PR #1021, opened 2026-05-19
Move download links to top of examples

TimMonko:
# References

During the pilot napari workshops, we observed new users having difficulty in navigating to the download buttons because of their location within the page.

Built on #1026. two-dot diff: https://github.com/TimMonko/docs/compare/shared-gallery-button-skin..TimMonko:docs:gallery-links

# Description

This PR moves the download buttons to the top of the example by completely overriding the gallery insertion buttons and making our own button section which is shoved into the sphinx-gallery ready header.

To turn off the sphinx-gallery buttons we use css https://sphinx-gallery.github.io/stable/advanced.html#using-sphinx-gallery-sidebar-components

Perhaps this should be upstreamed to sphinx-gallery? It might be easier, but perhaps people aren't trying to have the download buttons be so obvious as we are, due to our drag and drop.

Present PR:

image

Original This PR:

image


DragaDoncila (approved):
Much prefer having the buttons on the top 🤩

psobolewskiPhD:
ehh...
I dunno, i don't love it.
Can we at least have the description above the buttons?

I think first we should make clearer what to do with these links somewhere, e.g. address:
- https://github.com/napari/docs/issues/852
- https://github.com/napari/docs/issues/678
or even
- https://github.com/napari/docs/issues/540

Not a blocker, just my opinion -- i might be missing some context, so feel free to ignore.

Czaki:
Some conclusion? Should it land in 0.7.1

psobolewskiPhD:
I took another look and to me the layout still looks wierd:

Title
`<buttons>`
Description
tags

I think it would look more purposeful, if the buttons were below the description or tags.
🤷

DragaDoncila:
I think this should go into 0.7.1. It's true we can address a few other points to make running our examples easier, but I do think this is a strict improvement, and we know from users at the workshops that the old buttons were lost on the page. Happy to improve in future PRs as well!

DragaDoncila:
> I think it would look more purposeful, if the buttons were below the description or tags.

I can agree with this though. @TimMonko can we see what that looks like :pray:

TimMonko:
I was a bit salty about trying to move it below the description, because I thought I had tried to troubleshoot that already. Turned out I missed the very useful "GENERATED FROM PYTHON SOURCE LINES" that exists after the tags and just before the script itself. I incorrectly thought that `gen_rest.EXAMPLE_HEADER` was the only repeated code present in the built .rst's.

So, I've moved the buttons below the description/tags and have updated the OP to reflect the change. I hope it's enough, because it's important to me that this is in the upcoming release so that it's ready for our July tutorials (in case following release is delayed)

psobolewskiPhD:
The new look (top screenshot in OP) looks much better, for me anyways.
Thanks!

DragaDoncila:
Appreciate you playing with this and getting it below the description @TimMonko! I think it ended up looking lovely :heart:

PR #1024, opened 2026-05-19
Add admonition to gallery examples that require additional packages

TimMonko:

Similar reasoning as #1021. When doing drag and drop from the gallery _and_ now that examples are growing in their complexity and dependencies, it's important for us to warn users that they may not have all the dependencies to run the example. It's _especially_ important now that we feature homepage examples that do not have typical napari dependencies

Built on #1021 (and #1026). two-dot diff https://github.com/TimMonko/docs/compare/gallery-links..TimMonko:docs:gallery-deps

# Description

This adds an admonition _if_ examples require dependencies not in "typical" napari installs (bundle, napari[all]). It identifies the missing dependencie(s) adds to the instructions, and links to other resources to install.

This uses mostly llm generated logic to compare napari[gallery] dependencies against napari[all] dependencies. It requires extra logic because we build against a remote, so we can't just special case anything in napari[gallery].

Note that pooch is _not_ a direct dependency of napari (it comes from `skimage[data]` for the sample data), I had to special case ignore it.

Updated PR:

image

Original PR look:
image


docs/conf.py
```diff
@@ -396,22 +405,184 @@ def napari_scraper(block, block_vars, gallery_conf):
...
+def _gallery_extra_deps_note(extra_deps: list[str]) -> str:
+    deps = ', '.join(f'``{package}``' for package in extra_deps)
+    return f"""
+.. admonition:: Extra packages required
+
+   This example requires additional packages that are not available in typical
+   napari installations.
+   For this example to work in :doc:`recommended napari installations <{NAPARI_INSTALL_GUIDE}>`,
+   you will need to install additional packages: {deps}.
+   See the :doc:`the plugin installation guide <{PLUGIN_INSTALL_GUIDE}>`
```
DragaDoncila: Discussed at docs meeting, we think this might still be confusing for users because the "plugin installation guide" doesn't really talk about additional/other non-plugin packages, so it may not be clear that you can install them the same way as plugins.

Let's add a short section to the bottom of the finding-and-installing-plugins guide that says how you can install additional packages via the plugin manager (even if it's a bit redundant with the plugins bit). Then link to that from here.

psobolewskiPhD: Yeah, linking the plugin install guide feels off.
I think we'd want a page in the plugin manager docs on using the plugin installer to install arbitrary packages?

psobolewskiPhD:
I noted this in the other PR too, but I really think this should be below the description. 
I dislike how this distracts from the actual example!

Maybe the proper order is:
title
description
image
link buttons
code
?

Czaki:
I remove it form milestone. If you gen into conclusion, it might still get into release

docs/conf.py
TimMonko: I made napari/napari-plugin-manager#213 to address this and point the link now to that resource

TimMonko:
Moved it below description like in #1021, and updated the admonition to address the comments. Hopefully this is favorable now! I would really like to see this in 0.7.1 :)

image

psobolewskiPhD:
The screenshot posted above looks a lot better to me.
Thanks!
I also approved the napari-plugin-manager PR.

melissawm:
I LOVE the look of the buttons and admonition, also I agree with the position as it is right now (below the description). I'll approve but note that I think we should follow-up to clean up the dependency comparison code in conf.py - maybe a new contributor wants to help us there? 😄

melissawm (approved):

PR #1025, opened 2026-05-19
Document layer locking in napari GUI and API

### LINKED ISSUE #1017: Document locking layer feature
## 🧰 Task

Document the result of this PR: 

https://github.com/napari/napari/pull/8736

TimMonko:

Depends on napari/napari#8736
Closes #1017 

# Description

Adds information to the layers guide about layer locking with a small blurb to
reinforce that the lock does not prevent changes to other layer properties,
since this is much requested.

### DIFF
--- a/docs/getting_started/layers.md
+++ b/docs/getting_started/layers.md
@@ -73,7 +73,7 @@ You can lock or unlock selected layers from the layer list by right-clicking and
 choosing **Toggle lock**, or in Python by setting `layer.locked`:
 
 ```python
-layer.locked = True
+layer.locked = True  # or a napari.layers.base.LayerLock enum value
 
 # unlock again
 layer.locked = False
--- a/docs/getting_started/layers.md
+++ b/docs/getting_started/layers.md
@@ -84,10 +84,9 @@ with the napari UI. Actions that delete the original layer as part of their work
 are also disabled, including converting between image and labels layers and
 splitting or merging image stacks.
 
-Locking currently applies to these destructive layer-list operations. It does
-not make the layer data read-only or disable other non-destructive layer
-controls. You may still programmatically interact
-with the layer regardless of the lock state.
+Locking currently applies only to these destructive layer-list operations. It does
+not make the layer data read-only or disable other non-destructive changes, such as use of the layer
+controls. Finally, you may still programmatically interact with the layer regardless of the lock state.
 
 (layer_opacity)=
 
--- a/docs/getting_started/layers.md
+++ b/docs/getting_started/layers.md
@@ -64,6 +64,30 @@ button. Note that you can Option/Alt-click on the `visibility` button to show
 `visibility` button of a layer a second time, the visibility state of all layers
 will be restored.
 
+## Layer locking
+
+All our layers support a `locked` property that protects a layer from
+destructive layer-list actions.
+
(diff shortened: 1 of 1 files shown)


docs/getting_started/layers.md
```diff
@@ -64,6 +64,30 @@ button. Note that you can Option/Alt-click on the `visibility` button to show
...
+
+You can lock or unlock selected layers from the layer list by right-clicking and
+choosing **Toggle lock**, or in Python by setting `layer.locked`:
+
+```python
+layer.locked = True
+
+# unlock again
+layer.locked = False
+```
```
Czaki, line 80: You suggest here using lock instead of enum. There is a bool fallback, but from code we might want to suggest more verbose API.

TimMonko: Understandable as a suggestion. I think I would prefer to see that elsewhere in the docs or in the example gallery.
This is in the "Getting started" section and for new/novice pythonistas, Enums are still opaque. I would prefer that in the getting started guides we use the simpler API. In addition LayerLock is still in a private module

I took this mostly from the docstrings in base.py https://github.com/napari/napari/pull/8736/changes#diff-e1159e82aa9035d4bc79b72d6cb2c5cdcc2af42493edc51755c71ef5c26c0587R995-R1000

What about this, which gives enough breadcrumbs for an experienced dev to look up `LayerLock`, but perhaps is ignored by a novice? I still think that I would rather this be in the examples gallery or something

```suggestion
```python
layer.locked = True  # or a LayerLock enum value

# unlock again
layer.locked = False
```
```

Comment (Czaki) [reply docs/getting_started/layers.md:80]: Rather full path to enum, to all
````suggestion
```python
layer.locked = True  # or a napari.layers.base.LayerLock enum value

# unlock again
layer.locked = False
```
````

File: docs/getting_started/layers.md [hunk 2]
```diff
@@ -64,6 +64,31 @@ button. Note that you can Option/Alt-click on the `visibility` button to show
...
+
+When a layer is locked, you will not be able to delete it from the layer list
+with the napari UI. Actions that delete the original layer as part of their workflow
+are also disabled, including converting between image and labels layers and
+splitting or merging image stacks.
+
+Locking currently applies to these destructive layer-list operations. It does
+not make the layer data read-only or disable other non-destructive layer
+controls. You may still programmatically interact
+with the layer regardless of the lock state.
```
psobolewskiPhD: Small clarification?
```suggestion
Locking currently applies only to these destructive layer-list operations. It does
not make the layer data read-only or disable other non-destructive changes, such as use of the layer
controls. Finally, you may still programmatically interact with the layer regardless of the lock state.
```

docs/getting_started/layers.md
psobolewskiPhD: I get where Tim is coming from, but I do like the idea of providing the proper way somewhere in the docs, otherwise how will people learn it?
I think Grzegorz's suggestion is the best?

TimMonko: Yes I like Grzegorz suggestion! I didn't realize the enum was public on the module

docs/getting_started/layers.md
TimMonko: Excellent!

Czaki (approved):

PR #1026, opened 2026-05-20
Clean up button css for homepage and gallery

TimMonko:
# Description

In working on #1021, this work was made more difficult by our messy css.
This PR cleans up the class names and combines the css.
It also sets up for #1021 (and #1024) to reduce the review diff.

This improves the css for the gallery buttons by actually override text and outline, instead of the almost invisible buttons of before

Note that before we did *not* style the "alt" homepage button and it was just inheriting from pydata-sphinx-theme. This PR explicitly styles the button to vaguely match what was before, but with a shared outline and text color to not be so intense. This is what I had originally intended for the homepage but got quite
confused by the css. 

I added lots of comments as I debugged along what each was for. :)

image

image

image

image

Main: 

image

image

image

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -1,7 +1,7 @@
 /* ── Buttons for homepage and gallery ───────────────────────────────────── */
 
-.homepage-button,
-.homepage-example-button,
+.button-primary,
+.button-alt,
 div.sphx-glr-download a {
   background-image: none !important;
   border-radius: 4px;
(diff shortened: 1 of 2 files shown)


DragaDoncila (approved):
Looks great, and the csss is more readable too.

psobolewskiPhD (approved):
Looks good to me.

Czaki (approved):

PR #1028, opened 2026-05-22
Add `--override-channels` flag to the `conda install` commands. 

carlosmariorr:
builds on #1023

# Description
This adds the flag `--override-channels` to the `conda install` commands since installing from an anaconda distribution might result in a different `PyQt` version due to priorities to the default channel. 

# Edit:
 This was incorrect, the priorities are working as expected. PyQt is not the culprit. installing python without `-c conda-forge` was the culprit. Installing this from conda-forge is already addressed in the docs so this PR is unnecessary.

### DIFF
--- a/docs/getting_started/installation.md
+++ b/docs/getting_started/installation.md
@@ -87,7 +87,7 @@ conda-forge channel. We also recommend this path for users of arm64 macOS machin
 ([Apple Silicon](https://support.apple.com/en-us/116943), meaning a processor with a name like "M1"). You can install it with:
 
 ```sh
-conda install -c conda-forge napari pyqt6
+conda install -c conda-forge napari pyqt6 --override-channels
 ```
 
 You can then upgrade to a new version of napari using:
@@ -99,7 +99,7 @@ conda update napari
 If you want to install napari with PySide6 as the backend you need to install it using
 
 ```sh
-conda install -c conda-forge napari pyside6
+conda install -c conda-forge napari pyside6 --override-channels
 ```
 
 ````{note}


carlosmariorr:
I discussed this with @Czaki and @brisvag last hackathon.... I dunno whats your opinions on this.

Czaki (approved):

psobolewskiPhD:
Can you clarify why this is needed? 
I mean it should be harmless, but it adds boilerplate and potential confusion.
A channel passed using -c is automatically higher priority than those in .condarc so only packages that are missing from conda-forge would fall through to the channels in .condarc

Czaki:
If I correctly remember, the problem is that, with proper config file content, the package still could be installed from the default channel. It happens for @carlosmariorr that keeps getting PyQt6 from default instead of conda-forge.

psobolewskiPhD:
I looked at the docs some more and I still have doubts about this change.
Would be good to have insight into the issue being solved with a reproducer -- or at least outputs of:
`conda info`
and
`conda config --get`

It sounds like either a very old conda was used or someone disabled the default channel priority:
https://docs.conda.io/projects/conda/en/stable/user-guide/tasks/manage-channels.html#strict-channel-priority

I can't reproduce any issues with just `defaults` in my conda channels (that's the anaconda channels), when passing `-c conda-forge`: all packages are from conda-forge as expected.

I'd feel more comfortable if @jaimergp chimed in before we make this change.

carlosmariorr:
well, dug in a little bit more and I still get the same problem.

what i did:

- created a new account on a fedora and a mac computers
- installed anaconda from the installer in their website.
- did 
```sh
conda create -n napari-default python=3.13
conda activate napari-default
conda install -c conda-forge napari pyqt6
```
- on linux machine it didn't work, napari tries to launch then aborts. in mac it DID work.
- Then 
```sh
conda deactivate
conda create -n napari-override python=3.13
conda activate napari-override
conda install -c conda-forge napari pyqt6 --override-channels
```
- This worked for both computers no problem.

The `pip list` output for "override" environment: PyQt6    6.11.0
The output for the "default" environment: PyQt6   6.8.1

This output is true for both Mac and Linux but only linux fails to launch. My mac computer is still installing PyQt6 from default channel though.

If you don't use the `--override-channels` flag you get a line during env creation or any install command that says in bold green leters: "2 channel Terms of Service accepted" which is probably related? that means its pulling from both channels even though I used the `-c conda-forge`

I do have the outputs of napari-info and pip list for both environments if anyone wants to take a look at them but I think the problem is just that, that it is pulling versions from the default channel when it is there if you don't type --override-channels

Then, because it failed on the linux and not on the Mac, i went on and tried it with an ubuntu machine that is not mine here in the lab to try to figure out if it is something of my configurations. I had the same problem as the Fedora machine, napari didn't launch so i think its not just my setup but something else. I didn't try this using X11.

psobolewskiPhD:
I think the issue is that you don't use `-c conda-forge` here:
`conda create -n napari-default python=3.13`
So the env is already using defaults channel?
(we do specify that in the instructions and I copy pasted from docs in my tests)

Also, can you post `conda config --get` output?

(Also you can use `--dry-run` to avoid making the envs, just look at whether it's all conda-forge or not)

psobolewskiPhD:
PS. If your'e installing anaconda products on **non-personal devices**, I recommend carefully reading the terms of service and/or reaching out to your orgs IT/legal.

carlosmariorr:
> PS. If your'e installing anaconda products on **non-personal devices**, I recommend carefully reading the terms of service and/or reaching out to your orgs IT/legal.

Don't worry, no organization computer, miniforge install nor graduate students were harmed in this process (The graduate student was bribed with chocolate). This all came about because I was using a very old computer from 2015 (Thinkpad T440p for those who still appreciate good keyboards on laptops) that just happened to have anaconda installed. 

Back on topic, yes, using `-c conda-forge python=3.13` during channel creation does solve the problem. The output of any environment with `conda config --get` is the same:
`--add channels 'defaults'      # lowest priority`
after checking the install, the different pyqt6 versions are actually both coming from conda-forge so the `-c` flag is working correctly, it is that python install from the default channel which is breaking the environment. So yeah, I guess that as long as the python is installed from conda-forge it works fine. The `--override-channels` flag actually re-installs some stuff along python which fixes the environment for napari, but the main cuplrit is not the pyqt6 from `default` channel. I'll close this PR if no one has any other comments.

TimMonko:
> Back on topic, yes, using `-c conda-forge python=3.13` during channel creation does solve the problem. The output of any environment with `conda config --get` is the same: `--add channels 'defaults' # lowest priority` after checking the install, the different pyqt6 versions are actually both coming from conda-forge so the `-c` flag is working correctly, it is that python install from the default channel which is breaking the environment. So yeah, I guess that as long as the python is installed from conda-forge it works fine. The `--override-channels` flag actually re-installs some stuff along python which fixes the environment for napari, but the main cuplrit is not the pyqt6 from `default` channel. I'll close this PR if no one has any other comments.

Woah, this is incredibly helpful and insightful! I'll at least stick this in my memory bank for troubleshooting purposes. It's still wild to me that even python from a different channel can break things

carlosmariorr:
So, python install in new env without  `-c` flag in anaconda was the main culprit.

Issue #326, opened 2024-01-18
Audit our use of robots.txt


jni:
## 🐛 Bug

When @GenevieveBuckley searched for "napari working groups" on Google, the top hit pointed to 0.4.15 docs ([ref](https://napari.zulipchat.com/#narrow/stream/298358-working-group-documentation/topic/update-docs-working-group-contacts/near/416509330)). I just tried on DDG and it returned the dev docs. Both cases are bad — a search should return the stable docs. 

scikit-image solves this by disallowing dev and old versions of the docs (but not the home page) in a robots.txt file, [here](https://github.com/scikit-image/skimage-web/blob/main/robots.txt). Note that that file *must* be deployed to the actual served page. (That's [here](https://github.com/scikit-image/scikit-image.github.com/blob/main/robots.txt).)

Incidentally the skimage file is currently a couple of versions behind and needs updating also. 🙃

Issue #364, opened 2022-07-26
noindex, nofollow, rel-canonical for non-stable docs ?


Carreau:
## 📚 Documentation

After the beautiful work from @melissawm on napari/napari#4635, it would be good to have the ability to update the older documentation pages with ["noindex", "nofollow"](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name) and rel="canonical". This should make sure search engine do not index old docs, and give the proper weight to the right pages. 

In particular I think that all the [rel='canonical'](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#attr-canonical) should point to "stable". 

I don't have particular thoughts on whether /dev/ should be no-index/no-follow. 

I believe a good way of doing that would be to inject a placeholder in the theme that is easy to search/replace.

TimMonko:
The plan for this is to monitor the docs/search results after 0.7.1 and decide to continue forward with napari/napari.github.io#444 as appropriate (whether `noindex,follow` or `noindex,nofollow`) to update old docs.
`rel='canonical'` is now used because of #1029 have a set base url in sphinx

PR #1029, opened 2026-05-22
Add baseurl to favor stable versions in search engines

Yann-P:

https://github.com/napari/docs/issues/364

# Description

Set html_baseurl which translates to canonical to point to the stable version. 

However, for this to be effective we need to set noindex as suggested on https://github.com/napari/docs/issues/364, which is what I did in this additional pull request:

https://github.com/napari/napari.github.io/pull/444


TimMonko:
@Yann-P this is awesome!!! Is it really this simple? I've spent hours searching for how to make it so search engines point towards `napari.org/stable` rather than some other version. 

Does this also close #326? 

I guess my question is does this solve "all" the related problems like:
1. search engines pointing towards an old version, despite there being a version on stable
2. a search engine pointing towards a page that no longer exists on the new version (or has been moved)

The reason that I'm asking is because I'd love to upstream this somehow so that folks like me can find it easier. I searched places, including the sphinx docs, and nothing ever stood out to me to meet this need. 
The same with the fix in https://github.com/napari/napari.github.io/pull/444

Certainly _now_ that I have a bit more context, sphinx linking to [RFC6596](https://datatracker.ietf.org/doc/html/rfc6596.html) makes more sense.

However, it would be nice if the [ `html_baseurl` docstring ](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_baseurl) was more explicit on it's function. What I gather is that this probably sets stable to `rel='canonical'` ... or maybe all the old versioned pages?
And then sets all those to `noindex,nofollow`? But in 
https://github.com/napari/napari.github.io/pull/444 you seem to set to `noindex,follow`

Yann-P:
Hello @TimMonko 

> Does this also close https://github.com/napari/docs/issues/326?

yes

> I guess my question is does this solve "all" the related problems like:

1. I think this solves it, what we know for sure is that napari does not have a canonical meta in the latest version HTML (view-source:https://napari.org/stable/) as opposed to other repos using pydata sphinx theme that have it (view-source:https://docs.ray.io/en/latest/ray-core/scheduling/placement-group.html). And the docs explicitely say that setting html_baseurl produces a canonical meta.
2. this would be solved by https://github.com/napari/napari.github.io/pull/444. However maybe we could try to merge this PR (I mean, this one you are reading right now), wait a week and see if the PR that deindexes is still necessary. Perhaps setting a canonical version penalizes the old versions enough that it is not a problem anymore

https://github.com/vroncevic/readthedocs.org/blob/master/docs/guides/canonical.rst

TimMonko:
@Yann-P thank you for taking the time to help me learn. This is absolutely a highlight of my week and a highlight of what makes open source so special to me. You took the time to share your expertise across borders, and I am truly grateful.

I learned something new with `view-source:` ! I always new about the dev-tools view in browsers, but this probably serves much of that need for me now! 

In addition, it helped me understand the canonical part better, for example, I went to view-source:https://docs.ray.io/en/releases-2.55.0/index.html and see that it still points towards the latest url. 

I'm merging this PR so its in our (SoonTM) next release. Thank you so much again!!!! 😁 🌟

TimMonko (approved):

TimMonko:
`view-source:` of the built docs :)
`<link rel="canonical" href="[https://napari.org/stable/index.html](view-source:https://napari.org/stable/index.html)" />`

PR #1033, opened 2026-05-28
Document automatic Wayland/Nvidia workaround in troubleshooting

niksirbi:

Depends on napari/napari#9009
Related to napari/napari#8808

napari/napari#9009 makes napari automatically apply the Wayland/Nvidia startup workaround (setting `QT_QPA_PLATFORM=xcb` and `PYOPENGL_PLATFORM=glx`), and adds an in-app warning for the cases it can't cover. That warning links to the troubleshooting section updated here (`#wayland-and-nvidia`). See that PR
for the full background.

# Description

This PR updates the "Running napari on Wayland with Nvidia cards" troubleshooting section to reflect the new automatic workaround in napari/napari#9009:

- Explains that recent napari versions set the required environment variables automatically at startup, so most users (launching via the `napari` command or a Python script) no longer need to do anything.
- Documents the manual workaround for the case the automatic fix can't cover: when a Qt application is created before `import napari` (e.g. in IPython or Jupyter when `%gui qt` runs first), the user must set the variables themselves before starting the session.
- Notes that user-set values are always respected.
- Adds a stable `(wayland-and-nvidia)=` anchor that the in-app warning links to, while keeping the original heading so existing links don't break.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -36,7 +36,7 @@ menu or you're trying to save layers and can't find the plugin in the listed wri
 plugin using the deprecated `npe1` engine.
 
 You can check if this is the problem by installing `napari<=0.6.6` and turning off the `Use npe2 adaptor` setting
- under `Preferences -> Plugins -> Use npe2 adaptor` and restarting napari.
+under `Preferences -> Plugins -> Use npe2 adaptor` and restarting napari.
 If this fixes your issue, you'll need reach out to the plugin developer to update their plugin in order to use it
 with `napari>=0.7.0`.
 
@@ -99,11 +99,13 @@ If you meet an exception starting from `RuntimeError: Mix of local and non local
    In some rare situations this solution may not fix the environment,
    so you may need to recreate, as described in the first option.
 
+(wayland-and-nvidia)=
 
 ### Running napari on Wayland with Nvidia cards
 
-The Nvidia driver is not yet fully ready for Wayland. 
-We found that on computers with Nvidia cards and a wayland-based desktop environment, napari fails to start with errors such as this:
+The Nvidia driver is not yet fully ready for Wayland.
+We found that on computers with Nvidia cards and a Wayland-based desktop environment, napari may fail to start with errors such as this:
+
 ```pytb
 OpenGL.error.GLError: GLError(
     err = 1280,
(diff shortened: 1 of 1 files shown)


brisvag (approved):
Love it!

docs/troubleshooting.md
```diff
@@ -112,9 +114,24 @@ OpenGL.error.GLError: GLError(
...
-The problem might be solved by setting following environment variables:
+
+The problem can be worked around by forcing Qt to use X11 (via XWayland) and PyOpenGL to use GLX, by setting the following environment variables:
+
 ```sh
 QT_QPA_PLATFORM=xcb
 PYOPENGL_PLATFORM=glx
 ```
 
+Recent versions of napari set these variables automatically at startup, so launching via
```
TimMonko:
```suggestion
napari sets these variables automatically at startup, so launching via
```

This suggestion is only because our docs are versioned, so "recent" has (to me) confusing transient implications. Thus, I think it's better to just factually state the current functionality as-is for each versioned release of the docs (because one can go back and see older doc versions as needed)

docs/troubleshooting.md
```diff
@@ -112,9 +114,24 @@ OpenGL.error.GLError: GLError(
...
+the `napari` command or a Python script usually works without any action on your part.
+napari only sets these variables if you haven't, so your own values are always respected.
```
TimMonko:
```suggestion
napari only sets these variables _for the current session_ if you haven't, so your own values are always respected.
```

This is the intent, right? That it is respecting global vars and just using session-scope (or whatever the word is for Linux) vars?

docs/troubleshooting.md
```diff
@@ -99,11 +99,13 @@ If you meet an exception starting from `RuntimeError: Mix of local and non local
    In some rare situations this solution may not fix the environment,
    so you may need to recreate, as described in the first option.
 
+(wayland-and-nvidia)=
```
TimMonko, line 102: 👍 saw your comment in the napari PR and this is perfect. We have always respected these anchors -- even sometimes when they _really_ drift from what the header becomes, but hey, that's their purpose! Thank you :)

TimMonko (approved):
Absolutely lovely! Thank you!!!! 🤩 

I left two comments, but you are more then welcome to do with them as you please. :)

docs/troubleshooting.md
niksirbi: I yeah, that's absolutely right. I was thinking of people who may still be running older version of `napari`, but completely forgot for a moment that the docs are versioned.

docs/troubleshooting.md
niksirbi: Yes, that's correct.

niksirbi:
Thanks both! Probably it make sense to wait till we settle on how https://github.com/napari/napari/pull/9009 lands before this one gets merged, right?

TimMonko:
> Thanks both! Probably it make sense to wait till we settle on how [napari/napari#9009](https://github.com/napari/napari/pull/9009) lands before this one gets merged, right?

Yep! And because you have the "Depends on" in the OP our CI blocks it :)

niksirbi:
> Yep! And because you have the "Depends on" in the OP our CI blocks it :)

Good to know! I should start using that too.

niksirbi:
I pushed an additional commit [`b8531b4` (this PR)](https://github.com/napari/docs/pull/1033/commits/b8531b49dec200d02d1f834d4947b45138aeaeef), as discussed in   https://github.com/napari/napari/pull/9009.

TimMonko:
Looks great! Thank you for your continued effort and time on this!

PR #1035, opened 2026-06-02
Fix: Opaque, instead of translucent, sticky version banner

TimMonko:

Follow-up to napari/napari-sphinx-theme#228

# Description

The version banner had some transparency (see screenshots), likely as a hack to make the button more prominent. This modifies the css to blend with a theme-appropriate color (namely `napari-gray` for light, and `napari-dark-gray` for dark) so that there is no longer transparency. This _mostly_ matches what we saw before.

Top is current, bottom is this PR:
image

image

**Note:** I actually like the transparent banner because I feel like it makes it more noticeable, but for it to look intentional we would also want transparency on the navbar.

**Note 2:** We should also consider having our own theme color set in napari-sphinx-theme for the banner, because this is inheriting from pydata-sphinx-theme colors. I just wanted to mention. It's also potentially a good thing that it's quite noticeably a different color from the theme, but still vaguely matches the palette -- I think that makes it stand out more.

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -101,11 +101,11 @@ html[data-theme="dark"] .button-alt:hover {
 
 /* Version warning banner color */
 #bd-header-version-warning {
-  background-color: color-mix(in srgb, var(--pst-color-secondary-bg), transparent 30%);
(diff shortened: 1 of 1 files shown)


melissawm (approved):

melissawm:
Merging as an urgent usability fix.

PR #1036, opened 2026-06-02
Update homepage video with reproducible napari-animation script

### LINKED ISSUE #680: napari.org video: https://napari.org/stable/
The flagship video needs to be updated for 0.6.0 changes.
Ideally it would include toggle visibility of the labels @TimMonko  did in https://github.com/napari/docs/pull/667

TimMonko:

Closes #680 🥳 🎉 👏 

# Description

## Reproducibility 

napari/napari-animation#261 introduces a timeline-widget to the plugin. I wanted to both practically review this, and take advantage of the easily modifiable and saveable json to redo the homepage video. To accomplish this PR I had to learn quite a bit, including about napari-animation generally, more about zarrs local vs remote, async, and video encoding (this was the hardest part!).

This script uses a uv header and is currently pointed at https://github.com/TimMonko/napari-animation.git@766dbbbcd0ec8607e67f49da0402ad41d4e512fb because this has Windows patches and webm functionality (because webm really is 10X smaller in this scenario). 

In the future, re-rendering the homepage is just: `uv run docs/_scripts/homepage_video.py render`
One really cool part of how napari-animation works is that the data can be remote still, the rendering take a _long_ time (because of that) and it still is a 1-to-1 rendering compared to having the data locally 😁 
This also serves as functional insight into where napari-animation can/should go.

## Design choices

Some opinionated choices include the following, the I think for all of them being able to swap with napari-animation mid timeline would be cool! 
1. using the `magma` LUT instead of grayscale. I like this pop of color and I think it does give a bit of insight into rendering. 
2. attenuated_mip rendering only. 
3. the way it spins... this is troublesome because the animation interpolation wants to only do positive angles so it causes compelte overrotation in some cases causing these to be quite finnicky
4. the overall length is shorter, and I kind of dislike how it has less "human pauses" but I've quickly gotten used to it
5. used ndevio to load the image because I knew it would handle the metadata as I wanted :P can re-evaluate
6. turned the scalebar on to show off our unit aware rendering and scalebar
7. turned on the axes overlay to better show the dimensionality of the viewer
8. added axis labels

image

image

## And more notes

- I used the css for the feature example to add a drop shadow and rounded corners to the video. 
- I overwrote the screenshot -- just noting that the html suggests an mp4, but I don't see one committed
- This could be a higher resolution, but I opted to keep the filesize comparable to before. I didn't notice that much difference either doubling the quality, though it double the filesize to ~16MB.


brisvag:
> the way it spins... this is troublesome because the animation interpolation wants to only do positive angles so it causes compelte overrotation in some cases causing these to be quite finnicky

Mmm, this is annoying... Ideally we would expose the camera rotation as quaternions to avoid this. I'll think on how to do this without adding too much special casing.



> the overall length is shorter, and I kind of dislike how it has less "human pauses" but I've quickly gotten used to it

I think that the fact that all interpolation seems sto be linear (the default) also results in a video that feels "less human". Have you tried playing with that?

psobolewskiPhD:
Phone post as I see my post from last night is missing:
Looks great! 🤩 

But I wonder if it's too fast??! 
Do we give a false impression of performance? Or set unrealistic expectations?
I guess it's because the frames are rendered out 1 at at time and then played back at arbitrary fps...

TimMonko:
> I think that the fact that all interpolation seems sto be linear (the default) also results in a video that feels "less human". Have you tried playing with that?

Yep! It's pretty much all linear with a few "step" ones. Once I actually right the review in napari-animation I'll discuss some of the UX issues I had with that. I will play with some of the interpolations -- any recommendations. TBH there are too many so it became overwhelming to try to sort through them. Will play with this though.




> Phone post as I see my post from last night is missing: Looks great! 🤩
> 
> But I wonder if it's too fast??! Do we give a false impression of performance? Or set unrealistic expectations? I guess it's because the frames are rendered out 1 at at time and then played back at arbitrary fps...

Thanks! Curious what you mean by too fast? With the data locally (which is what this specific .webm is from) I get this type of real-time performance, and it's how napari could _theoretically_ work -- though that's where I point out it feels like it's missing the human pauses. I also avoided doing human-impossible things like moving both dims sliders simultaneously. (The exception sort of being rotating the camera and moving the dims slider, but that is technically possible by hitting play on the slider and then interacting with the canvas)

As for arbitrary fps, I'm curious what you refer to? I'm a bit confused on the UX (@brisvag) because I can set fps both in the animation-timeline widget and also when I save the video, I'm not sure if they interact, or if the timeline fps was just for tuning. I designed this around 30fps because those videos are smaller but still look smooth enough to me

docs/_static/images/tribolium.webm
TimMonko, line 1: Noting that locally this file is 7.2MB for me, but 1.6MB on github. Interesting, so maybe the remote compresses my compressed file? Anyways, I'm quite happy about that

melissawm:
I'd be in favor of merging and revisiting this later if it bothers people enough!

docs/_static/images/tribolium.webm
TimMonko: Oh Melissa and I think this might be a (secret) delta

TimMonko:
I don't have much time to prioritize this myself  right now either, so if folks want to merge and make changes themselves that would be great! I'm happy with it to go in as is.

melissawm (approved):

Czaki:
Did I correctly understand that this PR depends on forked version of napari animation? Is there an opened PR for that?

brisvag:
Yes: https://github.com/napari/napari-animation/pull/261

PR #1045, opened 2026-06-16
Fix formatting in version_switcher.json (remove spurious comma)

psobolewskiPhD:
# References and relevant issues
Versions switcher live is broken, see napari.org and https://napari.zulipchat.com/#narrow/channel/215289-release/topic/0.2E7.2E1/near/603721824

# Description
Removes the extra comma which I think is the cause.

PR #1046, opened 2026-06-16
Update pre-commit config, apply rules

Czaki:

# Description

When checking #1045 I found that we do not run pre-commit at all. 

This PR adds check-json (to avoid bugslie solved in #1045).  

Update pre-commit hooks. Apply hooks to codebase.


.pre-commit-config.yaml
```diff
@@ -31,7 +33,7 @@ repos:
       types_or: [ python, pyi ]
 
 - repo: https://github.com/executablebooks/mdformat
-  rev: 0.7.22
+  rev: 1.0.0
   hooks:
     - id: mdformat
       additional_dependencies: [mdformat-myst, mdformat-ruff]
```
TimMonko: so I think this is what we want to turn off in this PR, and then rerun pre-commit on the config so we don't touch .mds for now (except like end-of-file-fixer or whatnot)

Czaki: done

TimMonko (approved):
LGTM thanks!

PR #1048, opened 2026-06-18
Fix linkchecker action constraints to file for py312

psobolewskiPhD:
Noticed that the action was failing, see: https://github.com/napari/docs/actions/runs/27728696054

# Description
Updates the constraints file.

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -41,12 +41,12 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install "napari/[pyqt5, docs]"
         env:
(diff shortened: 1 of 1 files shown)


TimMonko:
Thank you Peter! This looks good to me. Would be nice to figure out how to not have to maintain all these constraints in the workflows. Quite easy to miss these when updating. That, or we should doc all the constraints places

TimMonko (approved):

.github/workflows/linkcheck.yml
TimMonko:
```suggestion
          python-version: "3.12"
```
adding 3.12 here too

PR #1070, opened 2026-07-15
Update homepage video for 0.8.0/0.9.0 changes

TimMonko:
# Description

Re-renders homepage video especially after base dark theme changes with 0.8
Also adds the new floating axes and histogram to the video :)
Making in time for SciPy tools plenary. No rush to merge (of course)

image


jni (approved):

TimMonko:
hmmm I noticed this video is taller. Note to my self to figure out why. Probably because of enabling the histogram somehow makes the viewer bigger (or not shrink donw) on the resize

brisvag:
@TimMonko still planning to get this in?

TimMonko:
Updated and ready to go. Thanks for the patience

brisvag:
Looking so cool! Some thoughts:

- what do you think about themoving the box on the axes?
- I'm biased, but I feel that starting in 3D would be the more impressive thing to show off first... :P

brisvag (approved):

TimMonko:
> Looking so cool! Some thoughts:

Thanks! I look forward to us having the headspace to push on napari-animation 🫣 

>     * what do you think about themoving the box on the axes?

 I thought about it when setting this up again. I opted for keeping it as-is because it mirrors the GUI experience of toggling it on. I think the closer this video looks to what someone sees out-of-the-box is better and makes it seem more polished. If it wasn't a homepage example I think customizing more would be nice to see. Now that we auto inherit viewer axis labels, this video can be entirely replicated GUI-only (napari-metadata made it possible before but 🤷 )

>     * I'm biased, but I feel that starting in 3D would be the more impressive thing to show off first... :P

Understandable! I'm going to leave it this way for now to focus on other work, but I'd be open to seeing a timeline like that. I quite like the start in 2D actually, because it's also like "wow matplotlib but 🚀 for slices"

DragaDoncila:
I think let's get this in because it's a strict improvement, and we can follow-up with tweaks pretty easily whenever now that it's a script.

PR #1091, opened 2026-07-28
Bring several NAPs up to date

brisvag:
# Description
NAPs are supposed to be Accepted, Widthdrawn or made Final when appropriate. We haven't been on top of this and several NAPs that have _effectively_ changed status were never formally changed. This PR does that, for those naps whose informal status is hopefully obvious :P

Feel free to disagree below and I'll update accordingly, let's leave this open for a while!

cc @napari/core-devs


willingc (approved):
Thanks @brisvag

jni (approved):

PR #1093, opened 2026-07-30
Unify search across sub-sites

ctrueden:
This PR is in response to [an issue discussed on the napari Zulip](https://napari.zulipchat.com/#narrow/channel/619705-2026-naPLari-Krakow-public/topic/Two.20napari.2Eorg.20workshops.20pages.3F/near/613499280). For convenience, here is the relevant summary:

> We noticed two different napari workshop pages:
>
> * https://napari.org/stable/further-resources/napari-workshops.html
> * https://napari.org/workshops/
>
> Only the latter seems to have an obvious link to the Express workshops. The other one ends at 2024, and seems to be all external links—e.g. searching for "express" on this external-links-oriented page does not match anything. (And actually, searching for "express workshop" in the napari.org search bar turns up nothing either.)

This PR addresses the issue by switching to [Pagefind](https://pagefind.app/), a static client-side search capable of aggregating across multiple indices. It's the perfect (and only, as far as I am aware :sweat_smile:) solution to the fragmentation of search between sub-sites of documentation.

This is a first cut that works in my local testing (on macOS) between `docs` and `workshops`. For it to fully work, the companion PR for napari/workshops#69 must also be merged.

Here is a screenshot of the docs search seeing content from workshops:

search-workshops-from-docs

When the page is wider, the search bar moves to the right instead of centered, along with the search results. A remaining issue is that the new Pagefind-based search results panel is not centered and enlarged like the current docs search—hopefully some additional styling can fix that.


willingc (commented):
Very cool addition @ctrueden. I didn't review all the css and js. If it works, it's a very helpful enhancement for users. ✨

brisvag:
@aganders3 maybe you can take a look at the code for review? :)

melissawm:
Hi all, I don't quite understand why we would want to have this "hard dependency" on the workshops to build the main docs. I am probably missing something but ideally we would only need the index of the workshops and keep all the implementation (javascript, css, search button etc) here in the napari docs repo. Is that possible?

ctrueden:
> I don't quite understand why we would want to have this "hard dependency" on the workshops to build the main docs.

@melissawm The only reference to the workshops is for Pagefind's index merging feature:
```
  data-merge-paths="/workshops/pagefind/"
```
The docs here continue to build independently of the napari workshops content; it's just that the client-side javascript tries to grab the search index at the `/workshops` base as well and combine it, if available, when using the search bar.

Does that answer your question?

melissawm:
I see! So this would only be a hard requirement for the first time, because the search index on the napari-workshops side is still not set up, correct? 

In that case, I'm fine with merging this and iterating if we see opportunity for improvement. I have briefly looked at the CSS and javascript and (not being an expert) it all looks fine to me, so I'll approve!

melissawm (approved):

PR #1103, opened 2026-08-10
Prefetch data that are problematic and causing docs build failure

### LINKED ISSUE #1089: Add prefetch steep to download all remote resources on begin of docs build
## 🧰 Task

We currently have multiple failures of docs build in the middle of a long run because of the inaccessibility of some remote resources. 

As far as I checked, all downloads are done in a cacheable way (pooch or other lib like this).

So we could add a prefetch steep befor start building html. 

Advantages:
1. Early fail on inaccessible resources, no more parsing of long logs 
2. Better ability to cache it. 
3. If data are too big on cache, we could put them on our server, and prepopulate before tests with source codes pointing to original data. 

Disadvantages
1. Next list to keep in sychronisation

Czaki:

closes #1089

# Description

This PR adds a step of prefetching fragile data required during the build of docs. It does not prevent docs build from failing, but it should happen in the first minute. 

So it should be easier to understand why docs build fails and use less CI time. 

The next step might be to use this script to download data from an alternative location and put it in the expected place.

### DIFF
--- a/docs/_scripts/prefetch_data.py
+++ b/docs/_scripts/prefetch_data.py
@@ -1,3 +1,6 @@
+import os
+from pathlib import Path
+
 from nilearn import datasets
 import pooch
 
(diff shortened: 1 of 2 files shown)


jni:
@Czaki what about adding the download dirs to the actions cache so that they don't need to be re-downloaded?

Czaki:
> @Czaki what about adding the download dirs to the actions cache so that they don't need to be re-downloaded?

Is there a reason to keep it outside the default Pooch cache directory?

DragaDoncila:
@Czaki I had a look at this and it seems reasonable? But.. how do I check/know whether it's worked and the cached data is being used rather than refetched?

brisvag (commented):
Add some print? xD

Czaki:
For `pooch.retrive` it is guaranted by `known_hash='5b43ed0269eaa1eebf4c48079270a30e6cc40e87f20cc36c1b3d5a07c51c7b20'` and pooch documentation. 

For `nilearn` I checked their codebase. It might change in any release, but for now it checks if a file exists on the drive and avoids download in such a case. 

EDIT. 

For cache using `action/cache`. I dont know. Somehow patch the pooch and count miss/hits?

jni:
> Is there a reason to keep it outside the default Pooch cache directory?

Sorry, I missed that we were already using it! I understand now! :pray:

TimMonko:
> Sorry, I missed that we were already using it! I understand now! 🙏

So, do you approve? 😁

jni (approved):

PR #1113, opened 2026-08-18
Info about Xarray metadata inheritance

TimMonko:

Follow-up to napari/napari#9316

# Description

Update: per Lorenzo's suggestion, this has been primarily added to the units and scale guide.

Feedback requested: Where should this info actually live?

I placed some info about the Xarray metadata inheritance into the Image guide, but it is still applicable also to Labels layers.
This is just the section where we talk about array types, so it seemed like a level-of-detail appropriate place to stick it.


brisvag (commented):
How about putting it in https://napari.org/stable/guides/units.html#scale-and-unit-aware-rendering?

jni (approved):

PR #1115, opened 2026-08-19
Clean Pydantic BaseModel objects out of API docs

### LINKED ISSUE #964: Some API docs are much harder to parse with 0.7.0 (pydantic v2?)
## 🧰 Task

Compare 0.6.6: https://napari.org/stable/api/napari.Viewer.html

Image

To latest: https://napari.org/dev/api/napari.Viewer.html

Image

Doesn't effect every class, but navigate around and you'll be able to find others with this pattern (like napari.utils.Colormap)

napari.layers seem to be the same, for example

TimMonko:

Closes #964

# Description

Pydantic v2 causes every `EventedModel`-based API page (Viewer, Camera, Colormap, ...) cluttered with inherited `BaseModel` members. Here we use `conf.py` to hide them from the tables and the autodoc Details sections.
napari's own overrides (`model_dump`, `json`, `model_json_schema`) are kept. IMO this is desirable because the API docs aren't just for "everything" but should be a bit more opinionated, especially for _most_ users that don't need these.

Dropped from the tables:
- **20 methods** — 11 deprecated v1 (`copy`, `dict`, `construct`, `parse_*`,  `schema*`, `validate`, `from_orm`, `update_forward_refs`), 8 v2  (`model_construct`, `model_copy`, `model_dump_json`,  `model_parametrized_name`, `model_rebuild`, `model_validate*`), and  `model_post_init`
- **5 attributes** — `model_computed_fields`, `model_config`, `model_extra`,  `model_fields`, `model_fields_set`

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -743,65 +743,43 @@ def qt_docstrings(app, what, name, obj, options, lines):
 # -- Pydantic model internals ---------------------------------------------
 
 
-#: Attributes inherited from pydantic's ``BaseModel`` that are internal
-#: schema/serialization bookkeeping.  They are rendered with pydantic's own
-#: docstrings and only clutter napari's API reference, so we hide them.
-_PYDANTIC_MODEL_ATTRIBUTES = frozenset({
-    'model_computed_fields',
-    'model_config',
-    'model_extra',
-    'model_fields',
-    'model_fields_set',
-})
+def _skip_pydantic_base_model_methods(app, what, name, obj, skip, options):
+    """Skip methods inherited from pydantic's ``BaseModel``.
 
-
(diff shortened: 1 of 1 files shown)


brisvag (commented):
Is there a way to add at least a collapsed section with this, or at least a link to the pydantic docs?

TimMonko:
> Is there a way to add at least a collapsed section with this, or at least a link to the pydantic docs?

This is fully generated by Deepseek, but I managed to inject that into such docstrings:

image

Thoughts?

brisvag:
I like this!

TimMonko:
Looked at in docs meeting with Draga.
We think the dropdown should just be a admonition since it has such little content.
Also, found a bug where we broke crosslinking in the Attributes table:
this PR left, deployed right
image

PR #1116, opened 2026-08-19
API Docs: Move type hints out of signature and into params

TimMonko:
# Description

Moves the type hints out of the API docs signature. https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_typehints
and into the parameters section of the docstring. This does mean that parameters without descriptions are now shown in the bulleted (napolean? I think is what its called) parameter list.

Personally, I find this to be overall much more readable, and any issues with it are **our** problem because of lack of docstrings and means that we should be doing better inside teh napari source code.

This PR on left, 0.8 on right:

image

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -338,6 +338,10 @@ myst_substitutions = {
 
 }
 
+# -- Autodoc ----------------------------------------------------------------
+
+autodoc_typehints = 'description'  # Render parameter types in the descriptions rather than the signature.
+
 # -- Autosummary ------------------------------------------------------------
 
 autosummary_generate = True


Czaki:
For me, the main problem with readability comes from too many parameters. The PR healing symptoms, not reasons. 

I like to see type annotation next to the parameter name, but if the type is complex, it might reduce readability.

brisvag (changes requested):
This is wrong AFAICT? You can't pass any of those things to the viewer constructor. The current docs are correct. In fact, the constructor signature is wrong even on main.

TimMonko:
> The current docs are correct. In fact, the constructor signature is wrong even on main.

By the current docs, do you mean the parameters section is correct? And by the constructor signature is wrong, you mean that it sohuld _not_ contain the extra params?

brisvag:
Correct! Try

```py
import napari
napari.Viewer(camera='anything')
```

and it'll fail, because we override the init to only take those parameters there. ViewerModel the same.

TimMonko:
Closing this as its either due to pydantic-type issues or us doing things not quite as it should be in the API

PR #1117, opened 2026-08-19
Update `axis_labels` guidance for 0.9

TimMonko:

Follow-ups to napari/napari#9282 and napari/napari#9293 (and others)

# Description

Softly updates the guides to reflect the new axis labels inheritance behavior. For now, I didn't totally remove the ability to set it directly on the dims, but I think we are working towards that? I'm not sure how strongly we want to do that yet, given the popup renaming exists as does the API. 

Especially happy to hear folks ideas on how best to go about this, but I mostly wanted to be 0.9 ready.


brisvag (approved):

docs/guides/handedness.md
```diff
@@ -46,6 +46,7 @@ viewer, sst_layer = napari.imshow(
         sst.sst,
         name='sea surface temp',
         colormap='magma',
+        axis_labels=sst.sst.dims,
```
jni: This is unnecessary because of the xarray inheritance, right?

jni (approved):
All good, but minor query about the update to handedness.md?

jni:
@Czaki if you want to tag before @TimMonko wakes up, feel free to merge this: my concern in this PR is not critical to fix before release.

docs/guides/handedness.md
TimMonko: gah thanks for catching! It was from a bad merge conflict fix on my part -- I thought I even double checked this didn't come back, so thank you for noticing. This is what originally made me notice the whole thing needed updating like in #1118

TimMonko:
```suggestion
```

PR #1118, opened 2026-08-19
Update handedness guide for 0.9 changes

TimMonko:

Found this while working through #1117 and realized also relveant to #1113

# Description

Due to the changes with box Xarray and axis_labels inheritance, this guide can be cleaned up quite a bit,
which is nice because then it focuses on the actual handedness info!

I opted for a new PR since this touches both related things and it would be easier to review this way? I hope :laugh:


brisvag (approved):

jni (approved):
❤️

PR #1120, opened 2026-08-21
Use a pages first image for OpenGraph Preview if there is one

TimMonko:
# Description

This should use the first image on the page that isn't an svg or .ico
It doesn't filter for size or anything so maybe has weird edge cases
cc @jni


jni (approved):
I think it's worth trying. We can always revert if things are worse than the generic image.

Issue #1018, opened 2026-05-18
Look at alternative formater/linter for markdown


psobolewskiPhD:
## 🧰 Task

I came across this project today:
https://github.com/rvben/rumdl

supports a lot of markdown flavors:
https://github.com/rvben/rumdl/blob/main/docs/flavors.md

There is also:
https://github.com/akiomik/mado

which does just lint

melissawm:
Is the main goal to solve a problem with the current linter? If so, maybe we should start there - what would you like to see or not see  @psobolewskiPhD ?

There is also [Vale](https://vale.sh/) which is even more powerful...

TimMonko:
We currently don't have linting/formatting on (see #1047), so that's probably the priority atm

psobolewskiPhD:
I don't think there was a specific issue, rather just some interesting tools I came across -- probably in some random blog post or social media.
I recalled the complexities and issues with formatting our docs from https://github.com/napari/docs/pull/54

Probably closable with #878

PR #839, opened 2025-09-15
Add EffVer to release guide and link to checklist template

TimMonko:
# References and relevant issues

Closes #778
Addresses part of #833 

# Description

1. ~~Adds checklist to release guide (we should probably rip this out and put in a napari/napari Issue template?~~ Ripped out and put into napari/napari#8295
2. Cleans up release guide and adds more detail for some of my sticking points
3. Adds some info about EffVer and updates language throughout as needed.
4. Adds info about private vs public github assets for properly linked images in release notes.

PR #1123, opened 2026-08-26
Fix incorrect image link formatting in 0.9 release notes

jni:
This is causing a missing image in the release notes.

It might be worth manually copying the fix to 0.9.0 once deployed.


Czaki (approved):

Czaki:
It will be nice to resolve #1018 to avoid such a problems in the future.

TimMonko (approved):
noting that these are paths to files in the docs, while our release guide says to use Github assets so they render on the Github releases page
This is ok though and I'd rather get the fix in now

jni:
> while our release guide says to use Github assets so they render on the Github releases page

As far as I'm concerned that discussion [wasn't resolved, it just petered out](https://github.com/napari/docs/pull/839#discussion_r2367256562), and I am still dead set against building a dependence on GH for our hosting. Perhaps [more so](https://mrshu.github.io/github-statuses/?view=all) than I was when we last had the discussion.

PR #1132, opened 2026-09-14
Prepare tracks data docs

Olusesan:
This PR adds a documentation example for the new prepare_tracks_data helper function. It depends on PR https://github.com/napari/napari/pull/9015

The example demonstrates how to prepare track data from a pandas DataFrame when the DataFrame uses custom column names or a different column order than expected by Napari. It shows how to map DataFrame columns to the track ID, time, and spatial coordinates before passing the resulting data to viewer.add_tracks().

The example uses a simple cell-tracking dataset with particle, frame, depth, row and column information.


pixi.toml
```diff
@@ -10,6 +10,9 @@ version = "0.1.0"
 [feature.base.dependencies]
 python = "3.13.*"
 make = "*"
+proj = "*"
```
DragaDoncila, line 13: @TimMonko do you know why we'd be needing these extra dependencies here? It seems like a pretty vanilla addition...

docs/howtos/layers/tracks.md
```diff
@@ -166,6 +166,39 @@ viewer.add_tracks(tracks_data)
 napari.run()
 ```
 
+## Preparing tracks data from a pandas DataFrame
+The prepare_tracks_data helper maps DataFrame columns to the track ID, time, and spatial coordinates expected by napari. This is useful when track data uses custom column names or when the columns are not already in the order expected by napari.
```
DragaDoncila:
```suggestion
The prepare_tracks_data helper maps DataFrame columns to a numpy array with the track ID, time, and spatial coordinates in the order expected by napari. This is useful when track data uses custom column names or when the columns are not already in the order expected by napari.
```

DragaDoncila (commented):
Thanks for the PR @Olusesan! I had a look and left two comments - one for you and one for @TimMonko.

pixi.toml
TimMonko: I certainly don't know why these would be here. @Olusesan do you have any more context?
First (and its confusing) new dependencies would be added to napari's pyproject.toml docs dependencies. Second, I'm also already seeing rasterio as a part of the docs build: https://github.com/napari/napari/blob/main/resources/constraints/constraints_py3.12_docs.txt
Finally, this doesn't import any of these (it's a nice simple example!) so I'm not sure why this is here.

TimMonko:
Thanks for the docs @Olusesan, always great to see accompanying changes! It brings me great joy!

First, your code here has an additional `depth` column, which is cool, but is also missing from napari/napari#9015. Second, you can [cross-reference to gallery examples ](https://napari.org/stable/developers/contributing/documentation/index.html#cross-referencing-gallery-examples).

What I think should happen here is that the code here in the docs should be deleted, the example from your napari PR updated -- and then we don't have to maintain and keep in sync the two code blocks.

pixi.toml
Olusesan: Ahh yes I added those environment variables when I was trying to setup pixi because it was missing some dependencies and kept failing. It was my first time using pixi (docs strongly recommended it) and for context I'm on Mac with arm64 processor.

docs/howtos/layers/tracks.md
Olusesan: Thanks for the annotations @DragaDoncila, it's corrected and a new commit is incoming.

Olusesan:
Thanks @TimMonko and I thought about this for a moment and I just wanted to give the most explicit example in the docs . I will delete the code in docs and include the z dimensions in the PR.

PR #1134, opened 2026-09-16
Typing guide: Using Pyrefly and understanding the config

TimMonko:

For napari/napari#9395

# Description

This adds a **new** typing guide based on pyrefly to our contributing docs. I wrote this to self-review napari/napari#9395 and because it's a place we often get questions about. I focused on including a few ideas:

1. Contributors are not responsible for understanding/fixing the check fail
2. Basic check commands for any contributor, and explaining CI
3. Explaining how to navigate typing and pyrefly / expectations
4. An explanation of our current shortcomings
5. Some detailed breadcrumbs at the bottom so folks can figure out how everything links up (i.e. config, constraints, tox, CI)

I also added a pointer to the upcoming napari/island-dispatch#40 to prep for the real link (i.e. I want to keep the narrative separate from the technical), but if we want to merge this before that is ready, we should be sure to take it out and re-add later.


docs/developers/contributing/typing.md
```diff
@@ -0,0 +1,94 @@
...
+tox -e pyrefly -- suppress --remove-unused=all
+```
+
+Don't run pyrefly from your development environment (e.g. `uv run pyrefly check`):
+it resolves the packages you happen to have installed rather than the pinned ones,
+and reports a different set of errors. In addition,
+the [code editor integration](https://pyrefly.org/en/docs/IDE/) reads the same
+`[tool.pyrefly]` configuration, but it sees your environment too — so treat editor
+suggestions as a hint and `tox -e pyrefly` as the standard.
+Sometimes, a type will genuinely not properly solve because of
```
Aniketsy: i guess we can have one heading for `Suppressing an error`

docs/developers/contributing/typing.md
```diff
@@ -0,0 +1,94 @@
...
+```
+ERROR src/napari/utils/tree/node.py:64:31-53: Argument `int | None` is not assignable
+to parameter `object` with type `int` in function `list.insert` [bad-argument-type]
+                                                                 ^^^^^^^^^^^^^^^^^
+```
+
+Put that rule on the offending line:
+
+```python
+indices.insert(0, item.index_in_parent())  # pyrefly: ignore [bad-argument-type]
```
Aniketsy, line 58:
```suggestion
indices.insert(0, item.index_in_parent())  # pyrefly: ignore[bad-argument-type]
```

Aniketsy (commented):
@TimMonko thanks for the work ❤️ 

Looks good to me 👍

docs/developers/contributing/typing.md
TimMonko: Interestingly, `pyrefly suppress` adds it the way that I wrote it (see all the changes in napari/napari#9395)-- but the docs show it without the space: https://pyrefly.org/en/docs/error-suppressions/
I think we shall keep it in the format that it automatically gets added but I dunno lol

docs/developers/contributing/typing.md
TimMonko: Good idea -- I did that in #1135, so might as well bring that here.
[TimMonko/docs@`typing-guide...TimMonko:docs:typing-guide-tips`#diff-5633a0c99c](https://github.com/TimMonko/docs/compare/typing-guide...TimMonko:docs:typing-guide-tips#diff-5633a0c99c551ff404c4eda17d5c9e83e0eb1d984c286394d0f34edee92ae631R76-R83)

docs/developers/contributing/typing.md
Aniketsy: > I think we shall keep it in the format that it automatically

yes , lets keep if thats how pyrefly suppress adds

willingc (approved):
Beautiful work and a fabulous improvement over mypy for performant checks.

docs/developers/contributing/typing.md
Aniketsy: maybe you find this intresting 😁 https://github.com/facebook/pyrefly/pull/4992 , i've fixed the issue now pyrefly will generate ignore comment without space, and it can accept both, with or without space

